PR #2, opened 2016-03-28
WIP Button component

rgarcia:
I took the Button that @jacobsimon and @kshen0 have in the portal2 branch of the launchpad and extended it based off of @CarlNelson's spec (https://www.dropbox.com/s/pa4b0pwshg8fp1p/Button%20Spec.png?dl=0).

I also made a really quick/hacky index.html file that runs a `Demo` component that sort of showcases the `Button`. I added a webpack config that compiles all of this, and also mirrored this branch on a `gh-pages` branch. You should be able to see the result here http://clever.github.io/components/

Some todos:
- [ ] @CarlNelson in person review
- [x] @jacobsimon review on gh-pages/webpack setup, since you're thinking about this
- [ ] @rgarcia document `props` in readme
- [x] @rgarcia implement transitions from spec
- [x] Address active state issue @jacobsimon brought up
- [x] Webpack-ify Clever font

cc @Clever/front-end


eighnjel:
They look like tabs o.0 ... but it matches the design spec so ¯\_(ツ)_/¯

jacobsimon:
Nice! I think the font-size is broken. Have you tried importing this in a project? I'm not sure if having another webpack config will cause problems or not.

osdiab:
@jacobsimon screenshots:

screen shot 2016-03-29 at 11 29 24 am
screen shot 2016-03-29 at 11 29 29 am

osdiab:
@jacobsimon i think that should be fine given that `index.js` is listed as `main`, but it's worth checking.

src/Button/Button.less
```diff
@@ -0,0 +1,75 @@
...
+
+.Button--primary {
+  background-color: #4274F6;
+  color: #FAFCFC;
+  border-bottom: 3px solid #345CC4;
+  //box-shadow: 0 3px #b1cbf2;
+  &:hover,
+  &:focus,
+  &:active {
+    //outline: 0;
```
jacobsimon: This will get rid of the blue border when you click/focus, which is fine in terms of accessibility since we're also changing the color. We should probably add a different active state since there isn't any feedback when you actually click, but I'll leave that to carl

src/Button/Button.jsx
```diff
@@ -0,0 +1,23 @@
...
+    if (["primary", "secondary", "destructive"].indexOf(this.props.type) > -1) {
+      classes += `Button--${this.props.type} `;
+    }
+    if (["large", "regular", "small"].indexOf(this.props.size) > -1) {
+      classes += `Button--${this.props.size} `;
+    } else {
+      classes += `Button--regular `;
+    }
+    return (
+      <button className={classes} onClick={this.props.onClick} disabled={this.props.disabled || false}>
```
jacobsimon: Could we change this to render `<a>` if `href` is provided?

.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "presets": [
+    "es2015",
```
jacobsimon: Using es6+ here will prevent us from using this in any repo that hasn't been converted yet, unless we're going to compile the components down to es5 and then import them. Not necessarily bad but something to keep in mind.

rgarcia: AFAIK ES6 == ES2015, so this is just E6, but yeah, this might make it harder to import in other repos. I'd like to first try importing `clever/components` in another repo before settling for less than ES6.

package.json
```diff
@@ -15,11 +15,21 @@
...
+    "less-loader": "^2.2.3",
     "react": "^0.14.7",
-    "style-loader": "^0.13.0"
+    "react-dom": "^0.14.7",
+    "style-loader": "^0.13.0",
+    "webpack": "^1.12.14",
+    "webpack-dev-server": "^1.14.1"
+  },
+  "scripts": {
+    "dev-server": "node_modules/.bin/webpack-dev-server --watch --host 192.168.50.4"
```
kshen0: What host is this? Is this intentionally hardcoded?

src/Button/Button.jsx
osdiab: make sure it's accessible :3

package.json
rgarcia: this is the VM IP...this is the only way I could get the webpack dev server to load on my host machine's browser :-/

osdiab:
doesn't the spec specify ease-in transitions?

index.html
```diff
@@ -0,0 +1,7 @@
+<!DOCTYPE html>
```
osdiab: If this is just for demo purposes, can we not put it in the project root? Seems counterintuitive to be seeing stuff like this when you first see the project. Same goes for webpack config if it's just for demo purposes.

rgarcia:
@osdiab good catch, will add as a todo

src/Button/Button.less
```diff
@@ -0,0 +1,75 @@
...
+  &:active {
+    background-color: #D33541;
+    color: #FAFCFC;
+    border-bottom: 3px solid #A42933;
+  }
+}
+
+.Button--small {
+  padding: 10px 15px 5px 15px;
+  font-size: 10px semibold;
```
osdiab: Pretty sure these font-size declarations don't work.

index.html
jacobsimon: ideally this would live on the gh-pages branch

rgarcia: I'll give it a shot... so far `webpack-dev-server` has been really frustrating/difficult to get right, so I've tried to keep it simple.

src/Button/Button.less
rgarcia: any suggestions for how to fix them? From what I can tell the user agent stylesheet `font-size` is taking precedence, and I have no idea why.

package.json
osdiab: you should be able to forward your VM ports in such a way that this isn't necessary—but that's not for this PR, so I say we should remove the `--host` argument.

rgarcia: yeah, I tried that (in many different variations) and nothing worked

src/Button/Button.jsx
```diff
@@ -0,0 +1,23 @@
+import React from "react";
+
+require('./Button.less');
+
+class Button extends React.Component {
```
osdiab: I know this is simple but can we have doc comments to explain what these actually are, what props they can take, what state they have, etc?

Additionally, maybe this would be a good time to actually use [React `propTypes`](https://facebook.github.io/react/docs/reusable-components.html).

src/Button/Button.less
eighnjel: i think they need to be separate declarations:
```
font-size: 10px;
font-weight: semibold;
```

src/Button/Button.jsx
rgarcia: @osdiab I already have a TODO to add docs to the README--are you suggesting additional documentation as comments in the code? I'd be worried about keeping these in sync.

Agreed that we should add `propTypes`

src/Button/Button.less
rgarcia: :boom: that worked, thanks @eighnjel!

eighnjel: If you want to do shorthand, you should use `font` instead of `font-size` and need to specify at least font-size and font-family:
```
font: 10px "some-font", semibold;
```

rgarcia:
made some updates here: 272b6ad

- Put most gh-pages related stuff in docs/. There's still a webpack config and package.json in the root of the repo since moving these to sub-folders was a world of pain
- Added a deploy script for publishing the docs/ subfolder to the gh-pages branch
- Added support for buttons-as-links
- `npm run-script dev-server` runs the gh pages site w/ live reload. I also removed hardcoded VM IP (someone should try `npm run-script dev-server` and see if they can load localhost:8080, because I can't)

Still have TODOs listed in the PR description

rgarcia:
cc @cozmo  added a link to the spec in the PR description, and also a TODO to maybe figure out fonts

src/Button/Button.less
cozmo: Note - I did this as part of 8a16bcd

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
...
+      throw new Error("Small destructive buttons are not supported");
+    }
+  }
+  render() {
+    let classes = "Button ";
+    classes += `Button--${this.props.type} `;
+    let size = ["large", "regular", "small"].indexOf(this.props.size) > -1 ? this.props.size : "regular";
+    classes += `Button--${size} `;
+    if (this.props.href == null && !this.props.disabled) {
+      // use <button>s for all disabled buttons and things with no href prop
```
jacobsimon: This is currently using links when disabled instead of buttons

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
+import React from "react";
+
+require('./Button.less');
+
+const VALID_TYPES = ["primary", "secondary", "destructive"];
+const VALID_SIZES = ["large", "regular", "small", "regular"];
```
jacobsimon: regular is listed twice!

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
...
+
+export class Button extends React.Component {
+  constructor(props) {
+    super();
+    if (props.type && VALID_TYPES.indexOf(props.type) === -1) {
```
jacobsimon: We could get most of this validation for free/cheap if we declare propTypes like @osdiab mentioned. For example:
```javascript
Button.propTypes = {
  type: React.PropTypes.oneOf(['primary', 'secondary', 'destructive']).isRequired,
  size: React.PropTypes.oneOf(['large', 'regular', 'small']),
  value: React.PropTypes.string.isRequired,
  disabled: React.PropTypes.bool
}
```

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
...
+  }
+  render() {
+    let classes = "Button ";
+    classes += `Button--${this.props.type} `;
+    let size = ["large", "regular", "small"].indexOf(this.props.size) > -1 ? this.props.size : "regular";
+    classes += `Button--${size} `;
+    if (this.props.href == null && !this.props.disabled) {
+      // use <button>s for all disabled buttons and things with no href prop
+      return (
+        <button className={classes} onClick={this.props.onClick} disabled={this.props.disabled || false}>
```
jacobsimon: We don't need `|| false` if we assume `props.disabled` is bool

src/Button/Button.less
```diff
@@ -0,0 +1,89 @@
...
+      background-color: #D33541;
+      color: #FAFCFC;
+      border-bottom: 3px solid #A42933;
+    }
+  }
+
+  &.Button--small {
+    padding: 10px 15px 5px 15px;
+    font-size: 10px;
+    font-weight: semibold;
```
jacobsimon: don't think this is a valid [font-weight](https://developer.mozilla.org/en-US/docs/Web/CSS/font-weight), probably want `600`

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
...
+          {this.props.value}
+        </button>
+      );
+    } else {
+      const target = this.props.target != null ? this.props.target : "_self";
```
jacobsimon: we _could_ also use `getDefaultProps` for target, disabled, and any other optional params

src/Button/Button.jsx
```diff
@@ -0,0 +1,43 @@
...
+      const onClick = this.props.disabled ? function(e){ e.preventDefault(); } : null;
```
jacobsimon: this works but you could also just remove the href if it's disabled, or use `pointer-events: none` in the styles

package.json
```diff
@@ -15,11 +15,21 @@
...
+    "less-loader": "^2.2.3",
     "react": "^0.14.7",
-    "style-loader": "^0.13.0"
+    "react-dom": "^0.14.7",
+    "style-loader": "^0.13.0",
+    "webpack": "^1.12.14",
+    "webpack-dev-server": "^1.14.1"
+  },
+  "scripts": {
+    "dev-server": "node_modules/.bin/webpack-dev-server --content-base docs/ --inline --watch"
```
jacobsimon: Do we need webpack-dev-server?

src/Button/Button.jsx
rgarcia: cc @cozmo

package.json
rgarcia: `webpack-dev-server --inline` adds a bit of js into the bundle that detects changes and reloads the page for you. So far I've found it a lot better than `webpack --watch` + opening `index.html` + reloading manually.

src/Button/Button.jsx
rgarcia: cc @cozmo

cozmo:
@jacobsimon I pushed some changes (555fab7) that took a lot of your comments into account. I think it's looking pretty good now. Thanks!

@rgarcia and I are going to try to get the fonts working next.

PR #3, opened 2016-03-30
add Modal tests

kshen0:
Adds basic tests for the `Modal` component

cc @rgarcia


test/Modal/Modal_test.js
```diff
@@ -0,0 +1,54 @@
...
+
+  it('calls closeModal when the background is clicked', function() {
+    var stub = sinon.stub()
+    this.renderModal(<Modal title="Title" content={<p>Content</p>} closeModal={stub} />);
+    var background = this.modal.props.children[0];
+    background.props.onClick()
+    this.renderModal();
+    assert(stub.calledOnce);
+  });
+});
```
rgarcia: consider ES6!

rgarcia:
:heart:

test/Modal/Modal_test.js
```diff
@@ -0,0 +1,54 @@
+var assert = require('assert');
```
jacobsimon: Do you think we'll need more than 1 test file per component? If not we probably don't need a separate directory for each one

jacobsimon:
Awesome! We might want to add a [.npmignore](https://docs.npmjs.com/misc/developers#keeping-files-out-of-your-package) to exclude test files from the package

test/Modal/Modal_test.js
kshen0: abc6ec7

kshen0:
`.npmignore` added in b5a5a69

jacobsimon:
lgtm

PR #4, opened 2016-04-05
Easier imports

jacobsimon:
I added Button to `index.js` and changed this file to use the ES6 export syntax. This will simplify the import syntax for consumers -- for example, you can write:
```javascript
import {Button} from "clever-components"; // ES6
```
or
```javascript
var Button = require("clever-components").Button // ES5
```
instead of writing
```javascript
var Button = require("clever-components/src/Button/Button").default
```
Todo: Confirm that this works outside of Chrome.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,2 @@
-module.exports = {
-  Modal: require("./src/Modal/Modal")
-}
+export { Modal } from './src/Modal/Modal';
+export { Button } from './src/Button/Button';
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.1.0",
+  "version": "0.1.1",
   "description": "A library of helpful React components",
   "repository": {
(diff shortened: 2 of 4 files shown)


index.js
```diff
@@ -1,3 +1,2 @@
-module.exports = {
-  Modal: require("./src/Modal/Modal")
-}
+export { default as Modal } from './src/Modal/Modal';
+export { default as Button } from './src/Button/Button';
```
rgarcia: could this be `export Button from './src/Button/Button'`?

rgarcia:
lgtm!

index.js
jacobsimon: https://github.com/Clever/components/pull/4/commits/5e1a73994394dcd8de9671c519aa906dd1c6c571

rgarcia:
@jacobsimon looks like this breaks the tests `Error: Cannot find module '!style!css!less!./Modal.less'`

PR #6, opened 2016-04-06
Remove modal top position

kshen0:
Having both `top` and `margin-top` CSS attributes on the Modal window pushes the modal pretty far down the page. This PR removes the `top` attribute.

### DIFF
--- a/src/Modal/Modal.jsx
+++ b/src/Modal/Modal.jsx
@@ -5,17 +5,22 @@ require("./Modal.less");
 var DEFAULT_WIDTH = 400;
 
 var Modal = React.createClass({
+  propTypes: {
+    closeModal: React.PropTypes.func.isRequired,
+    title: React.PropTypes.string,
+    width: React.PropTypes.number,
+  },
   render: function() {
     var width = this.props.width || DEFAULT_WIDTH;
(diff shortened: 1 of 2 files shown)


src/Modal/Modal.jsx
```diff
@@ -5,17 +5,22 @@ require("./Modal.less");
...
         <div className="Modal--background" onClick={this.props.closeModal}/>
-        <div
-          className="Modal--window"
-          style={{
-            width: width + "px",
-            marginLeft: "-" + (width / 2) + "px"
-          }}>
+        <div className="Modal--window" style={modalStyles}>
           <div className="Modal--window--container">
             <h2>{this.props.title}</h2>
```
andrewhuang: Is it canonical for the title to be rendered by the Modal? I.e. why wouldn't the component consuming the Modal put the header in `children`?

jacobsimon: Could lead to inconsistencies in header size. But yeah there might be cases where we want an icon/graphic next to the title and it would be hard to add. Maybe we could change this to be 
```
{this.props.title && <h2>{this.props.title}</h2>} 
```
so that you could leave it out and supply your own if you want

PR #7, opened 2016-04-06
lint

rgarcia:
cc @jacobsimon


.eslintrc.yml
```diff
@@ -0,0 +1,47 @@
...
+  # vars are necessary for server-side requires right now
+  no-var: 0
+  vars-on-top: 0
+
+  # multi spaces only allowed for aligning variable/import declarations
+  no-multi-spaces:
+    - 2
+    - exceptions:
+        VariableDeclarator: true
+        ImportDeclaration: true
```
kshen0: More of a general question, do you have a strong opinion about whether import declarations should _always_ be aligned?

rgarcia: No strong opinion! I'm not sure there's an eslint rule to enforce that. I'm also still wrapping my head around ES6 imports, so not sure what aligning the different flavors of ES6 imports would look like

.eslintrc.yml
```diff
@@ -0,0 +1,47 @@
...
+    - 2
+    - "never"
+  new-cap:
+    - 2
+    - newIsCapExceptions: ["kayvee.logger"]
+  no-param-reassign:
+    - 2
+    - props: false
+  key-spacing:
+    - 2
```
kshen0: Thoughts on using `off`, `warn`, and `error` keywords here instead of codes?
http://eslint.org/docs/user-guide/configuring#configuring-rules

kshen0:
Drone tests are failing?

.eslintrc.yml
rgarcia: oh nice, didn't know you could do that. I will file it as a TODO since it will also need to be updated across other repos (launchpad, schools-dashboard)

rgarcia:
Yeah, the PR we merged yesterday to change how `less` files are required broke the tests...

kshen0:
LGTM

PR #8, opened 2016-04-06
Improve modal

cozmo:
- [x] Vertically center the modal (regardless of height of content)
- [x] Set the max height of the modal to 80% of the window's height, allow inner content to scroll if it's too tall
- [x] Add a "close" button
- [x] Allow hitting esc to exit the modal. 
- [x] Add the modal to the docs

Some open questions:
- I grabbed the close icon from noun project, is that acceptable design wise? 
- The close icon has no hover/active states yet. 
- The close icon is the first image we're webpacking. Am I doing it correctly? 
- The `maxHeight` calculation in the render function is a little weird, I'd love someone else to take a look at how we do it. 

Full Height

With a scroll bar


docs/docs.jsx
```diff
@@ -2,9 +2,28 @@ import React from 'react';
...
+  openModal(){
+    this.setState({modalIsOpen: true});
+  }
+  closeModal() {
+    this.setState({modalIsOpen: false});
+  }
   render() {
+    var modalElement;
+    if(this.state.modalIsOpen){
+      modalElement = <Modal title="Hello Modal" closeModal={this.closeModal.bind(this)}>
```
jacobsimon: do you need .bind(this)? I haven't done much with es6 classes yet so maybe you do

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
...
-        <div
-          className="Modal--window"
-          style={{
-            width: width + "px",
-            marginLeft: "-" + (width / 2) + "px"
-          }}>
-          <div className="Modal--window--container">
+        <div ref="modalWindow" className="Modal--window" style={style} >
+          <div className="Modal--header">
+            <a className="Modal--close" onClick={this.props.closeModal}></a>
```
jacobsimon: This isn't accessible because `<a>` can't receive focus unless it has `href`. I think you can add `href=""` and it will work. Or just use `button`

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
...
+
+
+  componentDidMount: function() {
+    window.addEventListener('keyup', this.handleKeyUp)
+    window.addEventListener('resize', this.handleResize)
+  },
+
+  componentWillUnmount: function() {
+    window.removeEventListener('keyup', this.handleKeyUp);
+    window.addEventListener('resize', this.handleResize)
```
jacobsimon: nit: semicolons after all these statements

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
 var DEFAULT_WIDTH = 400;
 
 export var Modal = React.createClass({
+  getDefaultProps: function() {
```
jacobsimon: we could do es6 style functions for all of these too `getDefaultProps() {...`

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
...
+    if (e.keyCode == 27){
+      this.props.closeModal();
+    }
+  },
+
+  handleResize: function(e) {
+    this.setState({windowHeight:window.innerHeight})
+  },
+
+
```
jacobsimon: extra whitespace

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
...
+  },
+
   render: function() {
-    var width = this.props.width || DEFAULT_WIDTH;
+    var style = {
+      width: this.props.width + "px",
+      marginLeft: "-" + (this.props.width / 2) + "px",
+    }
+    // The content is 80% of the window height (same as the max height of the modal) less 75px (the height of the header and padding)
+    var contentHeight = this.state.windowHeight*.8-75
```
jacobsimon: This makes sense! Can we make it match the way we set the other styles, for example
```javascript
var contentStyle = {
  maxHeight: (this.state.windowHeight * 0.8 - 75) + "px"
};
```

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
...
       <div className="Modal">
         <div className="Modal--background" onClick={this.props.closeModal}/>
-        <div
-          className="Modal--window"
-          style={{
-            width: width + "px",
-            marginLeft: "-" + (width / 2) + "px"
-          }}>
-          <div className="Modal--window--container">
+        <div ref="modalWindow" className="Modal--window" style={style} >
```
jacobsimon: are you using this ref? we could use it to get the height dynamically

docs/docs.jsx
```diff
@@ -2,9 +2,28 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import {Button} from '../src/Button/Button';
```
jacobsimon: you can change this to '../' as well I think

jacobsimon:
* Probably fine to use it from nounproject assuming it had a free license
* Could we just adjust the opacity for a hover effect?
* I think you're including the icon correctly, but it assumes the client is using url-loader, so we should just make a note of that in the docs. Otherwise it won't load
* Can you update the readme with the prop types?

docs/docs.jsx
```diff
@@ -2,9 +2,28 @@ import React from 'react';
...
+import {Modal} from '../';
 
 class Demo extends React.Component {
+  constructor() {
+    super()
```
rgarcia: constructor takes `props` and `super` should pass `props` as an argument

docs/docs.jsx
```diff
@@ -2,9 +2,28 @@ import React from 'react';
...
+    this.state = {modalIsOpen: false};
+  }
+  openModal(){
+    this.setState({modalIsOpen: true});
+  }
+  closeModal() {
+    this.setState({modalIsOpen: false});
+  }
   render() {
+    var modalElement;
```
rgarcia: s/`var`/`let`/

src/Modal/Modal.jsx
```diff
@@ -5,19 +5,55 @@ require("!style!css!less!./Modal.less");
 var DEFAULT_WIDTH = 400;
 
 export var Modal = React.createClass({
```
rgarcia: this is going to fail lint unless you use an ES6 class

kshen0:
@cozmo what's the status of this?

cozmo:
@kshen0 @jacobsimon @rgarcia Converted modal to ES6, fixed all the lint errors and also fixed other PR comments. Would love another look. 

@kshen0 I commented out the modal tests for now, are they worth fixing or are we changing how we test now?

src/Modal/Modal.less
```diff
@@ -15,18 +17,48 @@
 }
 
 .Modal--window {
+  position: relative;
+  top: 50%;
   left: 50%;
-  margin-top: 60px;
+  transform: translateY(-50%);
+  -webkit-transform: translateY(-50%);
```
jacobsimon: do we need `top: 50%` and `transform`?

kshen0:
@cozmo we're planning to move React tests to [enzyme](https://github.com/airbnb/enzyme) so it seems reasonable to leave tests commented out for now, and address these in a separate PR (front end guild has an action item for the Modal test in particular, unless you want to take a stab at it)

rgarcia:
lgtm

PR #9, opened 2016-04-06
make tests jsx files

kshen0:
* Renames `Modal_test.js` to `Modal_test.jsx` since the tests use jsx syntax ([example](https://github.com/Clever/components/blob/e008d279e18ee528f8acacd08d8e91e37e899898/test/Modal_test.jsx#L10-L14))
* Removes unused line from Makefile


Makefile
```diff
@@ -1,4 +1,4 @@
-JSX_TESTS := $(shell find ./test \! -name "._*.js" -name "*_test.js")
+JSX_TESTS := $(shell find ./test \! -name "._*.jsx" -name "*_test.jsx")
```
rgarcia: it doesn't look like this variable is used in the Makefile? The `test` target appears to invoke mocha on the whole test directory. Can probably just remove this line

PR #10, opened 2016-04-06
Remove inline loaders and fix Modal test

jacobsimon:
The inline loaders added in https://github.com/Clever/components/pull/4 caused new tests to fail. Fixing that revealed an issue in the test file where `this` was undefined and was fixed by changing `() =>` back to `function()`.

### DIFF
--- a/test/Modal_test.js
+++ b/test/Modal_test.js
@@ -1,3 +1,5 @@
+/*eslint func-names: "off"*/
+
 var assert = require("assert");
 var React = require("react");
 var sinon = require("sinon");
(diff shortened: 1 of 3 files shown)


test/Modal_test.js
```diff
@@ -5,7 +5,7 @@ var TestUtils = require("react-addons-test-utils");
 
 import {Modal} from "../";
 
-describe("Modal", () => {
+describe("Modal", function() {
```
rgarcia: `  8:19  warning  Missing function expression name  func-names`

could get around this by moving the function definition outside the call to `describe`. Could also special-case this file by putting a comment at the top: `/*eslint func-names: "off"*/`

jacobsimon: thanks! https://github.com/Clever/components/pull/10/commits/dc32fff94738cbb86dd4781b272980e483ce02f3

PR #11, opened 2016-04-07
add margin to buttons

rgarcia:
@cozmo any reason not to add this by default?


rgarcia:
cc @ipam73 I think you had issues with this

PR #12, opened 2016-04-07
onClick optional

rgarcia:
`React.propTypes.function` seems to make a prop required.

Changing it to `React.propTypes.func` makes it optional, which is correct for `onClick`.


jacobsimon:
👍

PR #13, opened 2016-04-07
Modal extends React.Component

kshen0:
Console warns unless the Modal class extends React.Component


eighnjel:
Are those linting errors supposed to be there / acceptable, the current master build looks green.

eighnjel:
Looks like you need to add Modal.propTypes and make it a stateless function?

kshen0:
Closed in favor of https://github.com/Clever/components/pull/14

PR #15, opened 2016-04-08
Convert to ES5 before publishing

jacobsimon:
We've been seeing some issues across the board when importing and testing components. Most of these problems stem from the way babel-loader and babel-register ignore `node_modules` by default. 

This PR adds a build step prior to publishing that will create an ES5 version of the components in `dist/`. The process of contributing and using this repo remain the same, but there should be significantly less confusion about how to configure webpack.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -10,3 +10,15 @@ MOCHA_JSX := node_modules/mocha/bin/mocha --compilers jsx:babel-register --recur
 
 test: lint
 	NODE_ENV=test $(MOCHA_JSX) $@
+
+BABEL := node_modules/babel-cli/bin/babel.js
+
+build:
+	@rm -rf dist
+	@echo '✓ Clean out dist directory'
+	@cp -r src dist
+	@echo '✓ Copy /src to /dist'
+	@$(BABEL) dist -d dist > ._babel_tmp && rm ._babel_tmp
+	@echo '✓ Convert ES6 to ES5'
+	@find ./dist -name "*.jsx" | xargs -d"\n" rm
+	@echo '✓ Remove JSX files'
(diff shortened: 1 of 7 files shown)


eighnjel:
update readme? https://github.com/Clever/components/blob/master/README.md#how-to-add-a-component

graveyard:
@jacobsimon have you tested this with schools dashboard and launchpad? Does it require targeting `js` files with babel-loader? This currently causes some issues in the schools dashboard. I want to make sure we don't end up playing a game of whack-a-mole here with webpack configs

rgarcia:
@jacobsimon have you tested this with schools dashboard and launchpad? Does it require targeting js files with babel-loader? This currently causes some issues in the schools dashboard. I want to make sure we don't end up playing a game of whack-a-mole here with webpack configs

jacobsimon:
@rgarcia This should fix everything. But you can continue to import directly using `require("clever-components/src/Modal/Modal").Modal` 😃

Makefile
```diff
@@ -10,3 +10,15 @@ MOCHA_JSX := node_modules/mocha/bin/mocha --compilers jsx:babel-register --recur
...
+
+build:
+	@rm -rf dist
+	@echo '✓ Clean out dist directory'
+	@cp -r src dist
+	@echo '✓ Copy /src to /dist'
+	@$(BABEL) dist -d dist > ._babel_tmp && rm ._babel_tmp
+	@echo '✓ Convert ES6 to ES5'
+	@find ./dist -name "*.jsx" | xargs -d"\n" rm
+	@echo '✓ Remove JSX files'
```
kshen0: Can this be simplified to avoid copying the src files over and later deleting them?

```
rm -rf dist && mkdir dist
$(BABEL) src -d dist
```

kshen0: Discussed offline; less files need to be copied too

kshen0:
LGTM

PR #17, opened 2016-04-14
Add Button type options and custom styles

jacobsimon:
This adds a "link" button type that's useful for patterns where we have a primary button with a secondary action that doesn't need to be a button like this:

screen shot 2016-04-14 at 3 16 07 pm

This also adds `type="button"` to all buttons so that they don't automatically act as submit buttons in forms. Because we already use `type`, then if you do want `type="submit"` you can set `submit` and it will handle it:

```jsx
<Button value="Submit" type="primary" onClick={handleClick} submit /> 
```

Additionally, support for custom styles is now allowed for convenience until we come to a consensus on spacing and style guidelines: 
```jsx
<Button style={{marginRight: 10}} ...  />
```

Also updated the readme to include all of this

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,6 +47,31 @@ var myCloseHandler = function() {
 }
 ```
 
+### Button
+
+This is a set of button components with various sizes and types.
+
+**Options**
+
+| Prop | Type | Description | Default |
+|----|----|----|----|
+| value | String | The text that appears on the button | None
+| type | String | One of `primary`, `secondary`, `destructive`, `link` | `secondary`
+| size | String | One of "large", "regular", "small" | `regular`
+| onClick (optional) | Function | Called when the user clicks on the button | None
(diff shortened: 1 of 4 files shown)


osdiab:
Do we want to support `reset` buttons?

osdiab:
LGTM

README.md
```diff
@@ -47,6 +47,31 @@ var myCloseHandler = function() {
...
 ```
 
+### Button
+
+This is a set of button components with various sizes and types.
+
+**Options**
+
+| Prop | Type | Description | Default |
+|----|----|----|----|
```
osdiab: Is this supposed to be aligned?

jacobsimon: Fixed! https://github.com/Clever/components/pull/17/commits/5f636abfdacef2b6eba783d2e03e515ddb6c214b

PR #19, opened 2016-04-20
Fix spacing in modal

jacobsimon:
Makes the max-height taller and fixes an overflow issue caused by the header size being slightly less than it should have been given it was using `border-box`. Tested in chrome and firefox.


kshen0:
LGTM

PR #21, opened 2016-04-23
Confirmation button

osdiab:
Shows a confirmation modal before acting.

### DIFF
--- a/src/ConfirmationButton/ConfirmationButton.jsx
+++ b/src/ConfirmationButton/ConfirmationButton.jsx
@@ -1,0 +1,52 @@
+import objectAssign from "object-assign";
+import React, {PropTypes} from "react";
+
+import {Button, ModalButton} from "..";
+import {propsFor, prefixKeys, unprefixKeys} from "../utils";
+
+require("./ConfirmationButton.less");
+
+export class ConfirmationButton extends React.Component {
+  constructor(props) {
+    super(props);
+    this.handleConfirm = this.handleConfirm.bind(this);
+    this.handleCancel = this.handleCancel.bind(this);
+  }
+
+  handleConfirm() {
+    this.refs.modalButton.hideModal();
+    if (this.props.onConfirm) this.props.onConfirm();
+  }
+
+  handleCancel() {
+    this.refs.modalButton.hideModal();
+    if (this.props.onClose) this.props.onClose();
+  }
+
+  render() {
+    const modalButtonProps = propsFor(ModalButton, this.props);
+    const confirmButtonProps = propsFor(Button, unprefixKeys(this.props, "confirmButton"));
+    return (<ModalButton {...modalButtonProps} ref="modalButton">
+      {this.props.children}
+      <div className="ConfirmationButton--dialog-buttons">
+        <Button type="secondary" value="Cancel" onClick={this.handleCancel} />
+        <Button {...confirmButtonProps} onClick={this.handleConfirm} />
+      </div>
+    </ModalButton>);
+  }
+}
+
+ConfirmationButton.propTypes = objectAssign(
+  {}, prefixKeys(Button.propTypes, "confirmButton"), ModalButton.propTypes, {
+    confirmButtonType: Button.propTypes.type,
+    confirmButtonText: PropTypes.string,
+    onConfirm: PropTypes.func,
+  }
+);
+
+ConfirmationButton.defaultProps = objectAssign(
+  {}, prefixKeys(Button.defaultProps, "confirmButton"), ModalButton.defaultProps, {
+    confirmButtonType: "primary",
+    confirmButtonValue: "Confirm",
+  }
+);
--- a/test/ConfirmationButton_test.jsx
+++ b/test/ConfirmationButton_test.jsx
@@ -1,0 +1,53 @@
+/* eslint func-names: "off" */
+
+import assert from "assert";
+import React from "react";
+import sinon from "sinon";
+import {mount} from "enzyme";
+import {ModalButton, ConfirmationButton} from "../src";
+
+describe("ConfirmationButton", () => {
+  it("renders a modal with confirm and cancel buttons", () => {
+    const confirmationButton = mount(
+      <ConfirmationButton value="Confirm pl0x" modalTitle="A title">
+        <p>Hello</p>
+      </ConfirmationButton>
+    );
+    assert.equal(confirmationButton.find(ModalButton).length, 1);
+
+    confirmationButton.find("button").simulate("click"); // render the modal
+    const dialog_buttons = confirmationButton.find(".ConfirmationButton--dialog-buttons");
+    assert.equal(dialog_buttons.length, 1);
+    assert.equal(dialog_buttons.find("button").length, 2);
+    assert.equal(dialog_buttons.find("button").first().text(), "Cancel");
+    assert.equal(dialog_buttons.find("button").last().text(), "Confirm");
+  });
+
+  it("calls onClose if present when cancel button clicked", () => {
+    const onCloseSpy = sinon.spy();
+    const confirmationButton = mount(<ConfirmationButton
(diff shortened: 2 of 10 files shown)


src/ConfirmationButton/ConfirmationButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+  toggleConfirmation(show) {
+    this.setState({showConfirmation: show});
+  }
+
+  render() {
+    const modalContents = (<div>
+      <p>{this.props.confirmationDescription}</p>
+      <p>
+        <Button type="secondary" value="Cancel"
+          onClick={_.partial(this.toggleConfirmation, false)}
```
jacobsimon: The partial application syntax is kind of confusing tbh, how do you feel about:
* es6 syntax: `onClick={() => this.toggleConfirmation(false)}` (this is what they do in the react docs)
* explicit handlers like `showConfirmation` and `hideConfirmation`
* just actually toggle state of taking an arg `setState({show: !this.state.show})`

src/ConfirmationButton/ConfirmationButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+        />
+        <Button type={this.props.confirmButtonType} value={this.props.confirmButtonText}
+          onClick={_.compose(_.partial(this.toggleConfirmation, false), this.props.onConfirm)}
```
jacobsimon: see comment above

src/ConfirmationButton/ConfirmationButton.jsx
```diff
@@ -0,0 +1,59 @@
...
+        />
+      </p>
+    </div>);
+
+    // don't pass properties not for buttons, nor onClick, to the underlying Button to display.
+    const buttonProps = _.omit(_.pick(this.props, _.keys(Button.propTypes)), "onClick");
+
+    return (<div className="ConfirmationButton">
+      <Button {...buttonProps} onClick={_.partial(this.toggleConfirmation, true)} />
+      {this.state.showConfirmation ? <Modal title="Confirm regenerating badges"
```
jacobsimon: this should be `title={this.props.confirmationHeader}`

src/ConfirmationButton/ConfirmationButton.jsx
```diff
@@ -0,0 +1,59 @@
+import React, {Component, PropTypes} from "react";
+import _ from "underscore";
+
+import {Button} from "../Button/Button";
```
jacobsimon: haven't tried this but can you just do `import {Button} from ".."`

jacobsimon:
Two thoughts: 
* There are cases where you want to have a confirmation modal, but you might not want/need a Button to trigger it (you're leaving the page, you're clicking a link or dropdown option, etc.). Maybe it would be better to create a ConfirmationDialog component and have ConfirmationButton consume that? 
* We should double check what happens when the client is using a different version of underscore, and use the webpack de-dupe plugin

src/ConfirmationButton/ConfirmationButton.jsx
osdiab: 929b690

src/ConfirmationButton/ConfirmationButton.jsx
osdiab: 929b690

src/ConfirmationButton/ConfirmationButton.jsx
osdiab: 4b8399d

osdiab:
With respect to underscore, there exist ES6 features that do the same thing; if we add in the `Object.assign` polyfill since node doesn't yet support that, then we don't need it for `_.extend`; and I can replace `_.omit` and `_.pick` by using `reduce`.

src/ConfirmationButton/ConfirmationButton.jsx
osdiab: ac666a6

osdiab:
I will modify this to be two components: `ModalButton` and `ConfirmationButton`.

`ModalButton` will be a button that opens and closes a modal. `ConfirmationButton` will be a special instance of `ModalButton` that presents a dialog for confirmation.

osdiab:
Done! I modified the approach so that rather than trying to rename the properties on Button and Modal, it just subsumes them; if this is a more confusing API I can do the propType omission/readdition tricks I did before.

Not too shabby!

osdiab:
Errors about bad `propTypes` are fixed in #22

jacobsimon:
I like it! And agree about closeModal/onClose. It would be really cool later on to generalize the modal handling logic as a higher order component that could apply to any component (custom or native) that you want to trigger a modal, e.g. `var ModalButton = ModalHandler(Button)`

osdiab:
That would be cool—I think we'd need some kind of standardized hook to do that composition, which is something we can think about next time this sort of behavior comes up.

osdiab:
I think another potentially valuable generalization would be to allow for more than one "action" button in the `ConfirmationButton`. Do you think that's valuable as part of this component, or can be made a more general one with this as a special, common case?

osdiab:
Check this latest edit @jacobsimon; I allow for the full breadth of `Button` options in `ConfirmationButton` by automatically prefixing `Button`'s props with `"confirmButton"`; and prefix `Modal` options in `ModalButton` with `"modal"`. Is this useful/does this make sense? I think it does!

jacobsimon:
LGTM!

osdiab:
@jacobsimon lookover the test style and the stylesheets pl0x! If good then i'll merge.

package.json
```diff
@@ -36,6 +36,7 @@
     "less": "^2.6.1",
     "less-loader": "^2.2.3",
     "mocha": "^2.4.5",
+    "object-assign": "^4.0.1",
```
jacobsimon: do we still need this since we have the babel plugin?

test/ConfirmationButton_test.jsx
```diff
@@ -0,0 +1,52 @@
...
+import React from "react";
+import sinon from "sinon";
+import {mount, shallow} from "enzyme";
+import {Button, ModalButton, ConfirmationButton} from "../src";
+
+describe("ConfirmationButton", () => {
+  it("renders a modal with confirm and cancel buttons", () => {
+    const confirmationButton = shallow(<ConfirmationButton value="Confirm pl0x" modalTitle="A title">
+      <p>Hello</p>
+    </ConfirmationButton>);
```
jacobsimon: nit: we've been doing indentation in other tests like this https://github.com/Clever/components/blob/master/test/Modal_test.jsx#L9-L13
```
shallow(
  <ConfirmationButton> 
   <p>Hello</p>
  </ConfirmationButton>
); 
```

test/ConfirmationButton_test.jsx
```diff
@@ -0,0 +1,52 @@
...
+    assert.equal(confirmationButton.find(ModalButton).length, 1);
+    assert(!!confirmationButton.find(ModalButton).props().children.find(
+      (child) => {
```
jacobsimon: could we express this test with selectors like you do in later tests? for instance you could do something like
```js
confirmationButton.find("button").simulate("click");
const cancelButton = confirmationButton.find("button.ConfirmationButton--dialog-buttons[value='Cancel'])");
const confirmButton = confirmationButton.find("button.ConfirmationButton--dialog-buttons[value='Confirm'])");
assert(cancelButton.length, 1);
assert(confirmButton.length, 1);
```

osdiab: 0856ed6

test/ConfirmationButton_test.jsx
osdiab: 0856ed6

package.json
osdiab: 0856ed6

osdiab:
@jacobsimon last look?

jacobsimon:
looks good!

PR #22, opened 2016-04-24
fix propType errors with single elements

osdiab:
When a single element is used in a Modal, it would complain that you didn't supply an Array; node type allows for single or multiple children elements.

Docs: https://facebook.github.io/react/docs/reusable-components.html


jacobsimon:
👍

PR #23, opened 2016-04-25
Add helper script and update readme

jacobsimon:
Wrote a helper script called `new_component` that automatically generates helpful stubs for your component. For example, running
```bash
./bin/new_component FooBar
```
would create FooBar.jsx, FooBar.less, and FooBar_test.jsx and also export `FooBar` in index.js for you. I also added a `make clean` command and a command to run tests for a specific component.

### DIFF
--- a/bin/example_component.jsx.txt
+++ b/bin/example_component.jsx.txt
@@ -2,6 +2,12 @@ import React from "react";
 
 import "./COMPONENT_NAME.less";
 
+/*
+  Add a description of your component here.
+
+  If your component can be expressed without state,
+  use the stateless function syntax instead!
+*/
 export default class COMPONENT_NAME extends React.Component {
   constructor(props) {
     super(props);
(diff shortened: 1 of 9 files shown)


osdiab:
Can we add `Object.assign` transform to this? It will let me not introduce the polyfill lib and have to use a different syntax in #21. https://babeljs.io/docs/plugins/transform-object-assign/

bin/example_component.jsx.txt
```diff
@@ -0,0 +1,21 @@
+import React from "react";
+
+import "./COMPONENT_NAME.less";
```
osdiab: Do we ever expect to produce stateless components? In that case the function syntax is cleaner, but this does seem like it would be the common case; maybe leave a comment to say that?

Also, should we be leaving doc comments on the classes? Especially if my roommate Prakash's component explorer supports extracting those.

osdiab:
Nice, I like it! LGTM

bin/example_component.jsx.txt
jacobsimon: Agreed, I'll add a comment about stateless components. I'm not sure what the doc format should be (jsdoc?) so I'll leave that til we know more

PR #24, opened 2016-05-03
use Object.assign instead of objectAssign

osdiab:


jacobsimon:
looks good, problem I was having seems to have disappeared

PR #25, opened 2016-05-03
Update modal styles

jacobsimon:
We are updating the look of the modal to match @mischelle-m's new designs. This PR also adds a fix for the issue where a scrollbar appears unnecessarily, and it fixes the error messages that would appear on resize due to improper removal of event handlers.

screen shot 2016-07-20 at 2 47 54 pm

I also updated the styles so that if you put `Button` in a `footer` inside of the Modal, it will automatically align the buttons to the right and set the margins correctly. This means you won't have to do any custom styling of the buttons anymore if you follow this pattern:

```
<Modal ...>
  <footer>
    <Button ... />
  </footer>
</Modal>
```

To do:

- [ ] Cross browser testing (IE 11, IE 10, Safari, Firefox, Chrome)
- [x] Reduce size of close icon


src/Modal/Modal.jsx
```diff
@@ -15,11 +15,12 @@ export class Modal extends React.Component {
   componentDidMount() {
     window.addEventListener("keyup", this.handleKeyUp);
     window.addEventListener("resize", this.handleResize);
+    this.handleResize();
   }
 
   componentWillUnmount() {
     window.removeEventListener("keyup", this.handleKeyUp);
-    window.addEventListener("resize", this.handleResize);
+    window.removeEventListener("resize", this.handleResize);
```
cozmo: womp

src/Modal/Modal.jsx
```diff
@@ -29,27 +30,33 @@ export class Modal extends React.Component {
...
-        <div className="Modal--window" style={windowStyle} >
-          <div className="Modal--header">
+        <div className="Modal--window" style={windowStyle}>
+          <header className="Modal--header">
             <button className="Modal--close" onClick={this.props.closeModal} />
             <h2>{this.props.title}</h2>
-          </div>
-          <div style={contentStyle} className="Modal--window--content">
+          </header>
+          <div style={contentStyle} className="Modal--window--content" ref={(ref) => {this.modal = ref;}}>
```
kshen0: Why do we need to set `ref`?

src/Modal/Modal.less
```diff
@@ -23,42 +22,71 @@
...
   box-sizing: border-box;
 
   h2{
     padding: 0;
     margin: 0;
+    font-size: 22px;
+    line-height: 30px;
+    font-weight: 500;
+    color: #121212;
+    font-family: "Proxima Nova";
```
kshen0: Can we fall back to Arial and sans-serif?

kshen0:
left you some comments

src/Modal/Modal.jsx
jacobsimon: using it to compare the modal's scrollHeight to the window's height, so we can enable scrolling when it's needed. this is the best way I could think of to fix the issue where the scrollbar always showed up

src/Modal/Modal.less
jacobsimon: https://github.com/Clever/components/pull/25/commits/cc356997acdcb96d16730a35d8059123939676bf

kshen0:
LGTM

PR #26, opened 2016-05-05
add babel-polyfill to build process

osdiab:
Allows ES6 methods like `String.startsWith` and `Object.assign`

docs: https://babeljs.io/docs/usage/polyfill/

Either this or #27. If we shouldn't use Babel polyfill for a library, then #27 is the one to do; but this catches all ES6 functionality, whereas #27 we'd have to be careful not to include any ES6 functions on builtin prototypes, which can be annoying

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -3,7 +3,4 @@
     "es2015",
     "react"
   ],
-  "plugins": [
-    "transform-object-assign"
-  ]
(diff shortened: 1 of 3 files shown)


osdiab:
oh this is probably bad since it's intended for applications, not libraries

PR #27, opened 2016-05-05
use indexOf instead of startsWith

osdiab:
We should come up with a way to know that we're not using any ES6 features that we can't if we take this route...

Here is a list of existent transforms for babel (don't think `startsWith` is one of them): https://babeljs.io/docs/plugins/#transform-plugins

Either this or #26


jacobsimon:
LGTM. I'm in favor of erring on the side of fewer dependencies and more browser compatibility

PR #30, opened 2016-07-13
Support className prop; Fix inconsistent heights

kofi-clever:
- Noticed while using Button that it doesn't support the className prop. Adding
  support for additional class names to be specified.

- There's also a 1px difference in the button heights that I can't unsee ;P
  Due to the fact that the secondary button has a border and the others don't.
button-heights

### DIFF
--- a/src/Button/Button.less
+++ b/src/Button/Button.less
@@ -21,6 +21,7 @@ a, button {
       color: #4B4B59;
       background-color: #D7D9D9;
       border: none;
+      border: 1px solid #D7D9D9;
       border-bottom: 3px solid #A4A6A6;
     }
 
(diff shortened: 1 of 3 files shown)


src/Button/Button.less
```diff
@@ -21,6 +21,7 @@ a, button {
       color: #4B4B59;
       background-color: #D7D9D9;
       border: none;
+      border: 1px solid #D7D9D9;
```
kofi-clever: not tied to this specific solution. we can use hard heights + border-box sizing for the buttons instead, if anyone feels strongly

nathanleiby:
classname prop seems reasonable. 

deferring to others for suggestions on the CSS

jacobsimon:
lgtm!

PR #31, opened 2016-07-14
Fix height text input

ipam73:
Weird issue, sometimes height doesn't work properly in chrome.  Adding task on frontend guild to fix at a later time.  This is a workaround that seems to solve the inconsistency issue.


jacobsimon:
lgtm!

PR #32, opened 2016-07-26
Add CopyableInput component

jacobsimon:
This adds a new CopyableInput component that allows the user to copy the content of the input to the clipboard. Under the hood it uses react-copy-to-clipboard. This also adds a show/hide button to TextInput.

To do:
- [x] Hide the default clear/reveal icons in IE and Edge with ms-clear and ms-reveal 
- [x] change the style of the readonly TextInput to have a light background and border.

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,6 +192,7 @@ This component is a `text input`.
 | required (optional) | Bool   | Marks input as required and adds indicator. | false
 | disabled (optional) | Bool   | Sets element as disabled. | false
 | readOnly (optional) | Bool   | Sets element as read only. | false
+| enableShow (optional) | Bool  | Displays a show/hide link that reveals passwords | false
 
 **Usage Example**
 
@@ -229,5 +230,4 @@ This is a special [TextInput](#textinput) that allows the user to show/hide the 
 
 | Prop             | Type     | Description                           | Default
 |------------------|----------|---------------------------------------|---------
-| enableShow | Bool | Display a Show/Hide toggle link | False
-| enableCopy | Bool | Display a Copy link | False
+| enableCopy (optional) | Bool | Display a Copy link | True
--- a/README.md
+++ b/README.md
@@ -221,3 +221,13 @@ var onChangeText = function(event) {
 }
 ```
 
+### CopyableInput
+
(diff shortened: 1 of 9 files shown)


ipam73:
did you think about adding an example to docs?

jacobsimon:
Yeah I'll do that and add a test as well

src/CopyableInput/CopyableInput.jsx
```diff
@@ -0,0 +1,66 @@
...
+    this.setState({hidden: !this.state.hidden});
+  }
+
+  copyPassword(e) {
+    e.preventDefault();
+    this.setState({copied: true});
+  }
+
+  render() {
+    const type = this.props.type === "password" && this.state.hidden ? "password" : "text";
```
ipam73: 👍

src/CopyableInput/CopyableInput.jsx
```diff
@@ -0,0 +1,66 @@
...
+import * as CopyToClipboard from "react-copy-to-clipboard";
+
+import "./CopyableInput.less";
+
+/**
+ * This is a text input that takes optional props
+ * enableShow and enableCopy that allow the user
+ * to show/hide and copy the value of the input.
+ */
+export class CopyableInput extends React.Component {
```
ipam73: Nit: The classname doesn't tell me much about the show/hide password.  It seems to me that there's two things here:  1) can you show/hide a password, and 2) can you enable copy.

ipam73:
This is looking great!  

Thinking about this now, it seems to me that there's 2 things going on:
1) add show/hide to passwords
2) enable copy input

Did you consider adding the show/hide password to the regular text input component?  I imagine that all password inputs should have this functionality.  Then keep the copy input here?  I also imagine that any component regardless of whether it is a password or not might use the copy function.

jacobsimon:
I've changed this as we discussed so that TextInput also supports the show/hide feature, and CopyableInput just re-implements that feature on top to prevent style collisions. Let me know what you think!

ipam73:
LGTM!  🚢

README.md
```diff
@@ -221,3 +221,13 @@ var onChangeText = function(event) {
 }
 ```
 
+### CopyableInput
```
ipam73: oh wait, maybe add `enableShow` to readme for text input too?

PR #33, opened 2016-07-27
make width in input component follow parent component

ipam73:
Context:  Instead of hard-coding width, make width respect parent component width.
Changes:  Make width 100%, wrap docs w/parent div at 300px


osdiab:
LGTM! whitespace diff ftw

PR #34, opened 2016-07-31
SegmentedControl

carolinemodic:
Adds a new SegmentedControl component to components library. 
jul-31-2016 14-04-39

### DIFF
--- a/README.md
+++ b/README.md
@@ -231,3 +231,39 @@ This is a special [TextInput](#textinput) that allows the user to show/hide the 
 | Prop             | Type     | Description                           | Default
 |------------------|----------|---------------------------------------|---------
 | enableCopy (optional) | Bool | Display a Copy link | True
+
+### SegmentedControl
+
+This component is a `segmented control`.
+
+**Options**
+
+| Prop             | Type     | Description                           | Default
+|------------------|----------|---------------------------------------|---------
+| selectableItems (required) | Object   | Map from keys => Display strings for each item in control. | None
+| defaultValue (optional)    | String   | Key of the selectableItem to initalize as selected, defaults to null | None
+| onSelect (optional)        | Function | Called with key of selected item on select | None
+
+**Usage Example**
+
+```jsx
+  <SegmentedControl
+    selectableItems={{one: "Option 1", two: "Option 2"}}
+    defaultValue="two"
+    onSelect={this.onSelect}
+  />
+```
+
+In this example, the function `onSelect` is called on component value change and updates the state of your component:
+
+```javascript
+var onSelect = function(selected) {
+  // do some validation
+  var errorMessage = "";
+  if (notValid) errorMessage = "Please enter a valid value";
+  this.setState({
+    selected: selected,
+    error: errorMessage,
+  });
+}
+```
--- a/README.md
+++ b/README.md
@@ -241,7 +241,7 @@ This component is a `segmented control`.
 | Prop             | Type     | Description                           | Default
 |------------------|----------|---------------------------------------|---------
 | selectableItems (required) | Object   | Map from keys => Display strings for each item in control. | None
-| defaultValue (optional)    | String   | Key of the selectableItem to initalize as selected, defaults to null | None
+| selected (optional)    | String   | Key of the selectableItem to initalize as selected, defaults to null | None
 | onSelect (optional)        | Function | Called with key of selected item on select | None
 
 **Usage Example**
@@ -249,7 +249,7 @@ This component is a `segmented control`.
 ```jsx
   <SegmentedControl
     selectableItems={{one: "Option 1", two: "Option 2"}}
-    defaultValue="two"
+    selected="two"
     onSelect={this.onSelect}
   />
 ```
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # Clever Front End Components
 
-**Jump to** [Modal](#modal), [Button](#button), [ModalButton](#modalbutton), [ConfirmationButton](#confirmationbutton), [TextInput](#textinput)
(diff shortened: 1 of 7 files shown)


src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -0,0 +1,59 @@
...
+  constructor(props) {
+    super(props);
+    this.state = {selected: props.defaultValue};
+    this.onSelect = this.onSelect.bind(this);
+  }
+
+  onSelect(e) {
+    const value = e.target.getAttribute("value");
+    this.setState({selected: value});
+    if (this.props.onSelect !== undefined) {
```
kofi-clever: safer to use `if (this.props.onSelect) {...}` since `this.props.onSelect = null` would pass this check

src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+  render() {
+    const cssClass = SegmentedControl.cssClass;
+    const selectableItems = lodash.map(this.props.selectableItems, (name, value) => {
+      const isSelected = value === this.state.selected;
+      const selectedClass = isSelected ? "selected" : "";
+      return (
+        <span
+          className={`${cssClass.SELECTABLE_ITEM} ${selectedClass}`}
+          onClick={this.onSelect}
```
kofi-clever: nit: no need to bind in the constructor if you use an arrow function (preferred es6 pattern)
```js
onClick={e => this.onSelect(e)}
```

src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -0,0 +1,59 @@
...
+          key={value}
```
kofi-clever: consider using a combination of `value` and the index, since we don't (and probably shouldn't) require that `value` is unique for each option.

src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+/*
+  SegmentedControl selection. Shows several options
+  as defined in the props and allows the user to select
+  one of those options.
+*/
+export class SegmentedControl extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = {selected: props.defaultValue};
```
kofi-clever: react adopts a specific pattern with form controls that we might want to follow here (https://facebook.github.io/react/docs/forms.html):
`defaultValue` is only used if an explicit value hasn't been selected.

i think it's also advised not to initialize state from props, since props can change and go out of sync with the state.

in this case, you could defer the defaulting logic to the `render()` method instead and leave `state.selected = null` until it's explicitly set

carolinemodic: ahh is selected the right name for this prop then?

README.md
```diff
@@ -231,3 +231,39 @@ This is a special [TextInput](#textinput) that allows the user to show/hide the
 | Prop             | Type     | Description                           | Default
 |------------------|----------|---------------------------------------|---------
 | enableCopy (optional) | Bool | Display a Copy link | True
+
+### SegmentedControl
```
kofi-clever: nit add anchor link to this section at the top of the readme

package.json
```diff
@@ -47,7 +47,9 @@
     "webpack-dev-server": "^1.14.1"
   },
   "scripts": {
-    "dev-server": "node_modules/.bin/webpack-dev-server --content-base docs/ --inline --watch"
+    "dev-server": "node_modules/.bin/webpack-dev-server --content-base docs/ --inline --watch --host 192.168.50.4"
```
kofi-clever: should this be `localhost`? or is this just for testing?

src/SegmentedControl/SegmentedControl.less
```diff
@@ -0,0 +1,27 @@
...
+    &.selected {
+      background-color: #4274F6;
+      color: #FAFAFC;
+    }
+  }
+  border: 1px solid #E3E6EB;
+  border-radius: 3px;
+}
+
+.SegmentedControl:first-child {
```
kofi-clever: nit: move under `.SegmentedControl` as `&:first-child`

also, should this be under `.segmented_control--selectable-item` instead?
`:first-child` affects the component it's assigned to, not its children

src/SegmentedControl/SegmentedControl.jsx
kofi-clever: i think `defaultValue` and `value` are appropriate here - similar to what react uses for the `<select>` element.

`defaultValue` should be applied if the control's state hasn't been modified by the actual user and `value` should always override the visual state, regardless of what the user previously selected.

`value` would have to be used with `onChange` to keep the control's state in sync with the parent (i think react locks the control if a `value` is specified without an `onChange`

simpler to just support `defaultValue`, IMO, since `value` is more involved - can always add that later if there's a need

package.json
carolinemodic: just for testing, forgot to remove

carolinemodic:
@kofi-clever Addressed PR comments, I left out implementing value for now as you suggested.

src/SegmentedControl/SegmentedControl.less
```diff
@@ -0,0 +1,27 @@
...
+      background-color: #4274F6;
+      color: #FAFAFC;
+    }
+    &:first-child {
+      border-left: none;
+    }
+  }
+  border: 1px solid #E3E6EB;
+  border-radius: 3px;
+}
```
kofi-clever: nit: a little tough to visually parse these rules.
i'd recommend:
- write all `.SegmentedControl` rules first before sub-component rulesets
- add an empty line before nested rulesets 

optional uber-nit: alpha for the rules in each block, since it's a lot easier to find a specific rule when scanning the list

kofi-clever:
optional nit - LGTM!
excited for this component

PR #35, opened 2016-08-02
Move react-copy-to-clipboard to dependencies

jacobsimon:
I'm seeing an issue where I need to install react-copy-to-clipboard in a different repo to get it to build. I think moving this from devDependencies to normal dependencies will let it install correctly.


eighnjel:
:shipit:

PR #36, opened 2016-08-05
Update the SegmentedControl to support more use cases.

kofi-clever:
Update the SegmentedControl to support more use cases.

Ran into a few corner cases when using SegmentedControl in context in the
Schools Dashboard.

- Reducing font size to better fit with typical text sizes.
- Polishing up styling issues.
- Adding a disabled state.
- Updating hover state cursors.

seg-ctrl

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.5.1",
+  "version": "0.6.0",
   "description": "A library of helpful React components",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -49,6 +49,7 @@
     "dev-server": "node_modules/.bin/webpack-dev-server --content-base docs/ --inline --watch"
   },
   "dependencies": {
+    "classnames": "^2.2.5",
     "lodash": "^4.14.1",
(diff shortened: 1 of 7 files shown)


package.json
```diff
@@ -49,6 +49,7 @@
     "dev-server": "node_modules/.bin/webpack-dev-server --content-base docs/ --inline --watch"
   },
   "dependencies": {
+    "classnames": "^2.2.5",
```
jacobsimon: Do we need this

src/SegmentedControl/SegmentedControl.less
```diff
@@ -1,29 +1,63 @@
+@import "../colors.less";
```
jacobsimon: 💯

jacobsimon:
Looks great

package.json
kofi-clever: been using it for convenience and for cleaning up code related to conditional classes, but we can do without it for now.
worth considering, though. it's relatively small (~1K unminified).

stripped it out, along with some of the CSS colors we're not using here yet - https://github.com/Clever/components/pull/36/commits/06335a20e202543f9504af8c0f4faf6f73376902

PR #37, opened 2016-08-05
Fix disabled link button styling.

kofi-clever:
Remove background and borders from link buttons in the disabled state.

**Before:**
screen shot 2016-08-04 at 9 24 56 pm

**After:**
screen shot 2016-08-04 at 9 24 38 pm


jacobsimon:
LGTM, thanks!

PR #38, opened 2016-08-10
Enable usage of SegmentedControl as a controlled component.

kofi-clever:
Ran into a complex use-case that now requires this element to support controlled mode.
[More info: https://fb.me/react-controlled-components]

- Add a `value` prop that overrides the selected value.
- Requires `onSelect` to be specified; throws an error otherwise.
- Also requires only one of `value` or `defaultValue` to be specified.

Bonus: s/span/button/ so that the options get tabIndex and the button ARIA role `#accessibility`

### DIFF
--- a/src/SegmentedControl/SegmentedControl.jsx
+++ b/src/SegmentedControl/SegmentedControl.jsx
@@ -22,7 +22,7 @@ export class SegmentedControl extends React.Component {
     if (value && !onSelect) {
       throw new Error(
         "`onSelect` prop required when using SegmentedControl as a controlled component. "
-        + "Either provide the `onChange` prop, or replace `value` with `defaultValue` for an "
+        + "Either provide the `onSelect` prop, or replace `value` with `defaultValue` for an "
         + "uncontrolled SegmentedControl.\n"
         + "More info: https://fb.me/react-controlled-components"
       );
@@ -46,11 +46,7 @@ export class SegmentedControl extends React.Component {
       return;
     }
 
-    // Update internal state only if being used as an uncontrolled component.
-    if (!this.props.value) {
-      this.setState({selected: value});
-    }
-
+    this.setState({selected: value});
     if (this.props.onSelect) {
       this.props.onSelect(value);
     }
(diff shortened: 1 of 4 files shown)


src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -8,8 +8,32 @@ import "./SegmentedControl.less";
...
+        "SegmentedControl must either be controlled or uncontrolled "
+        + "(specify either the `value` prop, or the `defaultValue` prop, but not both).\n"
+        + "More info: https://fb.me/react-controlled-components"
+      );
+    }
+
+    if (value && !onSelect) {
+      throw new Error(
+        "`onSelect` prop required when using SegmentedControl as a controlled component. "
+        + "Either provide the `onChange` prop, or replace `value` with `defaultValue` for an "
```
rgarcia: s/onChange/onSelect/?

src/SegmentedControl/SegmentedControl.jsx
```diff
@@ -22,19 +46,23 @@ export class SegmentedControl extends React.Component {
       return;
     }
 
-    this.setState({selected: value});
+    // Update internal state only if being used as an uncontrolled component.
+    if (!this.props.value) {
+      this.setState({selected: value});
+    }
```
rgarcia: I found parsing this confusing at first read... at some point it might worth making two different components instead of one that can be in either controlled mode or uncontrolled mode

rgarcia:
lgtm! A little bit concerned about the barriers we might be creating to understand this component. I found it difficult to first get up to speed on controlled vs. uncontrolled, and then hold the subtle `this.props.value` vs. `this.state.selected` differentiator in my head while reading the code.

src/SegmentedControl/SegmentedControl.jsx
kofi-clever: thanks - fixed

src/SegmentedControl/SegmentedControl.jsx
kofi-clever: True - this conditional isn't really necessary, come to think of it. The internal need not depend on the value override. That override can just happen at render time.
Reverted it to try to simplify things a bit.

PR #39, opened 2016-08-10
Add segmented control button type=button so it's not a submit button

kofi-clever:
Forgot to set the button type when switching from `span` to `button` for the individual controls. That turned it into a submit button by default :/


nathanleiby:
:shipit:

PR #42, opened 2016-08-19
add loaders to package dependencies so that works elsewhere

schimmy:
This makes it more likely that the package just works when included elsewhere, as not all including repos have these loaders installed.


jacobsimon:
LGTM

PR #44, opened 2016-08-23
Use --save instead of --save-dev when installing components

agclever:
Using `--save` ensures that these get installed even if we do `npm install --production`


jacobsimon:
lgtm

PR #45, opened 2016-08-23
Add a Table component

kofi-clever:
Working off mocks at https://clever.atlassian.net/browse/DESIGN-56

**Supports:**
- Sorting
- Filtering
- Fixed and fluid column widths

**TODO:**
- Pagination controls
- Detached body scrolling

**Sorting Demo:**
table-sort

**Filtering Demo:**
table-filter


kofi-clever:
cc @agclever & @CarlNelson

kofi-clever:
Fixed the broken test - ready for review now.
LMK if you aren't able to get to it today - can re-assign if necessary.

src/Table/Cell.jsx
```diff
@@ -0,0 +1,36 @@
+import React, {PropTypes} from "react";
+
+import {joinClasses as cls} from "../utils";
```
osdiab: I don't see much of a benefit from aliasing, i'd just use it as `joinClasses` to make it more homogeneous

src/Table/Cell.jsx
```diff
@@ -0,0 +1,36 @@
...
+}
+
+Cell.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node,
+  noWrap: PropTypes.bool,
+  width: PropTypes.string,
+};
+
+Cell.cssClass = {
```
osdiab: Cute, i like :)

src/Table/Cell.jsx
```diff
@@ -0,0 +1,36 @@
...
+      {children}
+    </td>
+  );
+}
+
+Cell.propTypes = {
+  className: PropTypes.string,
+  children: PropTypes.node,
+  noWrap: PropTypes.bool,
+  width: PropTypes.string,
```
osdiab: If we want to be precise about the propTypes we can actually custom validate ones like this to be valid CSS dimensions.

src/Table/Column.jsx
```diff
@@ -0,0 +1,21 @@
+import {PropTypes} from "react";
+
+
+export default function Column() {
+  throw new Error("Configuration component - not meant to be rendered.");
```
osdiab: Whoa, this is awesome!

src/Table/HeaderCell.less
```diff
@@ -0,0 +1,43 @@
+@import "../colors.less";
+
+
+.Table--header--cell {
+  background-color: @neutral_off_white;
+  border-bottom: 3px solid @neutral_silver;
+  color: @neutral_black;
+  font-size: 16px;
```
CarlNelson: Should be 1rem

src/Table/Table.less
```diff
@@ -0,0 +1,24 @@
+@import "../colors.less";
+@import "../fonts.less";
+
+.Table {
+  border: 1px solid @neutral_silver;
+  border-collapse: collapse;
+  color: @neutral_black;
+  font-family: "Proxima Nova";
+  font-size: 14px;
```
CarlNelson: Nonstandard font-size. Bump back up to 1rem or down to .75rem

src/Table/HeaderCell.less
kofi-clever: ah cool, didn't know we were using `rem` units!
should the header be a little larger than the body size, then? 1.25rem?

src/Table/Table.less
kofi-clever: thx - will switch to 1rem

src/Table/HeaderCell.less
kofi-clever: actually, thought clever-common was setting the root to 14px, but it just leaves it as the default, so i'll leave this at 1rem and switch the body to 0.75rem and see what that looks like

osdiab:
This looks solid to me—LGTM!

src/Table/Table.jsx
```diff
@@ -0,0 +1,132 @@
...
+    const {cssClass} = Table;
+
+    return (
+      <table className={joinClasses(cssClass.TABLE, fixed && cssClass.FIXED, className)}>
+        <Header onSortChange={columnIndex => this._toggleSort(columnIndex)} sortState={sortState}>
+          {children}
+        </Header>
+        <tbody className={cssClass.BODY}>
+          {sortedData.map((rowData, rowIndex) => (filter(rowData) ? (
+            <tr className={cssClass.ROW} key={rowIDFn ? rowIDFn(rowData) : rowIndex}>
```
osdiab: if `rowIDFn` isn't present but this table is sortable, will the keys get jumbled every time sort order changes? seems potentially confusing

osdiab:
I have once concern about sorting actually, and some docs in the README would be great too.

.babelrc
```diff
@@ -4,6 +4,8 @@
     "react"
   ],
   "plugins": [
-    "transform-object-assign"
```
osdiab: Why is object assign being removed?

src/utils/index.js
```diff
@@ -52,3 +52,10 @@ export function unprefixKeys(obj, prefixToRemove, lowercaseFirst = true) {
     return prev;
   }, {});
 }
+
+/**
+ * Returns a className string containing all non-empty classes from the given list of candidates.
+ */
+export function joinClasses(...candidates) {
```
osdiab: Are there any edge cases with a custom implementation of this vs what React recommends here? https://github.com/JedWatson/classnames

src/Table/Table.less
kofi-clever: hm, @CarlNelson -

0.75rem body looks too small:
screen shot 2016-08-25 at 12 17 43 pm

and 1rem body looks a little too big:
screen shot 2016-08-25 at 12 29 59 pm

0.875rem (basically 14px) with 1rem header gets it looking sane, but 0.875 isn't currently available in design-system.
LMK if that's fine.

.babelrc
kofi-clever: it just got reshuffled when i added the new plugins (alphabetically arranged)

src/utils/index.js
kofi-clever: i'd actually much rather use `classnames` - initially introduced it in https://github.com/Clever/components/pull/36#discussion-diff-73739880 but reverted it due to concerns about bloat. `joinClasses` is a compromise, but down to switch to `classnames` if there are no objections

src/Table/Table.jsx
kofi-clever: that would just mean that React would re-render a row whose index has changed instead of re-using it - which, in the case of sorting, has no effect on performance, since most, if not all rows need to be re-rendered anyway due to the re-ordering.

it really only comes in handy for filtering, where React can cache a previously-rendered row when it's filtered out and pop it back in when the filter is removed, without re-instantiating it. even then, it's doesn't provide significant gains.

.babelrc
osdiab: oops doh

src/Table/Table.jsx
osdiab: Is that really true though? I know there's really weird stuff that happens when the keys don't actually uniquely identify content (like re-renders not occurring properly). I've had to deal with some devilish bugs because of keys that weren't correctly unique.

osdiab: If you're confident it's just unnecessary rerenders then go ahead, but fudging around with keys is one thing that puts me on edge

osdiab: i guess if we can guarantee that a table will never be used with anything that can potentially contain state it's probably safe

src/Table/Table.jsx
```diff
@@ -0,0 +1,132 @@
...
+  children: PropTypes.arrayOf(PropTypes.shape({
+    type: PropTypes.oneOf([Column]),
+  })),
+  className: PropTypes.string,
+  data: PropTypes.array.isRequired,
+  filter: PropTypes.func,
+  fixed: PropTypes.bool,
+  initialSortState: tablePropTypes.sortState,
+  onSortChange: PropTypes.func,
+  rowIDFn: PropTypes.func,
```
osdiab: I don't think we have a convention around this but I prefer camelcasing acronyms (i.e. `rowIdFn`) just to disambiguate with consecutive acronyms, and for consistency. minor nit

src/Table/Table.jsx
osdiab: Though the above assumption precludes use of things like `<input>` tags in a table, which seems like a pattern that isn't unlikely to occur, and would definitely be messed up by this without `rowIDFn` present; this is a pattern that is not enforced by `PropTypes`. what do you think?

src/utils/index.js
osdiab: I don't have an objection, it doesn't have any dependencies, and it is endorsed by React themselves so I'd like to see that.

src/Table/Table.jsx
osdiab: http://stackoverflow.com/a/27172000/1105281

kofi-clever: yeah, i feel the exact same way. i've done it the other way for most of my career, but i noticed we follow the golang convention of always capitalizing acronyms and it can get pretty annoying switching naming conventions between different languages. in any case, i'm fine with standardizing on either pattern, as long as we have a pattern

src/Table/Table.jsx
kofi-clever: yeah, things do get tricky with stateful components. ran into that issue with repeated form components once. we can make `rowIDFn` required to avoid any potential headaches. not really much overhead for clients to specify it

osdiab: yeah I think it's worth it because implicit in making a table like this is iterating over multiple elements. If we want to be super safe we can even keep track of the set of IDs calculated in this step and throw an error if there are duplicate IDs, that would be pretty cool

src/utils/index.js
kofi-clever: +1 - https://github.com/Clever/components/pull/45/commits/3ae96cc893aec7e0e7405d4eef0e8dfb03ebbe2c

src/Table/Table.jsx
kofi-clever: agreed - added requirement: https://github.com/Clever/components/pull/45/commits/baa397d642edbc629f6630d28fc6b4034f895435

osdiab:
K, with those things LGTM!

kofi-clever:
danke!

kofi-clever:
i'll add a README update as well - almost forgot

PR #46, opened 2016-08-24
Select component that works for a basic dropdown

agclever:
JIRA: https://clever.atlassian.net/browse/DIS2-179

Creates a new select component that works as a basic dropdown for now. Wraps react-select with styling based on mocks at https://projects.invisionapp.com/share/TG89776AU#/screens/180681819.

Wraps react-select and applies custom styling.

**Screenshots**

Focused on an option with none selected:
screenshot 2016-08-24 16 24 10

Focused with an option selected:
screenshot 2016-08-24 16 24 18

### DIFF
--- a/src/Select/Select.jsx
+++ b/src/Select/Select.jsx
@@ -1,0 +1,65 @@
+import React from "react";
+import ReactSelect from "react-select";
+
+import "react-select/dist/react-select.css";
+import "./Select.less";
+
+/*
+  Component to allow selecting options from a list. Right now this only supports a basic dropdown
+  with a fixed list of options. This component will be updated to allow searching for options that
+  may be fetched asynchronously.
+*/
+
+export function Select({id, name, label, onChange, options, placeholder = "", value}) {
+  const {cssClass} = Select;
+
+  let labelContainerClasses = cssClass.LABEL_CONTAINER;
+  if (placeholder && !value) {
+    labelContainerClasses += ` ${cssClass.LABEL_HIDDEN}`;
+  }
+
+  // The label container must be returned after the ReactSelect otherwise it does not get displayed
+  // in the browser.
+  return (
+    <div className={cssClass.CONTAINER}>
+      <div id={id}>
+        <ReactSelect
+          className={cssClass.REACT_SELECT}
+          clearable={false}
+          name={name}
+          onChange={onChange}
+          options={options}
+          placeholder={placeholder}
+          searchable={false}
+          value={value}
+        />
+      </div>
+      <div className={labelContainerClasses}>
+        <label className={cssClass.LABEL} htmlFor={id}>{label}</label>
+      </div>
+    </div>
+  );
+}
+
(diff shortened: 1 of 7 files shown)


agclever:
cc: @CarlNelson @joshatclever

CarlNelson:
Change display: none on label--hidden to opacity: 0 to allow for screen readers to read it. Use a transition to fade it in.

agclever:
Changed to use opacity in 72aef722c55f06aade155401164d626f. Did not use transition because the placeholder does not have transitions and there isn't a straightforward way for us to make sync those up

ipam73:
@ag could you document the new component on the readme?

agclever:
Updated README in ceb6b37

src/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+          placeholder={placeholder}
+          searchable={false}
+          value={value}
+        />
+      </div>
+    </div>
+  );
+}
+
+Select.cssClass = {
```
ipam73: nice way to simplify css classes!

src/Select/Select.jsx
```diff
@@ -0,0 +1,63 @@
...
+  CONTAINER: "Select--container",
+  LABEL: "Select--label",
+  LABEL_CONTAINER: "Select--labelContainer",
+  LABEL_HIDDEN: "Select--labelHidden",
+  REACT_SELECT: "Select--ReactSelect",
+};
+
+const selectValue = React.PropTypes.shape({
+  label: React.PropTypes.string,
```
ipam73: are these required for a select value?  or are they both optional?

src/Select/Select.less
```diff
@@ -0,0 +1,69 @@
...
+      }
+      .Select-value {
+        font-size: 15px;
+        line-height: 40px;
+        min-height: 40px;
+        padding-top: 10px;
+      }
+    }
+
+    // overrides styles from the exact same selector in react-select's CSS
```
ipam73: I'm not too familiar with best practices overriding other selections (if there's any) -- would probably defer on @CarlNelson on this

ipam73:
@agclever this looks good to me!  I'm deferring comments on style/design to @joshatclever and @CarlNelson

src/Select/Select.jsx
agclever: Good question. Made them both required in https://github.com/Clever/components/pull/46/commits/97885be5613578bf50b4460a703632e940c2909a

src/Select/Select.less
CarlNelson: Font size should be in them, Select value should be 1rem (16px)

agclever: Changed to use rem in 5765fe5521102e9ee23b90cdf671821082965b2f.

src/Select/Select.jsx
agclever: Borrowing from @kofi-clever's code

PR #48, opened 2016-08-26
Minor Table component fixes

kofi-clever:
- Reduce padding for inner cells to reduce cell spacing.
  **Before:**
screen shot 2016-08-26 at 1 51 06 am

  **After:**
screen shot 2016-08-26 at 1 51 19 am

- Handle data prop updates and update rows accordingly, while preserving
  sort state

Also, update the TableExample to demo updating the table with new data:
table-reload-demo

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ MOCHA := node_modules/mocha/bin/mocha
 MOCHA_OPTIONS := --compilers jsx:babel-register --recursive --require ignore-styles --require jsdom-global/register
 BABEL := node_modules/babel-cli/bin/babel.js
 
-.PHONY: test lint clean es5 build new $(TESTS)
+.PHONY: dev-server test lint clean es5 build new $(TESTS)
 
 clean:
 	@echo '✓ Clean out dist directory'
@@ -35,3 +35,6 @@ $(TESTS):
 	@echo "Running tests for $@"
 	@$(LINT) $@
 	@NODE_ENV=test $(MOCHA) $(MOCHA_OPTIONS) $@
+
+dev-server:
+	npm run-script dev-server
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -35,3 +35,6 @@ $(TESTS):
 	@echo "Running tests for $@"
 	@$(LINT) $@
 	@NODE_ENV=test $(MOCHA) $(MOCHA_OPTIONS) $@
+
+dev-server:
```
ipam73: could you update the readme to use `make dev-server`
https://github.com/Clever/components/blob/master/README.md#contributing

kofi-clever: thanks, good call - https://github.com/Clever/components/pull/48/commits/9e52fdda9d38810157003446c3c2165bab29ffdb

src/Table/Table.jsx
```diff
@@ -37,13 +46,13 @@ export class Table extends Component {
...
 
-    this._sort({
+    this._sort(this.props.data, {
       columnIndex,
       direction: sortDirection.ASCENDING,
     });
   }
 
-  _sort(sortState) {
+  _sort(data, sortState) {
```
ipam73: Is there a reason to pass `data` if it's not being used here?

ipam73: edit:  ignore this comment - see that it's being used below

ipam73:
@kofi-clever I'd refer to style changes to someone from design, but other than that LGTM

kofi-clever:
@CarlNelson - meant to ask actually: could you confirm the cell padding for the table?
i have it at 20px by 20px right now, based on https://clever.atlassian.net/browse/DESIGN-56
does that look ok?

PR #49, opened 2016-08-26
[Button] Remove href xor onclick constraint.

kofi-clever:
Valid Usage Scenario:
Link button with an href and an onclick listener that logs an event when
the user clicks the link.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.7.1",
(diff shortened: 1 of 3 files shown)


jacobsimon:
You'll need to pass onClick to the `<a>` element

kofi-clever:
> You'll need to pass onClick to the `<a>` element 

Missed that - thanks!

ipam73:
LGTM!

PR #50, opened 2016-08-31
Allow marking select as clearable

agclever:
Defaults to false. Doing a patch version bump since change is backwards compatible and we're pre 1.0

[https://gyazo.com/398043548710a6d99e53373a1a9b539a](https://gyazo.com/398043548710a6d99e53373a1a9b539a)


jacobsimon:
looks really good!!

jacobsimon:
don't forget to bump version

PR #51, opened 2016-09-02
[Table] Additional table component tweaks

kofi-clever:
- Hide sort controls when there are fewer than 2 rows visible.
- Filter and sort on render to simplify the code a bit. Shouldn't have an
  effect on performance, since rendering is already optimized based on
  prop/state comparisons.
- Also require column IDs to simplify the logic around sorting. Consistently
  identify sorted columns by ID instead of index.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.7.3",
+  "version": "0.8.0",
   "description": "A library of helpful React components",
   "repository": {
     "type": "git",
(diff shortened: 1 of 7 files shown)


CarlNelson:
LGTM

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.7.3",
+  "version": "0.8.0",
```
kshen0: This change is technically breaking - are we respecting semver pre-1.0?

kofi-clever: yeah - IIRC, a pre-1.0 minor bump is equivalent to a major version bump. npm won't automatically pick up 0.8.0 for `~0.7.x`, so clients would have to explicitly update.

currently only used in the apps dashboard, where we're already using column ids instead of indices

kshen0:
LGTM

PR #52, opened 2016-09-07
[Table] Add a pagination footer

kofi-clever:
- Add a pagination footer with default 10 rows per page.
  - Hidden if there's only 1 page of data available after filtering.
- Also show a 'NO DATA' row if the current filter has no matches.
- Update TableExample to include pagination over large sets of data.
table-pagination

- Also add a "linkPlain" Button type to support plain inline text links.
  - We've needed to manually strip the padding and centering in a few cases.
     This should make it more convenient to use in those scenarios.
- Update live examples with a 'linkPlain' Button.
screen shot 2016-09-07 at 1 48 21 am

- Minor bump 0.8.0 -> 0.9.0
  - Breaking change bump since tables that would usually display all data
     rows will now paginate by default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ This is a set of button components with various sizes and types.
 | Prop                | Type     | Description                                          | Default
 |---------------------|----------|------------------------------------------------------|---------
 | value               | String   | The text that appears on the button                  | None
-| type                | String   | One of `primary`, `secondary`, `destructive`, `link` | `secondary`
+| type                | String   | One of `primary`, `secondary`, `destructive`, `link`, `linkPlain` | `secondary`
 | size                | String   | One of "large", "regular", "small"                   | `regular`
 | onClick (optional)  | Function | Called when the user clicks on the button            | None
 | href (optional)     | String   | If provided, causes the button to behave as a link   | None
@@ -111,7 +111,7 @@ the string `"modal"` prepended to them (i.e. `modalWidth`, `modalTitle`).
 | onClick (optional)               | Function | Called when the user clicks the button               | None
 | onClose (optional)               | Function | Called when the user closes the modal                | None
 | value                            | String   | The text that appears on the button                  | None
-| type                             | String   | One of `primary`, `secondary`, `destructive`, `link` | `secondary`
+| type                             | String   | One of `primary`, `secondary`, `destructive`, `link`, `linkPlain` | `secondary`
 | size                             | String   | One of "large", "regular", "small"                   | `regular`
 | href (optional)                  | String   | If provided, causes the button to behave as a link   | None
 | target (optional)                | String   | For links, either "_self" or "_blank"                | "_blank"
@@ -151,7 +151,7 @@ Inherits all options from `ModalButton` and `Button`. `Button`'s properties have
 | modalTitle                       | String   | Header text for the modal                                   | None
 | modalWidth (optional)            | Number   | Width of the modal                                          | 400px
 | confirmButtonValue               | String   | The text that appears on the Confirm button                 | None
-| confirmButtonType                | String   | One of `primary`, `secondary`, `destructive`, `link`        | `secondary`
+| confirmButtonType                | String   | One of `primary`, `secondary`, `destructive`, `link`, `linkPlain`        | `secondary`
 | confirmButtonSize                | String   | One of "large", "regular", "small"                          | `regular`
 | confirmButtonHref (optional)     | String   | If provided, causes the Confirm button to behave as a link  | None
 | confirmButtonTarget (optional)   | String   | For links, either "_self" or "_blank"                       | "_blank"
@@ -314,7 +314,7 @@ var onSelect = function(selected) {
 
 ### Table
 
-Table component supporting sorting and filtering.
+Table component supporting sorting, filtering and pagination.
 
 **Options**
 
@@ -324,9 +324,17 @@ Table component supporting sorting and filtering.
 | data (required) | Array | The array of data items, each corresponding to a single potential table row | None
 | filter (optional) | Function | Called with data for a single row. Should return `false` if the row should be filtered out, or `true` otherwise. | None
 | fixed (optional) | Boolean | Whether or not table column widths should be fixed (vs fluid). | False
+| initialPage (optional) | Number | The initial page to be displayed initially. | None
 | initialSortState (optional) | `{columnID: String, direction: Table.sortDirection}` | The initial sort state of the table. | None
+| onPageChange (optional) | Function | Callback function for the displayed page change event. | None
 | onSortChange (optional) | Function | Callback function for the sort state change event. | None
+| pageSize (optional) | Number | The number of data rows to display on each page. Set to the length of the data set to effectively disable pagination. | 10
 | rowIDFn (required) | Function | Called with data for a single row. Should return the unique ID for that row. | None
+
+**API**
+
+- `setCurrentPage(page: Number)` - Explicitly sets the displayed page on the Table. Useful for reacting to data or filter changes that warrant resetting the current page.
+  - **NOTE:** The current page is automatically reset to the first page on every sort state change.
 
 #### Table.Column
 The `Table` component requires child components of type `Table.Column`, which provide configuration for the table header and row cells.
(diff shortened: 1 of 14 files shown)


kofi-clever:
cc @CarlNelson

kofi-clever:
ping @osdiab - fyi, got a couple other PRs depending on this one.
OOO today and getting back monday though, so not really pressing until then

README.md
```diff
@@ -324,10 +324,18 @@ Table component supporting sorting and filtering.
 | data (required) | Array | The array of data items, each corresponding to a single potential table row | None
 | filter (optional) | Function | Called with data for a single row. Should return `false` if the row should be filtered out, or `true` otherwise. | None
 | fixed (optional) | Boolean | Whether or not table column widths should be fixed (vs fluid). | False
+| initialPage (optional) | Number | The initial page to be displayed initially. | None
 | initialSortState (optional) | `{columnID: String, direction: Table.sortDirection}` | The initial sort state of the table. | None
+| onPageChange (optional) | Function | Callback function for the displayed page change event. | None
 | onSortChange (optional) | Function | Callback function for the sort state change event. | None
+| pageSize (optional) | Number | The number of data rows to display on each page. Set to the length of the data set to effectively disable pagination. | 10
```
osdiab: Should disabling pagination be an option in itself?

src/Table/Footer.jsx
```diff
@@ -0,0 +1,149 @@
...
+
+/**
+ * Pagination footer for the Table component.
+ * Only rendered if there are more than 1 page of data available.
+ */
+export default class Footer extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
```
osdiab: Why is this necesary?

src/Table/Footer.jsx
```diff
@@ -0,0 +1,149 @@
...
+  _selectPage(page) {
+    if (page === this.props.currentPage) {
+      return;
+    }
+
+    this.props.onPageChange(page);
+  }
+
+  _renderEllipsis() {
+    return <span className={Footer.cssClass.ELLIPSIS}>{"•••"}</span>;
```
osdiab: There is an ellipsis character, idk if you want to use it: …

README.md
kofi-clever: yeah, +1 - made pagination optional with a `paginated` prop - false by default

src/Table/Footer.jsx
kofi-clever: yeah, just wanted something a little more prominent than the default. switched back to hellip and bumped up the font-size instead.

kofi-clever:
thanks @osdiab - cleaned up per feedback. PTAL

src/Table/Footer.jsx
```diff
@@ -0,0 +1,129 @@
...
+
+import {Button} from "../Button/Button";
+import Cell from "./Cell";
+
+require("./Footer.less");
+
+
+/**
+ * Pagination footer for the Table component.
+ * Only rendered if there are more than 1 page of data available.
```
osdiab: is

src/Table/Table.jsx
```diff
@@ -5,17 +5,37 @@ import React, {Component, PropTypes} from "react";
...
+   */
+  setCurrentPage(page) {
+    if (page === this.state.currentPage) {
+      return;
+    }
+
+    this.setState({currentPage: page}, () => this.props.onPageChange(page));
+  }
+
+  resetPage() {
```
osdiab: Is this being used anywhere?

src/Table/Table.jsx
```diff
@@ -37,11 +57,23 @@ export class Table extends Component {
     }
 
     this.setState({sortState: newSortState}, () => this.props.onSortChange(this.state.sortState));
+    // Reset to 1st page since table sort has changed.
+    this.setCurrentPage(1);
```
osdiab: Is the first page the `0`th page?

src/Table/Table.jsx
```diff
@@ -98,19 +148,26 @@ Table.propTypes = {
...
   filter: () => true,
+  onPageChange: () => {},
   onSortChange: () => {},
+  pageSize: DEFAULT_PAGE_SIZE,
 };
 
 Table.cssClass = {
   BODY: "Table--body",
   FIXED: "Table--fixed",
+  NO_DATA: "Table--no_data_cell",
```
osdiab: underscores are kinda uncommon for a CSS name, no big deal but `-` is more customary i think

osdiab:
minor nits. This looks good to me! i'd love to see how this would be modified to work with paginated endpoints :)

README.md
```diff
@@ -324,10 +324,19 @@ Table component supporting sorting and filtering.
 | data (required) | Array | The array of data items, each corresponding to a single potential table row | None
 | filter (optional) | Function | Called with data for a single row. Should return `false` if the row should be filtered out, or `true` otherwise. | None
 | fixed (optional) | Boolean | Whether or not table column widths should be fixed (vs fluid). | False
+| initialPage (optional) | Number | The initial page to be displayed initially. | None
 | initialSortState (optional) | `{columnID: String, direction: Table.sortDirection}` | The initial sort state of the table. | None
+| onPageChange (optional) | Function | Callback function for the displayed page change event. | None
```
osdiab: Does this happen asynchronously as the page change occurs, before, or after it? I could imagine implementing tables that fetch new data on page changes by listening on this function somehow (and backfilling fake data in `data`) if the actual page change blocked on this completing; but idk if that's a pattern we want to prevent altogether.

src/Table/Table.jsx
```diff
@@ -5,17 +5,37 @@ import React, {Component, PropTypes} from "react";
...
     super(props);
 
-    this.state = {};
+    this.state = {
+      currentPage: props.initialPage || 1,
+    };
+  }
+
+  /**
+   * @param {number} page - 1-based index of the page to select.
```
osdiab: idk if I like that this is 1-based, I think it would definitely confuse me.

README.md
```diff
@@ -324,10 +324,19 @@ Table component supporting sorting and filtering.
...
 | onSortChange (optional) | Function | Callback function for the sort state change event. | None
+| pageSize (optional) | Number | The number of data rows to display on each page. | 10
+| paginated (optional) | Boolean | Whether or not to enable pagination. See `pageSize` | false
 | rowIDFn (required) | Function | Called with data for a single row. Should return the unique ID for that row. | None
 
+**API**
+
+- `setCurrentPage(page: Number)` - Explicitly sets the displayed page on the Table. Useful for reacting to data or filter changes that warrant resetting the current page.
```
osdiab: document that this is 1-based, or make it 0-based (i prefer the latter)

README.md
kofi-clever: yeah, decided to keep the base table component simple and not bake in any async functionality.
we can achieve implement the load-on-page-scroll pattern by having placeholder cell contents (or loading animation that get replaced once the data is fetched. can add an example to the table demo in a follow-up PR

src/Table/Table.jsx
kofi-clever: thought it would be confusing if the page number is out of sync with what's displayed. in either case, we'll have to deal with a disconnect. no strong feelings on which way that disconnect happens

src/Table/Table.jsx
kofi-clever: no, replaced with `setCurrentPage`

src/Table/Table.jsx
kofi-clever: hm, seen many different formats over time, but usually dashes delineate hierarchy or modification (like table-row-cell) whereas snake_case or camelCase is used for logical units/compound names (like table-headerCell or table-header_cell)

do we have a pattern in place here? would be down to do a pass over this repo and fix all of them once we've settled on one

README.md
osdiab: ping on this

osdiab: And if you want to make it a bit less error prone you can add bounds checking so that if someone does try to use page index 0, it would warn the user that that's invalid; you can implement that as a custom proptype

osdiab:
aside from that last doc question, LGTM!

README.md
kofi-clever: good call - added prop validation

PR #53, opened 2016-09-08
Add grid and type sizes, create distributable stylesheet

jacobsimon:
**Summary**
This PR modifies the build process to also create a main stylesheet (src/less/index.less) that contains all the component styles, with the idea of being able to use them in a non-React context. It also adds the grid styles, and type sizes. 

It uses webpack to build a stylesheet that is saved as `dist/css/style_(version).css` and can then be uploaded by drone to S3 automatically.

I went ahead and uploaded the stylesheet that this produces, available [here](https://s3.amazonaws.com/assets.clever.com/design/style_0.9.2.css)


Makefile
```diff
@@ -19,9 +20,15 @@ es5:
...
 	@find ./dist -name "*.jsx" | xargs -n1 rm
 	@echo '✓ Remove JSX files'
-	@mkdir dist/css
 
-build: clean es5
+build: clean es5 styles
+
+styles:
+	@echo "Building stylesheet"
+	@[ -d dist ] || mkdir dist
```
osdiab: can you just use `mkdir -p`?

docs/css/style.css
```diff
@@ -0,0 +1,894 @@
+@font-face {
```
osdiab: does the docs css have to be checked in? don't particularly like compiled styles being part of any css-changing PR, and doesn't seem like the kind of thing we can't just add to the deployment script.

src/less/index.less
```diff
@@ -0,0 +1,17 @@
...
+@import "./type-size.less";
+@import "./grid.less";
+
+// Component Modules
+@import "../Button/Button";
+@import "../ConfirmationButton/ConfirmationButton";
+@import "../SegmentedControl/SegmentedControl";
+@import "../TextInput/TextInput";
+@import "../CopyableInput/CopyableInput";
+@import "../Modal/Modal";
```
osdiab: We have to keep this up to date now, right? maybe the new component script should also append to this file?

osdiab: alternatively we can use [less-plugin-glob](https://github.com/just-boris/less-plugin-glob) to just discover less files to import.

osdiab:
back to you!

src/less/index.less
CarlNelson: I'll defer to @jacobsimon on this since I don't know the component side as much

Makefile
CarlNelson: Makes sense. Fixed.

docs/css/style.css
CarlNelson: Until we get the CDN set up we'll need this for it to be used by the gh-pages docs.

osdiab: No, what I'm saying is the script that pushes docs to gh pages can also just compile the CSS on its own so that it doesn't need to pollute the master branch.

src/less/index.less
jacobsimon: Ok I updated the new component script. I think I'd rather be explicit about the imports so it's not too magical and we can control the order of the imports.

docs/index.html
```diff
@@ -2,6 +2,7 @@
 <html>
     <head>
         <script src="bundle.js" type="text/javascript" charset="utf-8"></script>
+        <link rel="stylesheet" type="text/css" href="https://assets.clever.com/design/style_0.9.2.css">
```
osdiab: So does that mean updating the docs now has one more step of also the version here and uploading to the CDN?

I'd make those instructions visible somewhere at a minimum; or to set it up so that some static URL always has the latest styles; though truth be told I actually don't see much of a benefit of referring to the CDN for the docs page as this isn't as mission critical a service.

jacobsimon: https://github.com/Clever/components/pull/53/commits/ac0a6ac645b4581a6b64641f74ca7fc1924d68c3

PR #54, opened 2016-09-13
Add git-push step to deploy script to keep master and gh-pages in sync.

kofi-clever:
Run `git push origin master` after gh-pages deployment.

This keeps master in sync with the gh-pages branch, which is required in order for future deployments to work.

Tested manually with last deployment (for v0.9.0)

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -2,6 +2,20 @@
 
 set -e
 
+currentBranch=$(git symbolic-ref --short HEAD)
+if [[ $currentBranch != "master" ]]
+then
+  echo -e "\n\033[0;31m✖ The deploy script must be run from the master branch.\033[0m\n";
+  exit 1;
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -9,3 +9,5 @@ git add docs/
 git commit
 
 git subtree push --prefix docs origin gh-pages
+
+git push origin master
```
osdiab: Is it guaranteed that the person is on master branch? I think this is making some assumptions that may not be true.

kofi-clever: yeah, true - added error-checking

osdiab:
I trust that this works, LGTM!

kofi-clever:
Thanks! Verified in master:
screen shot 2016-09-14 at 12 34 45 pm

PR #55, opened 2016-09-14
[Table] Skip data chunking logic when pagination is disabled.

kofi-clever:
Switched to non-paginated by default at the last minute in the previous
PR, but missed this part. Was previously only hiding the footer.

Disabling the chunking logic as well when pagination is disabled.


jacobsimon:
lgtm

PR #56, opened 2016-09-23
[Global stylesheet] Add color and sizing styles.

kofi-clever:
[ Paired with @CarlNelson ]

- Add `.shade` mixin for consistent color shading.
- Add common size definitions along with margin/padding rulesets based on those sizes.
- Also replace hardcoded Button size/color styles with the new mixins as proof-of-concept.

**Button size changes:**
button-compare

**Button shading changes:**
button-compare-hover

### DIFF
--- a/genSizing.js
+++ b/genSizing.js
@@ -1,0 +1,81 @@
+const _ = require("lodash");
+const fs = require("fs");
+
+
+const FILEPATH = "src/less/";
+const SIZING_FILENAME = "sizing.less";
+const SPACING_FILENAME = "spacing.less";
+const sizes = {
+  none: 0,
+  "3xs": 0.125,
+  "2xs": 0.25,
+  xs: 0.5,
+  s: 0.75,
+  m: 1,
+  l: 1.5,
+  xl: 2,
+  "2xl": 2.5,
+  "3xl": 3,
+  "4xl": 3.5,
+  "5xl": 4,
+};
+
+let sizingContents = [];
+let spacingContents = [];
+
+_.forEach(sizes, (size, sizeName) => {
+  // Trim leading 0s on fractional sizes.
+  const sizeStr = `${size}`.replace("0.", ".");
+
+  const sizeVar = `@size_${sizeName}`;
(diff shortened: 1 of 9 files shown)


src/Button/Button.less
```diff
@@ -1,34 +1,50 @@
...
   &.Button {
-    background: #ffffff;
-    border: 0;
-    border-top-right-radius: 3px;
-    border-top-left-radius: 3px;
+    .type_medium;
+    border-radius: 3px 3px 0 0;
+    border: 1px solid transparent;
+    border-bottom-width: 3px;
+    box-sizing: border-box;
```
kshen0: Does this change the look of `Button`s? If so, can you post a screenshot?

genSizing.js
```diff
@@ -0,0 +1,81 @@
...
+
+      `.${classPrefix}--right--${sizeName} { ${type}-right: ${sizeVar} };`,
+
+      `.${classPrefix}--bottom--${sizeName} { ${type}-bottom: ${sizeVar} };`,
+
+      `.${classPrefix}--left--${sizeName} { ${type}-left: ${sizeVar} };`,
+
+      `.${classPrefix}--x--${sizeName} { ${type}-left: ${sizeVar}; ${type}-right: ${sizeVar} };`,
+
+      `.${classPrefix}--y--${sizeName} { ${type}-top: ${sizeVar}; ${type}-bottom: ${sizeVar} };`,
```
CarlNelson: It's a good practice to create visual rhythm based on typography which is the core of most UIs. Padding and margin based on it provides that.

src/Button/Button.less
kofi-clever: not too much of a difference, but sizing ends up being a little more proportional to the text size.
shading on hover is also a little darker/more noticeable.

added screenshots to the PR description

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.9.2",
+  "version": "0.9.3",
```
kofi-clever: +1

PR #57, opened 2016-09-26
[Grid] Add grid component and IE10-compatible flex mixins

kofi-clever:
*[ Tested: Chrome, Firefox, Safari, IE10/11/Edge ]*

- Copy over flex mixins we've been using in the Apps and Schools
  dashboards and reconcile with existing flex definitions.
- Add a Grid component for React clients.
- Some additional cleanup done on grid styles.
  - Use new mixins and tweak grid settings to ensure IE10 support.
  - Change the Grid to a column-axis flex box with rows as items
    (instead of the previous row-axis box with wrapped rows) to better
    fit the idiom.
- Minor bump 0.10.0 -> 0.11.0

grid-single

grid-multi

grid-nested

grid-standalone-row

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,3 +354,37 @@ The `Table` component requires child components of type `Table.Column`, which pr
 **Usage Example**
 
 [Sample Code](https://github.com/Clever/components/tree/master/docs/TableExample.jsx) ([Live Demo](http://clever.github.io/components/#table))
+
+### Grid
+
+12-column grid component for consistent, simple and flexible layouts.
+
+**`<Grid />` Options**
+
+| Prop             | Type     | Description                           | Default
+|------------------|----------|---------------------------------------|---------
+| children (optional) | Array<Grid.Row> | The rows to render in the grid. | None
+| className (optional) | String | Additional classname to apply to the grid. | None
+| wrapperComponent (optional) | Any | The tagname or component class for the wrapper component to render for the grid | "div"
+
+**`<Grid.Row />` Options**
+
+| Prop             | Type     | Description                           | Default
+|------------------|----------|---------------------------------------|---------
+| children (optional) | Array<Grid.Col> | The columns to render in the row. | None
+| className (optional) | String | Additional classname to apply to the row. | None
+| grow (optional) | Boolean | Fluidly grows the row to fill any available vertical space. The parent Grid should have an explicit height set for this to work. | False
+| wrapperComponent (optional) | Any | The tagname or component class for the wrapper component to render for the row | "div"
+
+**`<Grid.Col />` Options**
+
+| Prop             | Type     | Description                           | Default
+|------------------|----------|---------------------------------------|---------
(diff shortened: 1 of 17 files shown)


src/Grid/Grid.jsx
```diff
@@ -0,0 +1,59 @@
...
+Grid.Size = Size;
+
+Grid.rowPropType = (props, propName) => {
+  const value = props[propName];
+
+  if (value === null || value === undefined) {
+    return null;
+  }
+
+  if (!value.type || value.type !== Row) {
```
kshen0: `!value.type` is redundant with the second check

README.md
```diff
@@ -354,3 +354,37 @@ The `Table` component requires child components of type `Table.Column`, which pr
 **Usage Example**
 
 [Sample Code](https://github.com/Clever/components/tree/master/docs/TableExample.jsx) ([Live Demo](http://clever.github.io/components/#table))
+
+### Grid
+
+12-column grid component for consistent, simple and flexible layouts.
```
kshen0: It took me a second to parse the hierarchy of `Grid`s. Maybe worth describing the general structure in a sentence or two eg. "A Grid component is made up of any number of `Row` components. Each `Row` component is made of any number of `Col` components."

src/Grid/Col.jsx
```diff
@@ -0,0 +1,54 @@
...
+    <Wrapper className={classnames(cssClass.COL, colSpanClasses, className)} {...additionalProps}>
+      {children}
+    </Wrapper>
+  );
+}
+
+Col.colSpanPropShape = {};
+_.forEach(Size, size => {
+  Col.colSpanPropShape[size] = PropTypes.number;
+});
```
kofi-clever: sets up the prop type for the `span` prop. don't want to have to manually update it if we add a new `Size`, so setting it up dynamically.
added a comment to hopefully make it clearer

src/Grid/Grid.jsx
```diff
@@ -0,0 +1,59 @@
...
+
+  return (
+    <Wrapper className={classnames(cssClass.GRID, className)} {...additionalProps}>
+      {children}
+    </Wrapper>
+  );
+}
+
+Grid.Row = Row;
+Grid.Col = Col;
```
kofi-clever: wanted `Grid` to be a consistent entry point, so clients can use `<Grid />`, `<Grid.Row />` and `<Grid.Col />` instead of `<Grid.Row.Col />`

src/Grid/Grid.jsx
kofi-clever: thx - fixed

README.md
kofi-clever: added more info to the README - https://github.com/Clever/components/pull/57/commits/7452425f57dcfcce6161be1743de10748623383b

CarlNelson:
👍 LGTM thanks for cleaning it up and making a component out of it

PR #58, opened 2016-09-28
[Select] Fix incorrect `value` propType and enable searching.

kofi-clever:
* Enabling a pass-through of the `searchable` prop for use with large option lists.
Some style tweaking was needed to position the search input properly.
This component needs to be reworked a bit - the styling looks very brittle.

* `react-select` supports specifying actual values in the `value` prop instead of option objects, which is a lot more convenient in some cases. Expanding the clever `Select` proptypes to support that use case as well.

* Bump 0.11.0 -> 0.12.0

searchable-select

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.11.0",
+  "version": "0.12.0",
   "description": "A library of helpful React components",
   "repository": {
     "type": "git",
--- a/README.md
+++ b/README.md
@@ -250,6 +250,7 @@ a fixed list of options.
 | onChange (optional) | Function | Called with new value when it changes | None
 | options (optional) | Array | Possible options. Must contain objects with label and value attributes | None
 | placeholder (optional) | String | Placeholder text | ""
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.11.0",
+  "version": "0.12.0",
```
johnhuangclever: Isn't this a major version bump since the interface changed?

kofi-clever: for pre-1.0 releases, a minor bump counts as a major rev, since 0.x releases are considered unstable.
patch bumps become the minor bump in that case

jacobsimon:
Code lgtm but noticing a few stylistic things, like the input text seems a little too close to the label and the cursor doesn't line up with the placeholder text

kofi-clever:
Yeah, noticed those as well when I dived into this. I'd like to do a refresh of this component and try to fix/simplify the styles a bit. We've had to hack around react-select's styling, which won't be reliable long-term

PR #61, opened 2016-09-30
Adding type utility styles

CarlNelson:
Created basic type utility mixins and classes for use in components.

- .text--bold, .text--italic, etc...
- added some missing font weights and styles (italic, heavy)


src/less/type-utilities.less
```diff
@@ -0,0 +1,81 @@
+
+@import (reference) "./fonts.less";
+
+/* Basscss Typography */
+
+.font-family-inherit { font-family: inherit }
+
+.font-size-inherit { font-size: inherit }
+
+.text--decoration-none { text-decoration: none }
```
kshen0: What is the philosophy around making these as explicit classes vs. just including these styles as part of larger classes?

eg.
```
.my-link {
  color: #000;
}
...
<a class="my-link text--decoration-none" href="#">Click here</a>
```

vs.

```
.my-link {
  color: #000;
  text-decoration: none;
}
...
<a class="my-link" href="#">Click here</a>
```

src/less/index.less
```diff
@@ -9,6 +9,7 @@
 @import "./sizing.less";
 @import "./spacing.less";
 @import "./type-size.less";
+@import	"./type-utilities.less";
```
CarlNelson: Shakes fist at tab vs space

src/less/type-utilities.less
CarlNelson: Considering they're written as classes, they can be used as mixins in less or as classes to be used inline.

If you're building a content heavy page, classes often make more sense than embedding in custom blocks of css for each kind of element or object element.

In your above example .my-link wouldn't need to exist as you would have .text--black and .text--decoration-none and you could then re-use text--black and text--decoration-none without having to re-write the css for every different element that may use those classes.

kofi-clever:
(forgot to re-assign)

src/less/type-utilities.less
```diff
@@ -0,0 +1,81 @@
...
+
+.text--semi-bold    {
+   font-family: 'Proxima Nova', 'Helvetica Neue', Helvetica, sans-serif;
+   font-weight: 400;
+   font-style: normal;
+}
```
CarlNelson: Fixed.

PR #63, opened 2016-10-19
Bump version to 0.14.0 for type utilities

jonahkagan:
Also added some notes to the README about importing the design system styles and publishing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,14 @@ Install the NPM package and save it to your project using
 npm install --save clever-components
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -10,6 +10,14 @@ Install the NPM package and save it to your project using
 npm install --save clever-components
 ```
 
+Note: This library uses ES6 features not supported in Internet Explorer such as [`String.startsWith`](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith). To support use of this library in IE, include `babel-polyfill` in the entrypoint to your application. See the [docs](https://babeljs.io/docs/usage/polyfill/).
```
kofi-clever: ugh. is that really true?
good to know. we should rip those out and use lodash then.

jonahkagan: I have no idea if this is true - this was already in the README, I just moved it when reorganizing some things

PR #64, opened 2016-10-19
Type utility tweaks

jonahkagan:
- Corrects semi-bold `font-weight` to 600, instead of 400 (which is equivalent to `normal`)
- Removes letter spacing for all caps. `0.2em` was too much - the letters flew apart. I couldn't find a value I liked, so I just pulled it out for now. We can always readd it later once we have time to work through some examples. Currently, the all caps text in the product doesn't have modified letter spacing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.14.0",
+  "version": "0.14.1",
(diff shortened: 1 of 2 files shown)


CarlNelson:
LGTM

PR #67, opened 2016-10-26
Add FlexBox and FlexItem wrappers for abstracting away flex classes.

kofi-clever:
After using the flex classes for a bit in a few places, realised it'll be useful to have a component wrapper for them, to complement what we have with the `Grid`.

Abstracts the details of class names away into strict props and cuts down on duplicated CSS for components that reference the flex classes while hopefully making layout clearer from reading the jsx code.

First usage coming up in https://github.com/Clever/components/pull/68

flexbox-simple

flexbox-nested

flexbox-wrap

flexbox-column

### DIFF
--- a/README.md
+++ b/README.md
@@ -441,7 +441,7 @@ However `FlexBox` can be used without `FlexItem` in many cases.
 | alignSelf (optional) | ItemAlign | Sets the cross-axis alignment of this flex item only. | ItemAlign.STRETCH
 | children (optional) | React Node | The child items to render in the flex box. | None
 | className (optional) | String | Additional classname to apply to the flex box. | None
-| component (optional) | Any | The tagname or component class for the wrapper component to render for the flex item | "div"
+| component (optional) | Any | Tagname of class for the wrapper component. `FlexItem` renders as a `<div>` by default. It can be made to render as a different component by specifying the component tagname or class. e.g. `<FlexItem component="li" />` or `<FlexItem component={Button} />` | "div"
 | grow (optional) | Boolean | Fluidly grows the item to fill any available space along the main axis. | False
 
 #### `<FlexBox />` Options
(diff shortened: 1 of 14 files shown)


carolinemodic:
this is so cool! 
corgi_excited

CarlNelson:
👍  yeah!

osdiab:
Does it make sense to implement `Grid` based on this component?

README.md
```diff
@@ -421,3 +421,46 @@ viewport. Any additional `Grid.Col`s will wrap onto the following line.
...
+#### `<FlexItem />` Options
+
+([Source code](https://github.com/Clever/components/tree/master/src/flex/FlexItem.jsx))
+
+| Prop             | Type     | Description                           | Default
+|------------------|----------|---------------------------------------|---------
+| alignSelf (optional) | ItemAlign | Sets the cross-axis alignment of this flex item only. | ItemAlign.STRETCH
+| children (optional) | React Node | The child items to render in the flex box. | None
+| className (optional) | String | Additional classname to apply to the flex box. | None
+| component (optional) | Any | The tagname or component class for the wrapper component to render for the flex item | "div"
```
kofi-clever: yeah, will do a pass over this to improve the description a bit.

WRT the prop name, i'm borrowing from the pattern being used in `ReactTransitionGroup`, for example (also used in `react-router`'s `Route` component) - https://facebook.github.io/react/docs/animation.html#rendering-a-different-component

i initially used `wrapperComponent` like i did with the Grid component, but thought it might be useful to be consistent with other patterns out there

kofi-clever:
yeah, technically, we could implement `Grid` using `FlexBox`/`FlexItem` (and even support different grid sizes other than 12).
it would cause the react version of the grid to drift from the LESS-generated classes though, so we'd have to be prepared to maintain pieces of each separately.

for now, i think it's fine. at the base level, both the Grid classes and FlexBox are using the flex classes anyway.

PR #68, opened 2016-10-26
[TabBar] Add new TabBar component. 

kofi-clever:
Basically a row-direction FlexBox with buttons/anchors as child items.

tabbar-reprise

### DIFF
--- a/docs/TabBarExample.less
+++ b/docs/TabBarExample.less
@@ -1,0 +1,46 @@
+@import (reference) "../src/less/index.less";
+
+.TabBarExample {
+  .margin--top--2xl;
+  font-size: 0.875rem;
+
+  * {
+    box-sizing: border-box;
+  }
+
+  code {
+    .type_medium;
+    color: @neutral_gray;
+    font-weight: normal;
+  }
+
+  h2 {
+    .type_large;
+  }
+}
+
+.TabBarExample--config {
+  .margin--x--m;
+}
+
+h4.TabBarExample--configLabel {
+  padding: 0;
+  width: 5rem;
+}
+
+.TabBarExample--ViewContainer {
+  .margin--top--m;
+  .margin--x--m;
+  .padding--x--l;
+  background-color: @neutral_off_white;
+  border: @borderWidthSmall solid @neutral_silver;
+  border-radius: @borderRadiusMedium;
+  height: 15rem;
+  min-width: 37.5rem;
+}
+
+.TabBarExample--ViewContainer--heading {
+  .margin--bottom--m;
+  .margin--top--xl;
+  .margin--x--s;
+}
--- a/docs/TabBarExample.less
+++ b/docs/TabBarExample.less
@@ -29,12 +29,12 @@ h4.TabBarExample--configLabel {
 }
 
 .TabBarExample--ViewContainer {
+  .border--s(@neutral_silver);
+  .borderRadius--m;
   .margin--top--m;
   .margin--x--m;
   .padding--x--l;
   background-color: @neutral_off_white;
-  border: @borderWidthSmall solid @neutral_silver;
-  border-radius: @borderRadiusMedium;
   height: 15rem;
   min-width: 37.5rem;
 }
--- a/docs/TabBarExample.less
+++ b/docs/TabBarExample.less
@@ -9,13 +9,13 @@
   }
 
   code {
-    .type_medium;
+    .text--medium;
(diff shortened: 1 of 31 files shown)


docs/TabBarExample.less
```diff
@@ -0,0 +1,46 @@
+@import (reference) "../src/less/index.less";
+
+.TabBarExample {
+  .margin--top--2xl;
+  font-size: 0.875rem;
```
CarlNelson: We don't have a text--size for this, but I see it used semi-often. I wonder if we should make one or just enforce size--medium or size--small and remove this in between.

src/less/border.less
```diff
@@ -0,0 +1,6 @@
+@borderWidthSmall: .0625rem; // 1px
+@borderWidthMedium: .125rem; // 2px
+@borderWidthLarge: .1875rem; // 3px
+
+@borderRadiusSmall: .125rem; // 2px
+@borderRadiusMedium: .1875rem; // 3px
```
CarlNelson: I wonder if we should be more explicit about these similar to how we do the spacing file. Thoughts?

kofi-clever: ah, you mean generate classes for `border--top--s`, `borderRadius--topLeft--m`, etc?
that could be useful. i'll take a look and add a commit

CarlNelson: 👍

docs/TabBarExample.less
kofi-clever: true, might be worth adding a size definition for it. we do use .875rem and 14px quite a bit, but not firmly on one side or the other.
was thinking about this recently; 16px sometimes feels a tad too large for web app UI text. feels less of an issue for documentation/preamble text or for copy-heavy pages, but slightly off inside UI elements/controls. 

WDYT? i'm aware of my bias toward what i've used on other projects, so i'd defer to you on this one :)
definitely seen a trend toward larger in the last few years, but again that seems to be more focused on content-centric sites vs web apps.

docs/index.html
```diff
@@ -2,7 +2,6 @@
 <html>
     <head>
         <script src="bundle.js" type="text/javascript" charset="utf-8"></script>
-        <link rel="stylesheet" type="text/css" href="https://s3.amazonaws.com/assets.clever.com/design/style.css">
```
kofi-clever: this gets in the way of development a lot, because it pulls in outdated styles that interfere with the new ones we're testing. no need to have it here since all the demo components are pulling in the styles they need already (we just end up with duplicate styles in the browser).

docs/TabBarExample.less
CarlNelson: I tend to fall on the make the text bigger camp for legibility purposes even though I tend to prefer very small text sizes for myself.

10 / 12 / 14 / 16 feels like too many close type sizes to me and doesn't fit a scale very well. Will think on it.

kofi-clever:
Ok, added auto-generated border definitions in https://github.com/Clever/components/pull/68/commits/0c0218bd04ff90b96430a614cd2d05f96adb6671 

Added classes for `border-radius`, but leaving `border` as mixins for now, since it wouldn't be too useful (it would require defaulting to a specific color that would have to be changed in css anyway)

kofi-clever:
rebased with changes in master - ready for review now

docs/TabBarExample.less
```diff
@@ -0,0 +1,46 @@
...
+  }
+
+  code {
+    .type_medium;
+    color: @neutral_gray;
+    font-weight: normal;
+  }
+
+  h2 {
+    .type_large;
```
kofi-clever: thanks - updated

genBorder.js
```diff
@@ -0,0 +1,70 @@
...
+const fs = require("fs");
+
+
+const FILEPATH = "src/less/";
+const FILENAME = "border.less";
+const borderWidths = {
+  s: 0.0625, // 1px
+  m: 0.125, // 2px
+  l: 0.1875, // 3px
+};
```
kofi-clever: yeah, was thinking that too, but tested increasing the base font-size to see the effects and the borders growing along with the text actually didn't seem out of place at all. ends up working like an actual zoom.

that being said, not sure if that's actually what we want. seems fine, but will let @CarlNelson weigh in

CarlNelson: If the world were only ever made of rem and em units I would be happier. The ability for rem/em to scale more clearly is one of the parts that I think pixels don't do as well.

PR #70, opened 2016-10-27
replacing all type_size mixins with type--size

CarlNelson:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.14.3",
+  "version": "0.14.4",
```
kofi-clever: since this is a breaking change, we should run `npm version minor` instead to bump it to 0.15.0
(unless we make this a staged change, in which case 0.14.4 is fine - comment below)

src/Button/Button.less
```diff
@@ -22,7 +22,7 @@
 
 a, button {
   &.Button {
-    .type_medium;
+    .text--medium;
```
kofi-clever: slightly worried about making this a hard change - rather than leaving the old ones in first and then updating all codebases that are using them to the new one before deleting completely.

if they're being referenced only in LESS, it'll be fine since there will be build errors once people upgrade and they can fix them.
but if they're used as class names in JSX or jade/html, it'll be harder to detect the breaking change

CarlNelson: From my investigation there are very few references so far to .type_size mixins, the only repo that I found that used them was schools-dashboard so far and only in LESS.

kofi-clever: nice. yeah, just did a search and the only place where they're used in jsx are the user-search-ui repo i added recently. just 2 instances - i can update those one this is ready

CarlNelson:
Updated to 0.15.0. Merging.

PR #71, opened 2016-10-27
Add zindex variables

CarlNelson:
Patch update to be variable based instead of set values.

### DIFF
--- a/src/less/z-index.less
+++ b/src/less/z-index.less
@@ -3,42 +3,53 @@
 /* z-index issues.
 /*****************************/
 
+@zIndex_0: 0;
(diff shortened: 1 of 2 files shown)


src/less/z-index.less
```diff
@@ -3,42 +3,53 @@
 /* z-index issues.
 /*****************************/
 
+@zIndex_0: 0;
```
kofi-clever: i know our naming is all over the place now, but since we're converging on camel case in all other languages, we might as well do the same with our less - `@zIndex0`, etc in this case

CarlNelson: 👍

PR #72, opened 2016-10-27
Prop warning fixes

kofi-clever:
**[Button] Support HTML content in Buttons.**
We've been using <Button> with HTML labels for a while now. Updating
prop validation to officially support it and clear up the prop warnings
we've been seeing.

html-buttons

**[Table] Fix Footer prop warning when rendering tables with empty data.**
Skip data chunking if the data array is empty to avoid ending up with
an invalid value in the page number calculation.

### DIFF
--- a/src/Table/Table.jsx
+++ b/src/Table/Table.jsx
@@ -98,7 +98,7 @@ export class Table extends Component {
     const disableSort = displayedData.length <= 1;
 
     let pages = [displayedData];
-    if (paginated) {
+    if (paginated && displayedData.length > 0) {
       pages = lodash.chunk(displayedData, pageSize);
     }
     const numPages = pages.length;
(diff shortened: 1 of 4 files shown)


osdiab:
LGTM!

src/Table/Table.jsx
```diff
@@ -98,7 +98,7 @@ export class Table extends Component {
     const disableSort = displayedData.length <= 1;
 
     let pages = [displayedData];
-    if (paginated) {
+    if (paginated && displayedData.length > 0) {
```
osdiab: why does this condition need to be added?

kofi-clever: the chunking logic only works if there's actually data in the array, otherwise the assumptions that follow it breakdown when it returns an empty array instead of an array of arrays.

lemme see if i can rewrite that logic in a clearer way to avoid confusion for anyone reading through

PR #74, opened 2016-10-31
Make Grid.Col a flex item, not a flex box

jonahkagan:
Grid.Col should have normal flow for the children within it, so it should have display:
block, not flex. When desired, the `flexbox` class can be added to turn
a Col into a flexbox.

Also fixes some small bugs within flex.less.

Tested by visually comparing the grid examples.

### DIFF
--- a/src/less/flex.less
+++ b/src/less/flex.less
@@ -105,11 +105,11 @@
  * @initial {1}
  */
 .flexShrink(@value: 1) {
- -moz-box-flex-grow: @value;
+ -moz-box-flex-shrink: @value;
  -ms-flex-negative: @value;
- -webkit-box-flex-grow: @value;
- -webkit-flex-grow: @value;
- flex-grow: @value;
+ -webkit-box-flex-shrink: @value;
+ -webkit-flex-shrink: @value;
+ flex-shrink: @value;
 }
 
 /**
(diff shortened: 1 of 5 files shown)


src/less/flex.less
```diff
@@ -105,11 +105,11 @@
...
 .flexShrink(@value: 1) {
- -moz-box-flex-grow: @value;
+ -moz-box-flex-shrink: @value;
  -ms-flex-negative: @value;
- -webkit-box-flex-grow: @value;
- -webkit-flex-grow: @value;
- flex-grow: @value;
+ -webkit-box-flex-shrink: @value;
+ -webkit-flex-shrink: @value;
+ flex-shrink: @value;
```
kofi-clever: oh thanks for catching these! 🙏

src/less/flex.less
```diff
@@ -60,7 +60,7 @@
  * @param {number} basis - See `.flex-basis`. Defaults to auto.
  * @initial {0 1 auto}
  */
-.flex(@grow, @shrink: 1, @basis: auto) {
+.flex(@grow: 0, @shrink: 1, @basis: auto) {
```
kofi-clever: not sure it's useful to call `.flex()` with empty default arguments, since that's the browser initial value to begin with. probably better for people to be explicit about grow/shrink properties

jonahkagan: Ah ok. I probably just didn't understand what this meant. I'll revert it back to what it was before.

PR #76, opened 2016-11-08
Add Icon component

jonahkagan:
This PR adds an Icon component that renders icons from the icon library. See the docs for more detail.

Note that the images are added in a separate commit to make reviewing easier.

[https://gyazo.com/c382434a92d8777c710e8509e6742755](https://gyazo.com/c382434a92d8777c710e8509e6742755)

TODO
- tests
- figure out where to get missing large icons for microphone, mobile-apps, and typewriter (cc @CarlNelson)


CarlNelson:
We should get better assets for these so they are retina compatible which these aren't great for right now.

kofi-clever:
> open question: is putting the images in the assets dir and copying it to dist the best way to include the icons?

should probably host these on the CDN if possible. i think @jacobsimon did some work on getting the stylesheet there and would know more.

- WRT retina support - if we can, we should get 2x (and maybe 3x) versions of these and use media queries to pick the right one.

- also, have we considered spriting these? or at least group them into sprites of related icons? worried about load times and images popping in at different times after page load, but probably not a huge deal.

jonahkagan:
@kofi-clever I'd like to merge this as it stands in order to unblock my work with the left nav, which depends on this. I'm definitely down to come back and revisit adding higher-res/SVG icons and potentially hosting them on the CDN.

CarlNelson:
I just got access to the icon library file and can export them all as SVG if needed today.

jonahkagan:
@CarlNelson sure, whenever you have SVGs I can update this component.

CarlNelson:
SVG icons here... https://www.dropbox.com/sh/dqux2le6vvccaoz/AADeKcCW_wJCaK010eGsj7LHa?dl=0

jonahkagan:
Thanks Carl, I'll come back and add them soon.

PR #77, opened 2016-11-08
DESIGN-86 - guided wizard component

osdiab:
JIRA: https://clever.atlassian.net/browse/DESIGN-86

## Contents

* One (1) guided wizard component
* One (1) progress bar component

## Extreme Video Experience: Wizard

wizard-new

## Hardcore Gif Presentation: ProgressBar 

progress-bar-new

## Request for comments

* The abstraction is relatively complex—once I've written up the docs, please scrutinize the interface I've exposed, with an eye toward flexibility for future use cases.
* If you think I should break down the `Wizard` component into more discrete components, lmk—I ended up making it fairly monolithic because there's a lot of interactions across components that depend on shared state, and I decided to allow each step to modify a shared state in case there are complex interactions between them.

## TODO: 
- [x] post screenshots
- [x] update README
- [x] add tests
- [x] replace dumpy-looking colored box indicators with actual icons
- [x] improve progress bar styling
- [x] switch styles to use less helpers
- [x] accessibility check with @mischelle-m


src/Wizard/Wizard.jsx
```diff
@@ -0,0 +1,190 @@
...
+    } = this.props;
+
+    const classes = classnames("Wizard", className);
+    const curStep = steps[this.state.currentStep];
+    const validSteps = steps.map(step => step.validate(this.state.data));
+    const percentComplete = _.compact(validSteps).length / steps.length;
+
+    const nextDisabled = this.state.currentStep === steps.length - 1 ?
+      _.compact(steps.map(step => step.validate(this.state.data))).length !== steps.length
+    : !curStep.validate(this.state.data);
```
jacobsimon: There's a lot going on here. Can we break down these lines into something more readable? Or at least simplify with `steps.filter(...).length`

jacobsimon:
This looks really good! I don't have a lot of critique on the code assuming everything works as expected -- some test coverage would be nice here though.

I feel like the design could use a bit of polish:
* The layout feels like it takes up a lot of space horizontally
* Seems strange that the progress bar moves as soon as you enter input. I'd expect the progress bar to update when you progress to the next step. Don't feel too strongly though.

I'd also suggest that we make it a priority to ensure that existing consumers are properly excluding the components that aren't used (i.e. tree shaking) or consider putting this in a separate package with clever-components as a peer dependency.

osdiab:
If we do split it into a different package, i think that clever components is a hard dependency for the wizard, given that it uses buttons and the progress bar component here—what's the argument for it being a peer?

osdiab:
@mischelle-m any comment with the layout thing? btw the "Need help?" part collapses beneath the form when the window is too narrow. There isn't a behavior for mobile though

mischelle-m:
I think the layout works fine, especially once we go full width on both dashboard and portal. I agree that it is weird that the progress bar moves as soon as you enter input. Maybe it should move once we complete the validation check on a page per page basis?

osdiab:
@jacobsimon added docs.

osdiab:
also made `ProgressBar` have configurable direction (not used here, but figured that could be useful)

osdiab:
@mischelle-m updated the video, see if you like how it looks now with the behavior

osdiab:
Let's get the PR looking good here first before I split it into a different package, or else the comments will be fragmented. Or do we want to use the webpack 2 tree shaking?

mischelle-m:
@osdiab I like the new page validation interaction. I think it looks great!! Next step for me is to get you visual design and then content for CLASSROOM-996.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
+@import "../less/fonts.less";
+@import (reference) "../less/index";
+
+.Wizard {
+  display: flex;
+}
+
+.Wizard--sidebar {
+  display: flex;
+  flex-direction: column;
```
CarlNelson: You can use the flex mixins here.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+  flex: 0 0 200px;
+  padding: 0 10px;
```
CarlNelson: Use spacing mixins for all layout related padding / margin.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+  background-color: @neutral_off_white;
+  .border--right--m(@neutral_silver);
+
+  h2 {
+    font-weight: 300;
```
CarlNelson: Use type utilities mixins.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+.Wizard--contentGroup {
+  min-width: 600px;
```
CarlNelson: Use rem units.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+}
+
+.Wizard--step {
+  flex: 1;
+  padding: 0 20px;
+}
+
+.Wizard--WizardStep {
+  h1 {
+    width: 100%;
```
CarlNelson: This should probably be on a container element instead of directly on the h1.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+.Wizard--controls {
+  flex: 0 1;
+}
+
+.Wizard--stepsDisplay--stepLink {
+  width: 100%;
+  padding: 0 !important;
```
CarlNelson: Why are we using important here? We should avoid it at all costs.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+}
+
+.Wizard--stepsDisplay--step {
+  // ensure each item in list is vertically centered
+  display: flex;
+  align-items: center;
```
CarlNelson: Use the flex mixins.

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,129 @@
...
+  display: block;
+  // don't allow wrapping of text
+  overflow: hidden;
+  white-space: nowrap;
+  text-overflow: ellipsis;
+}
+
+.Wizard--stepsDisplay--currentStep, .Wizard--stepsDisplay--otherStep {
+  display: inline-block;
+  width: 20px;
```
CarlNelson: Use the spacing variables.

osdiab:
carl's comments: 9183146

src/Wizard/Wizard.less
```diff
@@ -0,0 +1,143 @@
...
+  .padding--none;
+}
+
+.Wizard--stepsDisplay--step {
+  // ensure each item in list is vertically centered
+  .flexbox();
+  .alignItems(center);
+  width: 100%;
+  .shade(color, @neutral_gray, 1);
+  .padding--y--2xs;
```
CarlNelson: This is more of a visual nit but the icon + label are too tight together.

CarlNelson:
Overall, LGTM.

One nit, the "Back" button should prolly be a Button--Link instead of Button--Secondary.

PR #78, opened 2016-11-11
DIS2-374: Add LeftNav component

jonahkagan:
This PR adds a LeftNav component. Docs and unit tests will be added in a later PR, since this is needed urgently.

[https://gyazo.com/7c254edec8b68d376a71cf075412e5bc](https://gyazo.com/7c254edec8b68d376a71cf075412e5bc)


jonahkagan:
@johnhuangclever @kofi-clever would love to get your initial thoughts. I haven't written docs or tests yet, but the code is functioning.

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  position: relative;
+
+  height: @size_3xl + @borderWidthS;
+  .border--bottom--s(@neutral_silver);
+}
+
+.LeftNav--subnav .NavLink {
+  .padding--left--s;
+  .padding--right--l;
+  .NavLink--icon {
```
kofi-clever: nit: +\n between class rules and nested rulesets

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  .list-reset;
+  .tint(background, @primary_blue, 6);
+  color: @neutral_black;
+  min-width: 9rem;
+  box-shadow: @borderWidthM 0 @borderWidthM @neutral_silver;
+}
+
+.NavLink {
+  .padding--left--m;
+  .padding--right--xl;
```
kofi-clever: more efficient to just use the plain properties when defining mixed margins/padding:
```less
padding: 0 @size_xl 0 @size_m;
```

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  .tint(background, @primary_blue, 6);
+  .shade(color, @primary_blue, 3);
+}
+
+.NavGroup--open {
+  .padding--top--none;
+  height: @size_3xl - @borderWidthS;
+  .border--top--s(@neutral_silver);
+  .border--bottom--m(@neutral_silver);
+  /*.text--regular;*/
```
kofi-clever: rm if unused

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  min-width: 9rem;
+  box-shadow: @borderWidthM 0 @borderWidthM @neutral_silver;
+}
+
+.NavLink {
+  .padding--left--m;
+  .padding--right--xl;
+  cursor: pointer;
+  position: relative;
+
```
kofi-clever: rm \n?

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  height: @size_3xl + @borderWidthS;
```
kofi-clever: not clear why this is needed - will `box-sizing: border-box` give you what you want?

more importantly, would be nice if we didn't have hard heights on these and just let font size and vertical padding define that for us.

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+  .padding--right--s;
+}
+
+.LeftNav--collapsed {
+  min-width: 0;
+  .NavLink {
+    .padding--right--s;
+  }
+  .NavLink--label {
+    display: none;
```
kofi-clever: can we do this in code instead? render null for the label if in the collapsed state.
it will be easier to animate it later that way

src/LeftNav/NavGroup.jsx
```diff
@@ -0,0 +1,33 @@
...
+    />
+  );
+}
+
+NavGroup.propTypes = {
+  id: PropTypes.string.isRequired,
+  icon: PropTypes.node.isRequired,
+  label: PropTypes.string.isRequired,
+  open: PropTypes.bool,
+  _onClick: PropTypes.func, // private - used by LeftNav only
```
kofi-clever: is that necessary? no need to make it restrictive if it doesn't affect functionality (we're wrapping the onClick anyway)

src/LeftNav/NavGroup.less
```diff
@@ -0,0 +1 @@
+
```
kofi-clever: rm if unused

src/LeftNav/NavLink.jsx
```diff
@@ -0,0 +1,37 @@
...
+  const selected = props.selected ? cssClass.SELECTED : null;
+
+  return (
+    <FlexBox
+      component="li"
+      className={classnames(cssClass.CONTAINER, props.className, selected)}
+      onClick={props.onClick}
+      alignItems={ItemAlign.CENTER}
+    >
+      <FlexItem className={cssClass.ICON}>{props.icon}</FlexItem>
```
kofi-clever: fyi: no need for `FlexItem` unless you're actually using one of the flex properties provided

src/LeftNav/NavLink.jsx
```diff
@@ -0,0 +1,37 @@
...
+      <FlexItem className={cssClass.LABEL} data-text={props.label}>{props.label}</FlexItem>
```
kofi-clever: is `data-text` used?

src/LeftNav/NavLink.less
```diff
@@ -0,0 +1,3 @@
+@import (reference) "../less/index.less";
+
+
```
kofi-clever: rm empty file

src/less/fonts.less
```diff
@@ -1,21 +1,14 @@
...
-  src: url('../../assets/fonts/ProximaNova-Bold-webfont.woff');
-  font-weight: bold;
-  font-style: normal;
-}
-
-@font-face {
-  font-family: 'Proxima Nova';
   src: url('../../assets/fonts/ProximaNova-Light-webfont.woff');
-  font-weight: 400;
+  font-weight: 200;
```
kofi-clever: finally! :tada:

src/LeftNav/LeftNav.jsx
```diff
@@ -0,0 +1,81 @@
...
+      // Configure NavGroups to open/close themselves on click
+      if (child.type === NavGroup) {
+        const open = child.props.id === this.state.openNavGroup;
+        return React.cloneElement(child, {
+          open,
+          _onClick: () => this.setState({openNavGroup: open ? null : child.props.id}),
+        });
+      }
+
+      return null; // Should never get here thanks to PropType validation
```
kofi-clever: note: prop validation only logs warnings, doesn't actually throw. that's still fine, since you're essentially filtering out unsupported children, but maybe we should document that, since it's not expected from regular React behaviour

src/LeftNav/LeftNav.jsx
```diff
@@ -0,0 +1,81 @@
...
+    });
+
+    // Find the open NavGroup so that we can render its children NavLinks in the drawer
+    const openChild  = React.Children.toArray(children).find(child => child.props.open);
+
+    const collapsed = this.props.collapsed ? cssClass.COLLAPSED : null;
+
+    return (
+      <FlexBox component="nav" className={cssClass.CONTAINER}>
+        <FlexItem component="ul" className={classnames(cssClass.TOPNAV, collapsed)}>{children}</FlexItem>
```
kofi-clever: here as well - this can just be a `<ul>` for simplicity. it's not using any FlexItem properties at the moment

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
+@import (reference) "../less/index.less";
+
+.LeftNav {
+  .shade(color, @primary_blue, 2);
+  background: @neutral_off_white;
+  user-select: none;
+  .text--light;
+  // Maintain computed width based on content
+  // (i.e. don't let the NavLink labels wrap)
+  .flexShrink(0);
```
jonahkagan: yep, thanks

src/LeftNav/LeftNav.less
jonahkagan: but then you have to remember the order of the four values. i find that a lot harder to read, so i want to stick with this

src/LeftNav/LeftNav.less
jonahkagan: this seems simpler to me for now. unless you have a strong reason to change it now, i'd rather just leave it since it works. but i'll keep that in mind if i come back and add animation for sure

src/LeftNav/NavGroup.jsx
jonahkagan: i just wanted to make it clear that you shouldn't add an onclick handler to a NavGroup as a user of the component since it won't get called

src/LeftNav/NavLink.jsx
jonahkagan: ahhhh good to know. thanks. that makes things cleaner

jonahkagan: what are your thoughts on using `<FlexBox>` as opposed to just adding `.flexbox` to the css?

src/LeftNav/NavLink.jsx
jonahkagan: nope, just remnants of a strategy that i abandoned

src/LeftNav/LeftNav.jsx
jonahkagan: will do

src/LeftNav/LeftNav.less
```diff
@@ -0,0 +1,80 @@
...
+}
+
+.LeftNav--topnav {
+  .list-reset;
+  min-width: 9rem;
+  box-shadow: @borderWidthM 0 @borderWidthM @neutral_silver;
```
jonahkagan: I'm going to come back and do visual design tweaks with Carl later, so I'll make sure to talk this over with him and see what he thinks

src/LeftNav/LeftNav.less
jonahkagan: this was the result of a lot of different strategies to try to thicken the border of a selected link without changing the positions of the links. i finally just now figured out a way to do it using box-shadow, so i was able to remove the hard heights

src/LeftNav/NavLink.jsx
jonahkagan: i tried it - i think it's cleaner because all the flex stuff does seem more like a style concern in this case. 8fa1121

jonahkagan:
@kofi-clever addressed your feedback in e97b7520a6fc075aba7cc5daba623ffef7650548

dist/Button/Button.js
```diff
@@ -0,0 +1,64 @@
+"use strict";
```
kofi-clever: is this intentional (checking in `/dist`)?

src/LeftNav/right-arrow.svg
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
kofi-clever: should make this a `RightArrow` component with svg content (e.g. https://github.com/Clever/apps-dashboard/blob/master/pages/ui/components/TopBar/Logo.jsx)

would mean that clients don't have to copy `right-arrow.svg` to the right place on the server in order to use the LeftNav

jonahkagan: clients don't have to copy it now. webpack builds it into the component. do you still think it needs to be made into a component?

kofi-clever: > webpack builds it into the component.

100% positive on that? we build to dist with plain babel, not webpack, so i think clients would have to manually include it in their webpack setup

PR #80, opened 2016-11-15
don't use es6 find in leftnav

osdiab:
Fixing ES6 usage in `LeftNav` causing it to fail on IE9, etc.

cc @jonahkagan fyi! It's kinda confusing—ES6 syntax that babel takes care of is fine, but functions that aren't polyfilled aren't (like Array methods and Object methods) because we don't polyfill in the context of the library.


jacobsimon:
lgtm!

PR #83, opened 2016-11-17
DESIGN-105: convert icons to svgs

CarlNelson:
Convert old png images to svg

OLD
https://i.gyazo.com/6191805c527d9976d572f045be092fe5.gif

NEW
https://i.gyazo.com/e454266095db225bdc301c23a4507a25.gif


src/Icon/images/blocks.svg
```diff
@@ -0,0 +1,19 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="46px" height="46px" viewBox="0 0 46 46" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 41 (35326) - http://www.bohemiancoding.com/sketch -->
```
kofi-clever: we should consider running our SVGs through an optimizer before merge to clean up some of the excess size.
things like these comments and the `title`, `desc` and empty elements can be stripped out without affecting the output.

osdiab:
Agree with kofi's comment, otherwise LGTM

src/Icon/images/blocks.svg
kofi-clever: looking into https://github.com/svg/svgo - i'll toy around with the various options and patch a commit to this PR when ready (will probably tackle after lunch)

thinking we should copy stuff over as `<filename>.min.svg` so we leave the originals intact but use the minified versions in production code.
would mean changing the references in the component to add the `.min` infix

PR #84, opened 2016-11-18
don't push wizard to bottom

osdiab:
Before, the controls defined on the wizard would get pushed all the way to the bottom of the wizard, so if you have a step with a lot of content it would be rendered with a lot of spacing in between the listed steps and the controls.

This just sets a padding between the steps list and the controls for the wizard.

A future development may be to have the wizard stick to a fixed position on screen regardless of how far you scroll down; this will be left for a future PR.

screen shot 2016-11-17 at 17 08 10

## TODO

- [ ] version bump

### DIFF
--- a/src/Wizard/Wizard.less
+++ b/src/Wizard/Wizard.less
@@ -6,8 +6,6 @@
 }
 
 .Wizard--sidebar {
-  .flexbox();
-  .flexDirection(column);
   .flex(0, 0, 200px);
   .padding--m;
   background-color: @neutral_off_white;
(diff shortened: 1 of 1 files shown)


src/Wizard/Wizard.less
```diff
@@ -71,7 +68,7 @@
 }
 
 .Wizard--controls {
-  .flex(0, 1);
+  .padding--top--l;
```
kofi-clever: use margin instead of padding for element separation

osdiab: 4f8364d

PR #87, opened 2016-11-18
Creating separate files for utilities and components

CarlNelson:
* 0.17.3
* you can now reference less/utilities for access to all non-component less utilities

### DIFF
--- a/src/less/index.less
+++ b/src/less/index.less
@@ -1,31 +1,5 @@
 // Clever Styles
-// Version 0.6.2
+// Version 0.17.3
 
-// Common constant, class and mixin definitions
-@import "./border.less";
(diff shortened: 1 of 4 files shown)


src/less/index.less
```diff
@@ -1,31 +1,5 @@
 // Clever Styles
-// Version 0.6.2
+// Version 0.17.3
```
osdiab: Not necessarily relevant to this PR, though is there a way we can tie the package version to this, or should we not put this here? since I'm pretty sure we won't remember to bump this version frequently

CarlNelson: Agreed. I imagine that's why it was stuck at 6.2 ;)

PR #88, opened 2016-11-19
Add back prefixed flex mixins to fix IE10 clients.

kofi-clever:
PostCSS isn't running as a build step, so clients don't get the benefit of the processing that was added recently. Since we removed the prefixes, clients are currently broken.

This partially reverts https://github.com/Clever/components/pull/82 and adds in additional mixin prefixes for special-case IE10 rules (mostly `flex-xxx` -> `xxx`).

### DIFF
--- a/src/less/flex.less
+++ b/src/less/flex.less
@@ -9,6 +9,9 @@
  * @target Flex container.
  */
 .flexbox() {
+  display: -moz-box;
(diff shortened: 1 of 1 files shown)


kofi-clever:
Merging now to fix clients, but we can re-visit later and remove in favour of PostCSS

PR #89, opened 2016-11-21
DIS2-476: Left nav visual design tweaks

CarlNelson:
Old Left Nav
old-left-nav

New Left Nav
new-left-nav


kofi-clever:
could you include a gif or screenshots for reference?

CarlNelson:
oops... done.

src/LeftNav/LeftNav.less
```diff
@@ -1,26 +1,32 @@
...
   .flexbox;
   .shade(color, @primary_blue, 2);
   background: @neutral_off_white;
   user-select: none;
-  .text--light;
+}
+
+.LeftNav--topnav,
+.LeftNav--subnav,
+.LeftNav--collapsed {
```
kofi-clever: `LeftNav--collapsed` might not be needed here since `LeftNav--topnav` is always assigned to the topnav anyway

src/LeftNav/LeftNav.less
```diff
@@ -99,3 +103,10 @@
     display: none;
   }
 }
+
+
+.expand {
```
kofi-clever: nit: could this be `.animateWidth`/`.transitionWidth` or something along those lines. on first read above, it seemed like it was applying expanded width styles

src/LeftNav/LeftNav.less
```diff
@@ -68,7 +74,7 @@
   // In order to have a background image (right arrow) and a background color
   // simultaneously, they both must be written in the same CSS rule and the
   // image must come second.
-  .tint(background, @primary_blue, 6);
+  .shade(background, @neutral_off_white, .15);
```
kofi-clever: q: (for future updates) do we need smaller shade steps maybe? also, will it be possible to prescribe a scale? then we could maybe have `.shade--1`, `shade--2`, etc.

src/less/animations.less
```diff
@@ -0,0 +1,66 @@
...
+
+.delayLong {
+  transition-delay: @timingSlowly;
+}
+
+.transition-timing-function(@function) {
+  transition-timing-function: @function;
+}
+
+.changeWidth(@width, @timing) {
```
kofi-clever: looks like this is unused now - remove?

src/less/animations.less
```diff
@@ -0,0 +1,66 @@
...
+}
+
+.transition-timing-function(@function) {
+  transition-timing-function: @function;
+}
+
+.changeWidth(@width, @timing) { 
+  animation-name: changeWidth;
+  transition: @width @timingPromptly;
+}
```
kofi-clever: nit: missing newline at the end

src/LeftNav/LeftNav.less
CarlNelson: Yeah, this was probably the first case where the existing steps didn't work at all. I've had a scale in my head, would be good to formalize eventually.

PR #91, opened 2016-11-21
add autoprefixer to styles webpack config

osdiab:


kofi-clever:
note: this would only work for clients that use the global stylesheet (rather than importing the various components/styles individually).
as long as we're leaving the manually-prefixed styles in place for now, we'll be fine wrt avoiding regressions. just making sure you're aware that this might have more limited reach than you intended

osdiab:
Oh, that's unfortunate. Hmm. I'll think a bit more on how this would generalize to included files then.

osdiab:
Sounds like there's a few possibilities that come to mind after talking with @kofi-clever :

1. Add autoprefixer to all clients (sucks)
1. Create a clever-components webpack plugin that does all the build steps and use that in all the consumers (not great, but probably better than doing it ad-hoc by each consumer)
1. Make styles accessible strictly through classes, and don't expose the LESS directly; give access to constants without the platform dependence on LESS without the necessity for build steps (this is what older-school systems like bootstrap do)

There may be better ways, this is a topic deserving of more research not from the "let's use autoprefixer" perspective but from the "what happens if we add any other build tools" perspective, as the current system precludes more sophisticated build setups (or consumers that use anything aside from LESS). Let's talk it over at frontend guild @CarlNelson @jacobsimon

osdiab:
Actually, can we use an npm `postinstall` step for this?

CarlNelson:
Decided to not move forward with this. Closing.

PR #92, opened 2016-11-22
[Select] Add support for custom option renderers

kofi-clever:
Just passes through the 'optionRenderer' prop to react-select.


osdiab:
Lgtm!

PR #93, opened 2016-11-22
DIS2-374: LeftNav tests and docs

jonahkagan:
This PR adds unit tests and documentation for the LeftNav component. It also makes two small tweaks to the LeftNav code - using React.Children.toArray in a place where we should be using it, and renaming a private prop. Each change is in a separate commit.

### DIFF
--- a/test/LeftNav_test.jsx
+++ b/test/LeftNav_test.jsx
@@ -1,0 +1,197 @@
+import assert from "assert";
+import React from "react";
+import sinon from "sinon";
+import {shallow} from "enzyme";
+import {LeftNav} from "../src";
+
+describe("LeftNav", function LeftNavTest() {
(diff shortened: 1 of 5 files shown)


test/LeftNav_test.jsx
```diff
@@ -0,0 +1,197 @@
+import assert from "assert";
+import React from "react";
+import sinon from "sinon";
+import {shallow} from "enzyme";
+import {LeftNav} from "../src";
+
+describe("LeftNav", function LeftNavTest() {
```
jonahkagan: Apparently if you want to use `this` to share context between `before` and `it` blocks, you have to have a named function somewhere in your test suite hierarchy. Otherwise you get an error from mocha that `this` is undefined.

kshen0:
LGTM

PR #95, opened 2016-11-28
LeftNav: use anchor tag for links for accessibility

jonahkagan:
This makes the LeftNav readable for screen readers and tabbable.

### DIFF
--- a/src/LeftNav/NavLink.jsx
+++ b/src/LeftNav/NavLink.jsx
@@ -7,13 +7,17 @@ export function NavLink(props) {
   const selected = props.selected ? cssClass.SELECTED : null;
 
   return (
-    <li
+    <a
       className={classnames(cssClass.CONTAINER, props.className, selected)}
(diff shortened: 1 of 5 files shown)


CarlNelson:
Since you're not actually passing an href to these <a> elements, they don't show up as tab accessible.

src/LeftNav/NavLink.jsx
```diff
@@ -7,13 +7,13 @@ export function NavLink(props) {
   const selected = props.selected ? cssClass.SELECTED : null;
 
   return (
-    <li
+    <a
```
CarlNelson: Since these don't include an href they're not accessible.

jonahkagan: ah ok i didn't actually understand how to test accessibility. fixed in 51b30fe

jonahkagan:
@CarlNelson and I paired, so this is good to go

PR #96, opened 2016-11-28
Makefile: use /bin/bash to ensure lint works on Drone

jonahkagan:
Previously, the lint script was erroring because Drone was using sh
instead of bash, but the error was causing the linter to pass instead of
fail. This change forces Drone to use bash, ensuring that lint failures
will make Drone fail.


cozmo:
Seems reasonable - Did we make sure this fails without the increase of the lint count or such?

jonahkagan:
Yeah, I only actually noticed this issue in the first place because the lint counts didn't match. Somebody must have not noticed that their changes increased the number of errors because they never ran the linter locally - only on Drone.

PR #98, opened 2016-11-28
Add "View Demo" links to each component heading in the README.

kofi-clever:
Updating demo links in the README.

Since we have deep demo links to all components now, adding demo links for all the sections in a consistent place for discoverability.

We can eventually move the descriptions into the docs themselves.


osdiab:
LGTM

PR #99, opened 2016-11-29
fix CopyableInput overflow issue

osdiab:
It would display text underneath the buttons. gross!

### Before...

copyable-input-before

### After!

copyable-input-after

### DIFF
--- a/docs/components/CopyableInputView.jsx
+++ b/docs/components/CopyableInputView.jsx
@@ -13,6 +13,7 @@ export default class CopyableInputView extends Component {
 
     this.state = {
       disabled: false,
+      readOnly: false,
       hasError: false,
       inputValue: "🙈 🙉 🙊",
       obscured: true,
@@ -39,6 +40,7 @@ export default class CopyableInputView extends Component {
               placeholder="CopyableInput Placeholder"
               onChange={e => this.setState({inputValue: e.target.value})}
               value={this.state.inputValue}
+              readOnly={this.state.readOnly}
             />
           `}
         >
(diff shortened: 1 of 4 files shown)


kofi-clever:
Nice! That's been bugging me :)
If you're up for some DTEC, TextInput needs the same treatment for when the "Show" button is enabled. Might even be able to fix it there for both

docs/components/CopyableInputView.jsx
```diff
@@ -69,6 +72,15 @@ export default class CopyableInputView extends Component {
           <label className={cssClass.CONFIG}>
             <input
               type="checkbox"
+              checked={this.state.readOnly}
+              onChange={({target}) => this.setState({readOnly: target.checked})}
+            />
+            {" "}
+            Read Only
```
kofi-clever: Could you add this to TextInputView as well? I forgot about readOnly

osdiab:
@kofi-clever just realized a problem: if there is an error or a required flag on the `CopyableInput`, those aren't going to show in the top right corner anymore, but to the left of the buttons :P If we wish to keep the semantics of CopyableInputs being validatable, I think a more flexible solution needs to involve providing custom buttons to the TextInput object.

osdiab:
I do want to get this resolved quickly, as it will show up every time in the Chromebook setup wizard I'm trying to push out quickly.

kofi-clever:
yeah, agreed. i think that's what i also landed on at a previous attempt at refactoring.
definitely no need to block on those issues, especially if this is pressing. we can revisit later. opened https://clever.atlassian.net/browse/FEE-19 earlier to try to address part of the issue

osdiab:
Cool, I think my plan is to merge it in in this form, not worry about the general TextInput case at the moment or the error/required things getting pushed over, and then fix those later. I think the most sensical scenario is probably to just allow string->callback mappings on TextInput and let this be a special case.

jacobsimon:
lgtm!

PR #101, opened 2016-11-30
Add onKeyPress as an optional param

peternga:
Allow passing in onKeyPress handler for TextInput component


src/TextInput/TextInput.jsx
```diff
@@ -97,6 +98,7 @@ TextInput.propTypes = {
   label: React.PropTypes.string,
   name: React.PropTypes.string.isRequired,
   onChange: React.PropTypes.func,
+  onKeyPress: React.PropTypes.func,
```
kofi-clever: Could you also add this to the TextInput docs in the README?

PR #104, opened 2016-12-01
don't make help push other stuff down

osdiab:
Had to use Javascript to detect whether or not the help text collapsed or not. This is now brittle to the width of the page.

Perhaps I should just do the movement in JS entirely... though that might be confusing to screen readers since the flow of the DOM would change?

### Before:

screen shot 2016-11-30 at 18 36 55

### After:

screen shot 2016-11-30 at 18 36 42

### Video:
help-text-overflow

### DIFF
--- a/docs/WizardExample.jsx
+++ b/docs/WizardExample.jsx
@@ -231,13 +231,19 @@ export default class WizardExample extends React.Component {
           handler: (wizardState, {resetWizard}) => resetWizard(),
           buttonValue: "Clear and start over",
         }]}
-        help={this.state.showHelp && <p>
-          Need any help? Check out our&nbsp;
-          <Button
-            onClick={() => alert("LOL, no help for you!")} type="link"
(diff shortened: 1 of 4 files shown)


osdiab:
ACK, this is not a good idea. Abort!!

docs/WizardExample.jsx
```diff
@@ -231,13 +231,19 @@ export default class WizardExample extends React.Component {
...
-          />
-        </p>}
+        help={this.state.showHelp && (
+          <div>
+            <h4 style={{marginTop: 0}}>Need any help?</h4>
+            <p>Don't fret! We got you!</p>
+            <p>
+              Check out our&nbsp;
+              <Button
+                onClick={() => alert("LOL, no help for you!")} type="link"
```
jacobsimon: 😈

jacobsimon:
I wish there were a cleaner way to do this, but I don't think there is, so LGTM!

PR #105, opened 2016-12-08
Left nav support invisible children

mb-dev:
With the addition of user roles, we will want to show/hide elements of the left nav. This PR adds a `visible` prop to Group and Link so the caller can toggle visibility and the component.

The behavior is:
1) If a Link visible=false, it shouldn't be visible
2) If none of the Group's Link are visible, the Group is not visible
3) If the Group visible=false, the group is not visible regardless of children.


jacobsimon:
I feel like it would be easier & more conventional to alter the children that are passed in rather than add another prop to the component, is there an advantage to doing it this way?

mb-dev:
okay, implemented it in schools-dashboard. closing

PR #106, opened 2016-12-15
Make sure button-like anchor controls have explicit block display styles

kofi-clever:
Related: https://clever.atlassian.net/browse/FEE-18

Some browsers (correctly) ignore spacing/sizing styles for inline elements.
Set anchor buttons to `display: inline-block` to keep display consistent with the pure button counterparts.

[Patch bump to v0.17.20]

**Before:**
screen shot 2016-12-15 at 1 53 10 pm

**After:**
screen shot 2016-12-15 at 1 53 24 pm

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.17.19",
+  "version": "0.17.20",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


kofi-clever:
> Any worry that this will impact existing users of the Button component who rely on `display: inline`?

was thinking through that - should be fine in all cases AFAICT. worst that would happen is a multi-word plain link button no longer wraps text within the button, but i didn't find any cases of that

PR #108, opened 2016-12-21
Stylelint updates

CarlNelson:
Added new rules for linting LESS

* No descending specificity
* Nested rules requiring an empty line before
* Max line length = 100


osdiab:
I like this, LGTM!

PR #110, opened 2017-01-12
Tweak Select component spacing to match TextInput

arsalansufi:
The `Select` component's internal spacing doesn't perfectly match `TextInput`'s. This results in a slightly off alignment when you use both in a form.

I've tried to match the spacing as best as I can. As per @CarlNelson's suggestion, I've also replaced all hard-coded sizes with sizing variables and mixins.

**Screenshots illustrating changes**

*Left-alignment before*

before

*Left-alignment after*

after

**Screenshots to show that my changes haven't broken the component**

*Standard*

general

*With multi option*

general-multi

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,7 +48,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 201
+LINT_MAX_LESS_PROBLEMS := 189
 lint:
 	@echo "Linting files..."
 	@$(LINT) $(JS_FILES) $(JSX_FILES)
--- a/src/Select/Select.less
+++ b/src/Select/Select.less
@@ -1,14 +1,16 @@
-@import "../less/colors.less";
+@import (reference) "../less/colors";
+@import (reference) "../less/spacing";
+@import (reference) "../less/type-size";
 
 .Select--container {
   position: relative;
 
   .Select--labelContainer {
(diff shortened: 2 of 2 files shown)


src/Select/Select.less
```diff
@@ -75,9 +79,10 @@
 
       .Select-input,
       .Select-value {
-        font-size: 1rem;
-        line-height: 40px;
-        top: 10px;
+        .padding--left--s;
+        .text--medium;
+        line-height: @size_2xl;
+        top: @size_s;
```
arsalansufi: `.padding--left--s = padding-left: 12px` Added to align with TextInput ([analogous value](https://github.com/Clever/components/blob/2bb27f80302a71cc7ccff822a38d15fb2304b653/src/TextInput/TextInput.less#L9))
`.text--medium = font-size: 1rem`
`@size_2xl = 40px`
`@size_s = 12px` Larger than the original value but identical to TextInput's [analogous value](https://github.com/Clever/components/blob/2bb27f80302a71cc7ccff822a38d15fb2304b653/src/TextInput/TextInput.less#L52)

Makefile
```diff
@@ -48,7 +48,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 201
+LINT_MAX_LESS_PROBLEMS := 189
```
kofi-clever: 🎉

PR #112, opened 2017-01-18
switching to CircleCI

nikhilpandit:
one thing to note here: I added `bin/` to `.npmignore` and removed `rm -rf bin` from the test steps (it was the last step in `test_drone.sh`)


.npmignore
```diff
@@ -1 +1,2 @@
 test/
+bin/
```
peternga: GH likes to have an empty line at the end

nikhilpandit: updated.

PR #118, opened 2017-02-04
Add TextArea component

peggyl:
Reference: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea

### DIFF
--- a/docs/components/TextAreaView.jsx
+++ b/docs/components/TextAreaView.jsx
@@ -1,0 +1,100 @@
+import React from "react";
+
+import Example from "./Example";
+import View from "./View";
+import {TextArea} from "src";
+
+import "./TextAreaView.less";
+
+export default class TextAreaView extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      disabled: false,
+      readOnly: false,
+      required: false,
+      inputValue: "",
+    };
+  }
+
+  render() {
+    const {cssClass} = TextAreaView;
+
+    return (
+      <View className={cssClass.CONTAINER} title="TextArea">
+        <Example
+          code={`
+            <TextArea
+              disabled={this.state.disabled}
+              name="TextAreaName"
+              label="TextArea Label"
+              readOnly={this.state.readOnly}
+              required={this.state.required}
+              spellCheck={this.state.spellCheck}
+              placeholder="TextArea Placeholder"
+              onChange={e => this.setState({inputValue: e.target.value})}
+              value={this.state.inputValue}
+            />
+          `}
+        >
+          <div>
+            <TextArea
+              disabled={this.state.disabled}
+              name="TextAreaName"
+              label="TextArea Label"
+              readOnly={this.state.readOnly}
+              required={this.state.required}
+              spellCheck={this.state.spellCheck}
+              placeholder="TextArea Placeholder"
+              onChange={e => this.setState({inputValue: e.target.value})}
+              value={this.state.inputValue}
+            />
+          </div>
+          <label className={cssClass.CONFIG}>
+            <input
+              type="checkbox"
+              checked={this.state.disabled}
+              onChange={({target}) => this.setState({disabled: target.checked})}
+            />
+            {" "}
+            Disabled
+          </label>
+          <label className={cssClass.CONFIG}>
+            <input
+              type="checkbox"
+              checked={this.state.readOnly}
+              onChange={({target}) => this.setState({readOnly: target.checked})}
+            />
+            {" "}
+            Read Only
+          </label>
+          <label className={cssClass.CONFIG}>
+            <input
+              type="checkbox"
+              checked={this.state.required}
+              onChange={({target}) => this.setState({required: target.checked})}
+            />
+            {" "}
+            Required
+          </label>
+          <label className={cssClass.CONFIG}>
+            <input
+              type="checkbox"
+              checked={this.state.spellCheck}
+              onChange={({target}) => this.setState({spellCheck: target.checked})}
+            />
+            {" "}
+            Spell Check
+          </label>
+        </Example>
+      </View>
+    );
+  }
+
+}
+
+TextAreaView.cssClass = {
+  CONFIG: "TextAreaView--config",
+  CONTAINER: "TextAreaView",
+};
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -1,0 +1,96 @@
+@import (reference) "../less/utilities";
+
+.TextArea {
+  .padding--left--s;
+  .padding--top--3xs;
+  .text--regular;
+  .border--s(@neutral_silver);
+  position: relative;
+  color: @neutral_black;
+  background: @neutral_white;
+  box-sizing: border-box;
+  width: 100%;
+}
+
+.TextArea--infoRow {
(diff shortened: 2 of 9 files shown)


src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
+@import (reference) "../less/colors";
+@import (reference) "../less/border";
+@import (reference) "../less/sizing";
+@import (reference) "../less/type-size";
+@import (reference) "../less/type-utilities";
+@import (reference) "../less/spacing";
+
+.TextArea {
+  .padding--left--s;
```
peggyl: Removes awkward whitespace on the right when a textbox is vertically but not horizontally resizable (the resize icon is on the bottom right).

src/less/index.less
```diff
@@ -4,3 +4,4 @@
 @import "./utilities";
 @import "./components";
 
+@import "../TextArea/TextArea";
```
peggyl: Added by `./bin/new_component` but should be unnecessary?

Makefile
```diff
@@ -48,7 +48,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 184
+LINT_MAX_LESS_PROBLEMS := 189
```
CarlNelson: What are the lint problems that required this increase?

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
+@import (reference) "../less/colors";
+@import (reference) "../less/border";
+@import (reference) "../less/sizing";
+@import (reference) "../less/type-size";
+@import (reference) "../less/type-utilities";
+@import (reference) "../less/spacing";
```
CarlNelson: You can just import ../less/utilities instead of multiple declarations.

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
...
+    }
+  }
+
+  .TextArea--input {
+    .text--medium;
+    display: block;
+    width: 100%;
+    border: 1px solid transparent;
+    resize: vertical;
+    line-height: @size_2xl;
```
CarlNelson: Line height is much too large. Use type utility for line height.

.text--line-height-3 for the input text.

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
...
+      position: absolute;
+      right: @size_l;
+    }
+  }
+
+  .TextArea--input {
+    .text--medium;
+    display: block;
+    width: 100%;
+    border: 1px solid transparent;
```
CarlNelson: user border mixin .border--s(transparent);

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
...
+
+  .TextArea--infoRow {
+    .text--tiny;
+    .text--caps;
+    position: absolute;
+    top: @size_xs;
+    width: 100%;
+    color: @neutral_dark_gray;
+
+    .TextArea--label {
```
CarlNelson: I believe the naming convention here should be: .TextArea--infoRow--label

@kofi-clever is that correct?

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
...
+    resize: vertical;
+    line-height: @size_2xl;
+    min-height: @size_2xl;
+    padding-top: @size_s;
+    background-color: transparent;
+    outline: none;
+  }
+
+  &.TextArea--placeholder-shown {
+    label { opacity: 0 !important; }
```
CarlNelson: Nit: There should be space after this.

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
...
+  &.TextArea--inFocus {
+    box-shadow: inset @size_2xs 0 @primary_blue;
+    transition: box-shadow .25s ease-out;
+  }
+
+  &.TextArea--disabled {
+    .border--s(@neutral_silver);
+    color: @neutral_dark_gray;
+    background: @neutral_silver;
+    .TextArea--input {
```
CarlNelson: Nit: should have a space before it.

src/TextArea/TextArea.less
```diff
@@ -0,0 +1,98 @@
+@import (reference) "../less/colors";
+@import (reference) "../less/border";
+@import (reference) "../less/sizing";
+@import (reference) "../less/type-size";
+@import (reference) "../less/type-utilities";
+@import (reference) "../less/spacing";
+
+.TextArea {
```
CarlNelson: Overall you should remove nesting except for explicit states. Most of these don't need to be nested since they are correctly using BEM.

docs/components/TextAreaView.jsx
```diff
@@ -0,0 +1,97 @@
...
+              required={this.state.required}
+              spellCheck={this.state.spellCheck}
+              placeholder="TextArea Placeholder"
+              onChange={e => this.setState({inputValue: e.target.value})}
+              value={this.state.inputValue}
+            />
+          `}
+        >
+          <div>
+            <TextArea
```
CarlNelson: Pass a label to the demo.

src/TextArea/TextArea.less
peggyl: Where should the space be?

peggyl: Oh, do you mean new line for opacity? I copied this from TextInput.less but it looks like Select.less put it on its own line.

src/TextArea/TextArea.less
kofi-clever: since this is a label for the TextArea  - and it will only ever have one label - it's fine to have it as `TextArea--label`.
(if there was an InfoRow component that abstracted this stuff away, then it would have to be `TextArea--InfoRow--label`)

but you should move these nested rules out into the top-level scope as Carl pointed out above.

src/TextArea/TextArea.less
kofi-clever: yeah, looks too exaggerated:
screen shot 2017-02-06 at 12 24 18 pm

i would maybe even just leave it as the browser default. unless does that look too cramped?

peggyl: This is the browser default:

screenshot 2017-02-06 14 13 14

peggyl: Definitely crowds into the label. 

screenshot 2017-02-06 14 22 59

peggyl: @CarlNelson I changed line-height to `.text--line-height-3`, which caused the crowding into the label shown above. Offset by adding `.padding--top--l` on the input—is this okay?

CarlNelson: That should be fine.

src/TextArea/TextArea.less
CarlNelson: Yeah, new line here.

Makefile
peggyl: It's actually back down to 184 after incorporating the feedback in this PR. They were probably initially due to me copying a lot of the styles in TextInput.less.

peggyl:
@CarlNelson I think I've addressed all of your comments, if you'd like to take another look. Thanks!

PR #121, opened 2017-02-10
add row click function to table

eighnjel:
- Mischelle wants to update the table component's design.
- Add onRowClick and hover states.

### DIFF
--- a/src/Table/Table.jsx
+++ b/src/Table/Table.jsx
@@ -69,6 +69,7 @@ export class Table extends Component {
       pageSize,
       paginated,
       rowIDFn,
+      onRowClick,
     } = this.props;
     const {currentPage, sortState = initialSortState} = this.state;
     const {cssClass} = Table;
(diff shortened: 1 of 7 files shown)


src/Table/Table.jsx
```diff
@@ -125,7 +126,11 @@ export class Table extends Component {
               </Cell>
             </tr>
             ) : pages[displayedPageIndex].map(rowData => (
-            <tr className={cssClass.ROW} key={rowIDFn(rowData)}>
+            <tr
+              className={classnames(cssClass.ROW, {rowClickEnabled: onRowClick})}
+              key={rowIDFn(rowData)}
+              onClick={onRowClick}
```
amir-clever: do you need to pass the row id?

`onClick={() => onRowClick(rowIDFn(rowData))}`
or
`onClick={(event) => onRowClick(event, rowIDFn(rowData))}`

The first seems sufficient right? Not clear with the component convention is. @osdiab any preferences / thoughts?

PR #123, opened 2017-02-21
Tweak TextArea component

arsalansufi:
This PR makes some small tweaks to the new `TextArea` component:

*Before*

before

*After*

after-1
after-2

Regarding the expander (in the bottom-right corner of the component), it isn't misaligned in the [demo](http://clever.github.io/components/#/components/text-area) because the demo page appears to apply `box-sizing: border-box` globally.

### Update

*Post PR Review GIF*

updated

### DIFF
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -25,6 +25,7 @@
   left: @size_s;
   opacity: 1;
   background-color: @neutral_white;
+  padding-right: @size_3xs;
 }
 
 .TextArea--required {
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -44,6 +44,7 @@
   display: block;
   width: 100%;
   .border--s(transparent);
+  border-left: @size_none;
   .text--line-height-3;
   .padding--top--l;
   padding-left: @size_4xs;
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -41,7 +41,15 @@
   display: block;
   width: 100%;
   .border--s(transparent);
+
+  /**
+   * Use a transparent border to position the TextArea input instead of padding or a margin because
+   *  1. Text is still visible in the padding region.
+   *  2. The underlying textarea element overwrites margin values when the user adjusts the
+   *     element's height.
+   */
   border-top-width: @size_l;
+
   .text--line-height-3;
   padding: @size_none;
   background-color: transparent;
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -19,23 +19,23 @@
   top: @size_xs;
   width: 100%;
   color: @neutral_dark_gray;
+}
 
-  .TextArea--label {
-    left: @size_s;
-    opacity: 1;
-    background-color: @neutral_white;
-  }
+.TextArea--label {
+  left: @size_s;
+  opacity: 1;
+  background-color: @neutral_white;
+}
 
-  .TextArea--required {
-    position: absolute;
-    right: @size_l;
-  }
+.TextArea--required {
+  position: absolute;
+  right: @size_l;
+}
 
-  .TextArea--error {
-    color: @alert_red;
-    position: absolute;
-    right: @size_l;
-  }
+.TextArea--error {
+  color: @alert_red;
+  position: absolute;
+  right: @size_l;
 }
 
 .TextArea--input {
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -23,6 +23,7 @@
   .TextArea--label {
     left: @size_s;
     opacity: 1;
+    background-color: @neutral_white;
   }
 
   .TextArea--required {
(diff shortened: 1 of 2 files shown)


src/TextArea/TextArea.less
```diff
@@ -44,9 +45,11 @@
   .border--s(transparent);
   .text--line-height-3;
   .padding--top--l;
+  padding-left: @size_4xs;
```
arsalansufi: This change, which isn't immediately apparent in the screenshots, is super subtle. It tweaks the left alignment of text in the component to match the left alignment of text in the `TextInput` component. (It was previously off by 1 pixel.)

src/TextArea/TextArea.less
```diff
@@ -79,6 +82,10 @@
   color: @neutral_dark_gray;
   background: @neutral_silver;
 
+  .TextArea--label {
```
CarlNelson: We should avoid nesting in basically all cases. Any way we can clean this up throughout this less file?

arsalansufi: 👍 
Happy to restructure things a bit, removing nesting wherever possible

arsalansufi: Hmm, actually, this specific case of nesting seems pretty necessary since it's a style that should only be applied when the component is disabled

CarlNelson: Cool. Thanks

src/TextArea/TextArea.less
```diff
@@ -19,34 +19,39 @@
...
 
-  .TextArea--label {
-    left: @size_s;
-    opacity: 1;
-  }
+.TextArea--label {
+  left: @size_s;
+  opacity: 1;
+  background-color: @neutral_white;
+  padding-right: @size_3xs;
```
arsalansufi: This right padding plus the removal of the left border on [line 47](https://github.com/Clever/components/pull/123/files#diff-f33c3b7a08d6659c4f1c4293d40f34cbR47) make the left alignments of text in the `TextArea` and `TextInput` components match.

src/TextArea/TextArea.less
```diff
@@ -19,34 +19,39 @@
...
+  position: absolute;
+  right: @size_l;
 }
 
 .TextArea--input {
   .text--medium;
   display: block;
   width: 100%;
   .border--s(transparent);
+  border-left: @size_none;
```
arsalansufi: This left border removal plus the left padding change on [line 50](https://github.com/Clever/components/pull/123/files#diff-f33c3b7a08d6659c4f1c4293d40f34cbR50) make the left alignments of text in the `TextArea` and `TextInput` components match. These alignments used to be off by 2 pixels.

osdiab:
A design question—why is it better for the title of the input to overlap with the contents of the input instead of having the label just take up the whole width and the text to be below it? Still looks weird to me to have `DESCRIPTION` eclipse the inputted text.

CarlNelson:
I tried making a full-width background with a drop shadow which matches your idea @osdiab but ran into some sizing issues that made it a challenge to get it perfectly aligned.

But I only spent about 15 minutes on it. Could be solved.

arsalansufi:
^ Happy to work on this! If I can't get it solved in a reasonable amount of time, I'll go ahead and merge this (as there are other fixes in this PR that I'd like to have merged)

osdiab:
That would be awesome if you have the time!

I think there are two main options: do a drop shadow kind of thing, or just let the description waste a line of space on top. The latter option is simpler and probably would look fine given that it already takes up that space in the single line case, if I'm not mistaken.

I'm imagining that the markup of the `TextInput` ought to look roughly like:

```
<div class="TextInput">
  <header>
    <label for="input_id">Description</label>
    <span>Required</span>
    // etc
  </header>
  <input id="input_id" /> // Aside: currently we do this wrong, using name instead of id (label is specced for id, I believe)
</div>
```

and that `.TextInput > header` would just be positioned vertically above the input element to achieve the simple "header takes up the width and wastes a line" simple scenario.

To do the drop shadow thing upon it getting taller, I would imagine you need some javascript to do that if the vertical height of the text spans more than 1 line. Not sure that's worth the effort though, what do you think @CarlNelson ?

src/TextArea/TextArea.less
```diff
@@ -19,34 +18,36 @@
...
+  position: absolute;
+  right: @size_l;
 }
 
 .TextArea--input {
   .text--medium;
   display: block;
   width: 100%;
   .border--s(transparent);
+  border-top-width: @size_l;
```
arsalansufi: Regarding why I've used a transparent border to position the `TextArea` input below the label instead of padding or a margin:

- Padding doesn't work because text is still visible in the padding region.
- Margins don't work because the underlying `textarea` element overwrites margins (and height) when the expander is used to change the height of the element.

src/TextArea/TextArea.less
```diff
@@ -78,10 +78,6 @@
   .border--s(@neutral_silver);
   color: @neutral_dark_gray;
   background: @neutral_silver;
-
-  .TextArea--input {
```
arsalansufi: Unneeded because the `TextArea` input is already [transparent](https://github.com/Clever/components/pull/123/files#diff-f33c3b7a08d6659c4f1c4293d40f34cbR47)

src/TextArea/TextArea.less
osdiab: Clever :) May be worth a comment to explain that?

arsalansufi: Comment added!

PR #126, opened 2017-03-01
TextInput: allow type=email

peggyl:
I'm building forms with email fields, and would like the input validation that comes "for free" with `<input type=email />` in HTML 5. 

Email is one of the few types (other than password) for the input element that makes sense as a text input: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input 

I'm open to whitelisting `search` and `url` too, but don't have concrete use cases for those right now.


peggyl:
This also doesn't seem to work "out of the box" without doing a form submit (or hack). 😞

peggyl:
Closing in favor of writing validation functions.

PR #127, opened 2017-03-02
Inline-block button to match base buttons

CarlNelson:
**Overview:**
ConfirmationButton was block element instead of inline-block giving it a different layout behavior from all other buttons. Also, default Cancel is now a link button.

**Screenshots/GIFs:**
Old - Two confirm buttons next to each other
image

New - Two confirm buttons next to each other
image

Old - cancel button
image

New - cancel button
image

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.18.2",
+  "version": "0.18.3",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


osdiab:
LGTM!

PR #130, opened 2017-03-05
Add a Label component

kofi-clever:
**Jira:** https://clever.atlassian.net/browse/FEE-42

**Overview:**
Adding a new Label (pill) component with color, sizing and optional tooltip support.

**Screenshots/GIFs:**
label

screen shot 2017-03-05 at 2 15 15 pm

**Testing:**
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/LabelView.jsx
+++ b/docs/components/LabelView.jsx
@@ -1,0 +1,199 @@
+import _ from "lodash";
+import loremIpsum from "lorem-ipsum";
+import React, {Component} from "react";
+
+import Example from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+import {Button, Label, SegmentedControl, Tooltip} from "src";
+
+import "./LabelView.less";
+
+
+export default class LabelView extends Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      color: Label.Color.SUCCESS,
+      tooltipPlacement: Tooltip.Placement.RIGHT,
+      tooltipTextAlign: Tooltip.Align.LEFT,
+      size: Label.Size.S,
+    };
+  }
+
+  render() {
+    const {cssClass} = LabelView;
+    const {color, size, tooltipPlacement, tooltipTextAlign} = this.state;
+
+    return (
+      <View className={cssClass.CONTAINER} title="Label">
+        <Example
+          code={`
+            <h3 className={cssClass.EXAMPLE}>
+              Labels are inline by default
+              <Label className={cssClass.LABEL} color={color} size={size}>
+                Simple Label
+              </Label>
+            </h3>
+            <h3 className={cssClass.EXAMPLE}>
+              And can vary in size and color
+              <Label className={cssClass.LABEL} color={color} size={size}>
+                With HTML Content <span className={"fa fa-check-circle"} />
+              </Label>
+              <Label
+                className={cssClass.LABEL}
+                color={color}
+                ref="focusableLabel"
+                size={size}
+                tooltip={
+                  <div>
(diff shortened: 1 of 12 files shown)


src/Label/Label.jsx
```diff
@@ -0,0 +1,104 @@
+import _ from "lodash";
+import classnames from "classnames";
+import React, {PropTypes, PureComponent} from "react";
+
+import {Tooltip} from "../Tooltip";
+
+import "./Label.less";
+
+
+export default class Label extends PureComponent {
```
osdiab: Cool, I didn't know `PureComponent` existed :)

src/Label/Label.jsx
```diff
@@ -0,0 +1,104 @@
...
+    const label = (
+      <span
+        className={classnames(
+          cssClass.CONTAINER,
+          color && cssClass.color(color),
+          size && cssClass.size(size),
+          tooltip && cssClass.WITH_TOOLTIP,
+          className
+        )}
+        ref="label"
```
osdiab: Do we prefer string name refs as opposed to function refs? like

```
ref={(elem) => this.label = elem}}
```

src/Label/Label.jsx
```diff
@@ -0,0 +1,104 @@
...
+
+    return (
+      <Tooltip content={tooltip} placement={tooltipPlacement} textAlign={tooltipTextAlign}>
+        {label}
+      </Tooltip>
+    );
+  }
+}
+
+Label.Color = {
```
osdiab: Thinking to the future, I feel like we should make these just non-default exported consts rather than members of the actual Label class, because Typescript will almost certainly yell at us for assigning nonexistent members; and I don't think it's any more clear this way either.

But given we're not doing that yet, and I don't think clients really consume this stuff, it's no big deal

docs/components/LabelView.jsx
```diff
@@ -0,0 +1,199 @@
+import _ from "lodash";
```
osdiab: Are we adding `README` docs?

src/Label/Label.jsx
```diff
@@ -0,0 +1,104 @@
...
+      <span
+        className={classnames(
+          cssClass.CONTAINER,
+          color && cssClass.color(color),
+          size && cssClass.size(size),
+          tooltip && cssClass.WITH_TOOLTIP,
+          className
+        )}
+        ref="label"
+        tabIndex={tooltip ? 0 : undefined}
```
osdiab: I kinda wonder if `tabIndex` is something `Tooltip` should be handling

src/Label/Label.jsx
```diff
@@ -0,0 +1,104 @@
...
+    const {cssClass} = Label;
+    const {
+      children,
+      className,
+      color,
+      size,
+      tooltip,
+      tooltipPlacement,
+      tooltipTextAlign,
+      ...passthroughProps,
```
osdiab: So `passthroughProps` is hitting the underlying `span` - maybe we should clarify that these are specifically `span` properties, and how do they interact with the existing `tabIndex` and `className` ones?

docs/components/LabelView.jsx
kofi-clever: started adding them inline with the live examples (e.g. `<PropDocumentation>` below).
just added an introductory paragraph as well just as a brief explanation of how it can be used - https://github.com/Clever/components/pull/130/commits/3122c8896832d746c4981bc3acaed874bcef771a

PR #133, opened 2017-03-13
Don't create style.css file

alsmola:
**Jira:**
[SEC-436](https://clever.atlassian.net/browse/sec-436)

**Overview:**
This change removes the `style.css` file, so that it doesn't get added to `https://assets.clever.com/design/style.css`, so that we can start caching all assets for a year in the users browsers (see https://github.com/Clever/ci-scripts/pull/23).

I can't tell anywhere were the `style.css` file is being used, so this *should* be a no-op.

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,6 @@ build: clean es5 styles docs
 styles:
 	@echo "Building stylesheet"
 	@$(WEBPACK) --config webpack_styles.config.js
-	@cp dist/css/style* dist/css/style.css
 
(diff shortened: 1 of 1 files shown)


CarlNelson:
Yeah, should be good to go. We had originally intended on doing something with style.css but with the npm package people now just import the less portion of clever-components

CarlNelson:
LGTM

PR #134, opened 2017-03-14
text-decoration none on tab hover

osdiab:
**Overview:**

Clever common sets `a:hover` to have `text-decoration: underline`, which causes tabs to also underline the text as well as the tab itself on hover. Explicitly say not to underline it, so that even on hover/focus states, the style remains the same. 

### Before

screen shot 2017-03-14 at 16 49 00

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.18.7",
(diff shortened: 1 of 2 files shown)


ahsojar:
LGTM

PR #135, opened 2017-03-15
leftnav polish

cozmo:
Polish the left nav
- Give it `focus` styles so you can tell where your keyboard highlight is at. 
- Add animations when opening/closing the drawer. 
- Allow passing in a component or an `href` so we aren't using `onClick` for everything. 

**Testing:**
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

### DIFF
--- a/src/LeftNav/LeftNav.less
+++ b/src/LeftNav/LeftNav.less
@@ -26,6 +26,7 @@
 .NavLink {
   font: inherit;
   background: inherit;
+  /* stylelint-disable-next-line */
   color: inherit;
   .flexbox;
   width: 100%;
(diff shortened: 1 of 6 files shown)


src/LeftNav/LeftNav.less
```diff
@@ -105,15 +127,15 @@
...
+  .animateWidth
 }
 
-.animateOpacity {
-  .delayMoment;
-  .timingPromptly;
-  .transition-timing-function(@motionSnappy);
-  transition-property: opacity;
+.LeftNav--drawer-anim-enter.LeftNav--drawer-anim-enter-active {
+  width: 12rem;
```
arsalansufi: Super minor but I know we've been trying to use the size macros whenever possible. [`@size_s`](https://github.com/Clever/components/blob/f6793755a4c5f2fb41fe02855cd164aee3c83f96/src/less/sizing.less#L13) would work here

arsalansufi: Oops, I take that back! @size_s = 12px, not 12rem

PR #137, opened 2017-03-17
More leftnav improvements

cozmo:
Turns out firefox has some bugs with buttons and flexbox. The suggested way to fix this is to put a div inside the button, and put flexbox on that. This does that. 

It also sets the color of the links explicitly because in SD1 they're inheriting some weird styles. 

Have confirmed this looks good in Chrome, IE, FF and Safari.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.18.10",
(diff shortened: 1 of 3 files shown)


potluck:
LGTM

PR #139, opened 2017-03-22
prevent nested prop values from being overwritten

natebrennand:
**Overview:**
Depending on the order of props provided to a component that leverages nested components, a nested property may be overwritten.

**Testing:**
- [x] Unit tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.20.1",
+  "version": "0.21.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.20.0",
```
cozmo: This could be a patch I feel like?

natebrennand:
CC @osdiab

package.json
natebrennand: Added a behavior change w/ https://github.com/Clever/components/pull/139/commits/8e595f2fa8205c27c2f915d3f2f77e3e78b49597 which I think warrants minor

src/ConfirmationButton/ConfirmationButton.jsx
```diff
@@ -35,16 +37,17 @@ export class ConfirmationButton extends React.Component {
...
+  {
     onConfirm: PropTypes.func,
-  }
+  },
 );
 
-ConfirmationButton.defaultProps = Object.assign(
-  {}, prefixKeys(Button.defaultProps, "confirmButton"), ModalButton.defaultProps, {
+ConfirmationButton.defaultProps = Object.assign({},
+  prefixKeys(Button.defaultProps, propPrefix), ModalButton.defaultProps,
```
amir-clever: Nate' requests Nate' to add a newline

PR #142, opened 2017-03-23
Table onviewchange

samfishman:
I want to be able to lazily load additional data for table rows as those rows are displayed on-screen. That's technically possible today using the `onPageChange` and `onSortChange` callbacks, but it's annoying to have to keep track of sort and page and calculate what rows are visible yourself. Instead, I think a better interface is a callback that just tells you what rows are visible whenever that changes.

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/bundle.js
+++ b/docs/bundle.js
@@ -58,123 +58,123 @@
 
 	var _ButtonView2 = _interopRequireDefault(_ButtonView);
 
-	var _ColorsView = __webpack_require__(613);
+	var _ColorsView = __webpack_require__(612);
 
 	var _ColorsView2 = _interopRequireDefault(_ColorsView);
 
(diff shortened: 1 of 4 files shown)


potluck:
can you include a screenshot of the console after you click new page?

potluck:
looks really good!

samfishman:
onviewchange

potluck:
OK, so just want to confirm - do we think that returning the indices to the data is sufficient? just curious whether we think it could be more valuable to return an array of objects rather than indices.

Regardless, LGTM, but curious whether you have thoughts

samfishman:
I made it return the whole row object, not just the indices (it's just that the demo only prints the indices so as to not spam the console).

PR #144, opened 2017-03-23
Add custom actions for individual Wizard steps

arsalansufi:
**Overview:**

This PR adds the ability to define custom actions for individual `Wizard` steps. For a project that I'm currently working on, I need to store data in Mongo at intermediate steps in a `Wizard`.

**Screenshots/GIFs:**

wizard-update

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/Wizard/Wizard.jsx
+++ b/src/Wizard/Wizard.jsx
@@ -42,12 +42,12 @@ export class Wizard extends React.Component {
   }
 
   nextStepHandler() {
-    const {steps, onComplete} = this.props;
+    const {onComplete, steps} = this.props;
     const {currentStep, data} = this.state;
 
     // If an onComplete handler is provided for the current step, and it returns false, we won't
     // proceed to the next step.
-    if (steps[currentStep].onComplete && !steps[currentStep].onComplete(data)) {
+    if (steps[currentStep].onStepComplete && !steps[currentStep].onStepComplete(data)) {
       return;
     }
     if (currentStep === steps.length - 1) {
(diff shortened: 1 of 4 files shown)


src/Wizard/Wizard.jsx
```diff
@@ -194,6 +202,7 @@ Wizard.propTypes = {
     validate: PropTypes.func.isRequired,
     help: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
     props: PropTypes.object,
+    onComplete: PropTypes.func,
```
amir-clever: Can we rename `OnComplete` to `OnStepComplete`?

src/Wizard/Wizard.jsx
```diff
@@ -42,11 +42,19 @@ export class Wizard extends React.Component {
...
 
   nextStepHandler() {
-    if (this.state.currentStep === this.props.steps.length - 1) {
-      this.props.onComplete(this.state.data);
+    const {steps, onComplete} = this.props;
+    const {currentStep, data} = this.state;
+
+    // If an onComplete handler is provided for the current step, and it returns false, we won't
+    // proceed to the next step.
+    if (steps[currentStep].onComplete && !steps[currentStep].onComplete(data)) {
```
amir-clever: From a usage perspective, it seems like it might be good to take an error message string as the response of onStepComplete and then above the buttons populate an a alert (which is not part of Clever components?)

`errorMsgStr = onStepComplete(data)`

What are your thoughts on this?

arsalansufi: Down the road, it'd be great if we built this error handling into the component, but we haven't yet ported the `ToastNotification` into Clever components, which is our primary method of surfacing errors (at least in the Apps Dashboard).

For now, I think that it's okay to keep this general. As a result, the values provided to `onStepComplete` will look very similar to the `onChange` handlers provided to buttons, etc. (here's an [example](https://github.com/Clever/apps-dashboard/blob/bc30edf1e78401cd3a56d6c880fb71e6e67475fe/pages/partner/applications/ui/components/panels/UpdateFooter.jsx#L31-L55)). The `onStepComplete` values will just have added `return [true|false]` statements.

Happy to talk through this in person if you have additional thoughts!

src/Wizard/Wizard.jsx
arsalansufi: https://github.com/Clever/components/pull/144/commits/cb9f9e204636650a41024b5c1a809badb751d1c0 Great suggestion

PR #145, opened 2017-03-24
Compound form

mischelle-m:
**Jira:**
https://clever.atlassian.net/browse/DESIGN-182

**Overview:**
Compound form layout guidelines page. 

**Screenshots/GIFs:**
http://g.recordit.co/YSmQLEmVXa.gif


docs/components/LayoutCompoundForm.jsx
```diff
@@ -0,0 +1,319 @@
...
+                </p>
+              </div>
+            </Col>
+          </Row>
+          <Row grow className="margin--bottom--xl">
+            <Col span={12} className="flexbox">
+              <div>
+                <h3>Do's and Don'ts</h3>
+                <p>
+                 Below is a working list of simple guidelines to help ensures that your form-design will turn out 🔥!
```
amir-clever: 💯

docs/index.html
```diff
@@ -2,7 +2,7 @@
 <html>
   <head>
     <title>Clever Design System</title>
-    <link rel="shortcut icon" href="favicon.ico?1" >
+    <link rel="shortcut icon" href="favicon.ico" >
```
mischelle-m: I'm curious about this too! With the `?1`, when I replaced the favicon with the new one, it wouldn't reflect the changes.

docs/components/SideBar/SideBar.jsx
```diff
@@ -20,6 +20,7 @@ export default function SideBar({className}) {
         <NavLink href="/design/typography">Typography</NavLink>
         <NavLink href="/design/sizing">Sizing</NavLink>
         <NavLink href="/design/page-layout">Page Layout</NavLink>
+        <NavLink href="/design/compound-form">Compound Form</NavLink>
```
mischelle-m: That's a good idea. I think initially we thought of putting Compound Form as a sub-nav under Page Layout. cc: @CarlNelson -- should we still do this?

CarlNelson: They're ranked by order of importance in the design section instead of alphabetical @amir-clever.

If we aren't sub-navving the layouts, we might want to make a Layout section similar to how we have LESS section.

mischelle-m: Gotcha. Should we do that now? Or maybe better to wait till we have more Layout guidelines written out?

amir-clever: I prefer alphabetical when I'm trying to get an overview. But I guess this is out of scope for this PR. Thanks for the clarification.

docs/index.html
kofi-clever: Yeah, it's basically just a cache buster. Favicons are cached by browsers so adding a different query param allows us to prompt the browser to re-download. You can just increment the count if you're updating the icon

PR #146, opened 2017-03-24
Number

kofi-clever:
**Jira:** https://clever.atlassian.net/browse/FEE-58

**Overview:**
Adding a `Number` component for consistently formatting (and optionally shortening) long numbers for display.

**Screenshots/GIFs:**
number

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,5 @@
+import Number from "./Number";
+
 export * from "./flex";
 export * from "./TabBar";
 export {Modal} from "./Modal/Modal";
@@ -19,3 +21,4 @@ export {Tooltip} from "./Tooltip";
 export {Label} from "./Label";
 export {FileInput} from "./FileInput/FileInput";
 export * from "./InfoPanel";
+export {Number};
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,6 @@
+// Number is a default export.
+// TODO(kofi): Convert all components to default exports to allow clients to import separate
+// components without pulling in the whole repo.
 import Number from "./Number";
 
 export * from "./flex";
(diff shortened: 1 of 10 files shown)


src/index.js
```diff
@@ -19,3 +21,4 @@ export {Tooltip} from "./Tooltip";
 export {Label} from "./Label";
 export {FileInput} from "./FileInput/FileInput";
 export * from "./InfoPanel";
+export {Number};
```
ahsojar: haven't written a component before, why can't we do `export {Number} from "./Number"` like we do for the lines above?

test/Number_test.jsx
```diff
@@ -0,0 +1,38 @@
...
+  it("supports non-React formatting", () => {
+    assert.equal("5,123", Number.format(5123));
+    assert.equal("6.8M", Number.format("6789345", true));
+  });
+
+  it("throws for non-numeric input", () => {
+    assert.throws(() => shallow(<Number>nope</Number>));
+    assert.throws(() => Number.format("nope", true));
+  });
+});
```
ahsojar: Nice tests 🎉

src/index.js
kofi-clever: it's a quirk of the es6 import/export pattern. `Number` is currently a default export, so it has to be imported as a default here and then exported as a secondary. (the others are exported as secondary imports, but i'm toying with the idea of providing default exports so clients can import individual components without pulling all of the components in)

ahsojar: ahh gotcha. perhaps worth adding a comment for `Number` since its a special case!

test/Number_test.jsx
kofi-clever: +1

PR #148, opened 2017-03-27
Tweak Wizard onStepComplete handlers to be async

arsalansufi:
**Overview:** I recently introduced the ability to define custom actions for individual `Wizard` steps, what I'm calling `onStepComplete` handlers ([PR 144](https://github.com/Clever/components/pull/144)). I'm realizing that these `onStepComplete` handlers will often need to be asynchronous. I've updated the `Wizard` component to handle this asynchronicity.

**Testing:** Tested manually by modifying [this line of code](https://github.com/Clever/components/blob/dcc789b7070405989e2a09edd6222ae05913eb09/docs/WizardExample.jsx#L49).

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


osdiab:
This is a breaking change, so bump version accordingly!

PR #150, opened 2017-03-30
Lazy table

samfishman:
Adds lazy-loading support to the Table component. This is enabled by setting the `lazy` prop to true on the table. Then:
* The total number of rows must be specified in the `numRows` prop.
* Data is retrieved from the `getData(query: {startingAfter: any, pageSize: number}): Promise<Array<Object>>` prop.
* The `initialPage` and `filter` props may not be set.

Data is loaded from `getData` as the user goes to pages that have yet to be seen. Data is kept cached so that if a previously-loaded page is visited (e.g. the user hits the "previous page" button), `getData` doesn't need to be called. It may make sense to look into expiring rows from this cache so that memory usage doesn't grow unbounded. Fortunately, the user of this component need not worry about that - caching is an implementation detail here and can be supported without changing the API. Another potential area for improvement is pre-fetching - we could try to load the next page before the user actually goes to it.

A few other behavioral details:
* When a page is loading, the current page is left on-screen until the new page is ready.
* When a user clicks on a page number ahead of the current page, all pages between the current page and that page are fetched. This is because we use `startingAfter`-style paging.
* The last page number isn't shown because clicking it could result in a metric ton of pages being fetched.

**Demo:**
lazy-table

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


src/Table/Table.jsx
```diff
@@ -19,29 +19,97 @@ export class Table extends Component {
...
+        onViewChange(displayedData);
       }
     });
   }
 
+  async _fetchLazy(pageIdx) {
+    const {pageSize, getData, rowIDFn} = this.props;
+    const {lazyPages} = this.state;
+
+    if (pageIdx < lazyPages.length) {
```
ahsojar: nit: Add a comment to explain what we're checking for

src/Table/Table.jsx
```diff
@@ -19,29 +19,97 @@ export class Table extends Component {
...
+      return;
+    }
+
+    this.setState({pageLoading: true});
+
+    const newPages = lazyPages.slice(0);
+    for (let idx = lazyPages.length; idx <= pageIdx; idx++) {
+      const query = {pageSize};
+      if (idx !== 0) {
```
ahsojar: Comment here as well

ahsojar: and here is well

src/Table/Table.jsx
ahsojar: nit: Add a comment to explain why we're doing this check

src/Table/Footer.jsx
```diff
@@ -51,64 +51,73 @@ export default function Footer({currentPage, numColumns, numPages, onPageChange}
...
-                className={cssClass.BUTTON_PAGE}
-                key={numPages}
-                onClick={() => selectPage(numPages)}
-                type="linkPlain"
-                value={`${numPages}`}
-              />
+          <div className={cssClass.LOADING_CONTAINER}>
+            {isLoading && (
+              <div className={cssClass.LOADING_OUTER}>
+                <div className={cssClass.LOADING_INNER}>Loading...</div>
```
ahsojar: Why do we need the nested outer/inner divs here?

src/Table/Footer.jsx
```diff
@@ -120,10 +129,13 @@ Footer.propTypes = {
   onPageChange: PropTypes.func,
   numColumns: PropTypes.number.isRequired,
   numPages: PropTypes.number.isRequired,
+  showLastPage: PropTypes.bool,
+  isLoading: PropTypes.bool,
 };
 
 Footer.defaultProps = {
   onPageChange: () => {},
+  showLastPage: true,
```
ahsojar: just checking, why do we want the default value to be true on this?

Makefile
```diff
@@ -47,7 +47,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 140
+LINT_MAX_LESS_PROBLEMS := 139
 lint:
```
ahsojar: 🎉

samfishman: Unclear how this happened since I didn't edit any existing less haha.

src/Table/Footer.jsx
samfishman: It's to make the styling look right: the outer div has `position: absolute` so that we can put it above the numbers and aligns text to the center. The inner div has `display: inline-block` so that it gets affected by the outer div's centering.

src/Table/Footer.jsx
samfishman: No good reason besides that that was the previous behavior - if the tables not lazy we want to show the last page.

src/Table/Table.jsx
samfishman: 👍

src/Table/Table.jsx
samfishman: 👍

src/Table/Footer.jsx
ahsojar: ah ok, makes sense. can we not use the `LOADING_CONTAINER` div to help with positioning? It seems strange to have 3 nested divs to display `Loading...`

samfishman: Haha agreed. I couldn't get the CSS-fu to work out, but I'll give it another shot. Feel free to try yourself too :P.

src/Table/Footer.jsx
ahsojar: gotcha. For some reason I thought we decided in #frontend to keep the behavior consistent between the 2 tables

Makefile
ahsojar: lol just roll with it

src/Table/Footer.jsx
samfishman: we did, but carl conceded that it's ok to not show the last page since it would result in a ton of fetches (you'd basically be paging through all the data)

ahsojar:
Thanks for updating the comments! Before you merge, I would:
1) See if you can simplify the `Loading` CSS
2) Make sure it looks good on all 3 browsers
3) Remember to bump the version

samfishman:
Got the `Loading...` CSS cleaned up - thanks for the gentle encouragement there.

PR #153, opened 2017-04-05
TextArea: display "optional" input note

peggyl:


peggyl:
@CarlNelson The `optional` label will always be displayed for a non-required field with the current implementation. I considered only showing it if the user explicitly passes in a prop, but having both `required` and `optional` props seemed confusing.

CarlNelson:
I think in most cases displaying `optional` is redundant as `required` is what people usually need to know. Only in specific circumstances is displaying `optional` ideal.

peggyl:
`make lint` passes locally (which is to say, I have 178 stylelint problems, while CircleCI has 177), which is odd.

peggyl:
@CarlNelson I added an `optional` prop, and updated the documentation to hopefully make this as not-confusing as possible given there is also a `required` prop. Let me know what you think.

README.md
```diff
@@ -306,12 +306,15 @@ This component is a `text area` that lets the user input multiple lines of text.
...
 | name (required) | String | Name for input element | None
 | onChange (optional) | Function | Called when value of input changes | None
+| optional (optional) | Bool | Marks input as optional and adds indicator | False
 | placeholder (optional) | Node | Placeholder node for input | None
 | readOnly (optional) | Bool | Sets element as read-only | false
 | required (optional) | Bool | Marks input as required and adds indicator | false
 | spellCheck (optional) | Bool | Turns on spelling/grammar check for input | false
 | value (optional) | Node | Value of input | None
 
+Note: use the `required` prop to add a "Required" input label on a required field and the `optional` prop to add an "Optional" input label on an optional field. By default, `required` and `optional` are both `false`, which means that no label is shown.
```
CarlNelson: While you're in here can you move this documentation to TextAreaView in the docs?

peggyl: Does this work?

screenshot 2017-04-05 11 09 01

kofi-clever: i would put any general description at the top before the demo, and then use a [`<PropDocumentation />`](https://github.com/Clever/components/blob/master/docs/components/PropDocumentation.jsx) instance below the demo to display the props.

CarlNelson:
Otherwise LGTM

README.md
peggyl: Whoa, TIL about `<PropDocumentation />`. 👌

PR #155, opened 2017-04-07
Lazy table length unknown

samfishman:
**Overview:**
The lazy table is nice, but sometimes we want a lazy table and we don't even know how many total items will exist in the table. This adds support by that for handling the case where `numRows` isn't defined.

**Screenshots/GIFs:**
lazy-table-2

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Table/Table.jsx
+++ b/src/Table/Table.jsx
@@ -25,22 +25,30 @@ export class Table extends Component {
           console.error(`Table: prop "${p}" may not be set if "lazy"`);
         }
       }
-      for (const p of ["getData", "numRows"]) {
+      for (const p of ["getData"]) {
         if (!props[p]) {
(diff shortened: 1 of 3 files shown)


src/Table/Table.jsx
```diff
@@ -25,22 +25,30 @@ export class Table extends Component {
           console.error(`Table: prop "${p}" may not be set if "lazy"`);
         }
       }
-      for (const p of ["getData", "numRows"]) {
+      for (const p of ["getData"]) {
```
kofi-clever: nit: can this be simplified to just an `if` now?

samfishman: 👍

PR #156, opened 2017-04-11
Add optional rowClassNameFn prop to Table

peggyl:
Supports (optional) conditional row styling by applying a function that dynamically applies a new className to a `<tr>` based on the row data. Tested with `make dev-server` locally and verified that the expected data had the desired className with Inspect Element.

I chose not to apply more styles to the example/demo table since it's a pretty complex example already, and this is an optional prop not required in the general use case, but you can inspect the element and see the class name.

### DIFF
--- a/src/Table/Table.jsx
+++ b/src/Table/Table.jsx
@@ -247,6 +247,7 @@ export class Table extends Component {
       fixed,
       paginated,
       rowIDFn,
+      rowClassNameFn,
       onRowClick,
     } = this.props;
     const {cssClass} = Table;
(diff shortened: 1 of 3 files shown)


src/Table/Table.jsx
```diff
@@ -271,7 +272,7 @@ export class Table extends Component {
             </tr>
           ) : displayedData.map(rowData =>
             <tr
-              className={classnames(cssClass.ROW, onRowClick && cssClass.CLICKABLE_ROW)}
+              className={classnames(cssClass.ROW, onRowClick && cssClass.CLICKABLE_ROW, rowClassNameFn && rowClassNameFn(rowData))}
```
kofi-clever: nit: can you break this up to keep the line length 
```jsx
className={classnames(
  cssClass.ROW,
  onRowClick && cssClass.CLICKABLE_ROW,
  rowClassNameFn && rowClassNameFn(rowData),
)}
```

PR #158, opened 2017-04-19
larger page buttons in Table footer

potluck:
**Jira:** https://clever.atlassian.net/browse/CLASSROOMS-1450

**Overview:** The buttons to navigate through pages of the `Table` component are pretty small. Make them a bit bigger.

**Screenshots/GIFs:** 
This gif involves me tabbing through a page with the new table footer (initial), and then one with the old footer with smaller buttons for comparison.
bigger-footer-buttons

### DIFF
--- a/src/Table/Footer.less
+++ b/src/Table/Footer.less
@@ -9,7 +9,6 @@
 }
 
 .Table--footer--cell {
-  font-size: @size_s;
   padding: @size_s;
   text-align: center;
   position: relative;
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.23.9",
+  "version": "0.23.10",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


src/Table/Footer.less
```diff
@@ -9,7 +9,7 @@
 }
 
 .Table--footer--cell {
-  font-size: @size_s;
+  font-size: @size_m;
```
osdiab: Maybe use the type size utilities instead here: https://github.com/Clever/components/blob/master/src/less/type-size.less

CarlNelson:
My suggestion is to instead turn the buttons into link buttons instead of linkPlain.

This makes the click target larger without increasing the type size

potluck:
link-buttons

potluck:
Updated ^

potluck:
I gotta bounce so will figure out CI stuff later, but let me know. Prob still wanna make the link button smaller

potluck:
link-buttons

potluck:
@CarlNelson let me know what you think ^
IMO I think it would still be nicer if the text was a bit larger. The footer at the bottom of the apps list in the schools dashboard just feels *so small* compared to the size of the table. On top of that it's at the bottom of the page, which I think further contributes to feeling like it could be bigger.

src/Table/Footer.less
```diff
@@ -9,7 +10,7 @@
 }
 
 .Table--footer--cell {
-  font-size: @size_s;
+  .text--small;
```
CarlNelson: Do we need this if we're using the buttons?

CarlNelson:
@potluck this looks good to me. The type size for these buttons shouldn't be larger than the text in the table as the table data should be the primary visually. 

Otherwise, ship it.

PR #161, opened 2017-04-24
Add transform-code-string.

samfishman:
Adds a babel plugin to fill in the the `code` property on `<Example>` components so that we don't have to duplicate our code all over the place. If the `code` prop is set manually, then this plugin does nothing so you can still override the default behavior.

This PR doesn't actually remove any of the manually-set `code` props, I did that separately in https://github.com/Clever/components/pull/162.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,7 @@
 const autoprefixer = require("autoprefixer");
 
 module.exports = {
-  entry: ["babel-polyfill", "./docs/docs.jsx"],
+  entry: "./docs/docs.jsx",
   output: {
     path: "docs",
     filename: "bundle.js",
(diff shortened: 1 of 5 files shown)


webpack.config.js
```diff
@@ -36,6 +36,7 @@ module.exports = {
         loader: "babel",
         query: {
           presets: ["react", "es2015", "stage-3"],
+          plugins: [`${__dirname}/transform-code-string`],
```
kofi-clever: the babel loader just concatenates this to the ones in babelrc i'm guessing?

samfishman: Yeah that's right - I removed the `presets` key here so it's less confusing/doesn't duplicate the config.

PR #163, opened 2017-04-24
FEE-72: update secondary border color

CarlNelson:
**Jira:**
[FEE-72](https://clever.atlassian.net/browse/FEE-72)

**Overview:**
- Update secondary button border color so that hover / focus state is much more visible
- Update left border on dropdown button to be full height of button

**Screenshots/GIFs:**
Old secondary button hover
1aeb2071959f2d9c342603f95aa89f3c

New secondary button hover
f4bcc88b146e90f1902a10e8236794e4

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/DropdownButton/DropdownButton.less
+++ b/src/DropdownButton/DropdownButton.less
@@ -58,35 +58,25 @@
  */
 
 .Button--primary.DropdownButton--toggle {
-  .DropdownButton--Caret--container {
-    .border--left--s(shade(@primary_blue, @shadeStepAmount * 2));
-  }
+  .border--left--s(shade(@primary_blue, @shadeStepAmount * 2));
 }
 
 .Button--secondary.DropdownButton--toggle {
-  .DropdownButton--Caret--container {
-    .border--left--s(@primary_blue);
-  }
+  .border--left--s(tint(@primary_blue, @shadeStepAmount * 2));
 }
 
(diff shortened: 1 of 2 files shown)


src/DropdownButton/DropdownButton.less
```diff
@@ -58,35 +58,25 @@
  */
 
 .Button--primary.DropdownButton--toggle {
-  .DropdownButton--Caret--container {
-    .border--left--s(shade(@primary_blue, @shadeStepAmount * 2));
-  }
+  .border--left--s(shade(@primary_blue, @shadeStepAmount * 2));
```
kofi-clever: we can get rid of all these now since it's simpler to just have the top-level toggle button expose it's border instead of repeating the Button styles here on the inner.

i.e. remove the `border-left: 0` on line 24, then remove all of these except the disabled state below on L72, since that still needs to override the default button styles.

should probably also  get rid of the transparent border setting on L80 below as well, since that was only needed because of the awkward short partition showing up next to the darkened background of the primary action button

CarlNelson: Ah excellent. Will do.

bitmoji

CarlNelson: Turns out we need some of them. Or it doesn't show the line break for the primary or destructive.

PR #165, opened 2017-04-24
Inline modal close icon

cozmo:
The modal close icon wasn't working in downstream consumers for reasons that were not entirely clear. It was also a png instead of an svg. This fixes both those issues by inlining an svg (from fontawesome) into the css (this is a pattern other components use successfully). 

At some point we should figure out what is wrong with the compilation step in order for the image not to be included, but that's a battle for another day.


src/Modal/Modal.less
```diff
@@ -52,17 +53,18 @@
...
   cursor: pointer;
-  height: 12px;
-  width: 12px;
-  background: url(./close.png);
-  background-size: 12px 12px;
-  right: 32px;
-  top: 24px;
+  height: @size_m;
+  width: @size_m;
+  background: url(data:image/svg+xml;base64,PHN2ZyB3aWR0aD0iMTc5MiIgaGVpZ2h0PSIxNzkyIiB2aWV3Qm94PSIwIDAgMTc5MiAxNzkyIiB4bWxucz0iaHR0cDovL3d3dy53My5vcmcvMjAwMC9zdmciPjxwYXRoIGQ9Ik0xNDkwIDEzMjJxMCA0MC0yOCA2OGwtMTM2IDEzNnEtMjggMjgtNjggMjh0LTY4LTI4bC0yOTQtMjk0LTI5NCAyOTRxLTI4IDI4LTY4IDI4dC02OC0yOGwtMTM2LTEzNnEtMjgtMjgtMjgtNjh0MjgtNjhsMjk0LTI5NC0yOTQtMjk0cS0yOC0yOC0yOC02OHQyOC02OGwxMzYtMTM2cTI4LTI4IDY4LTI4dDY4IDI4bDI5NCAyOTQgMjk0LTI5NHEyOC0yOCA2OC0yOHQ2OCAyOGwxMzYgMTM2cTI4IDI4IDI4IDY4dC0yOCA2OGwtMjk0IDI5NCAyOTQgMjk0cTI4IDI4IDI4IDY4eiIvPjwvc3ZnPg==);
```
osdiab: Maybe a comment saying that this is the `x` would be nice, as it is not super clear from the css

osdiab:
Make sure you fix the tests

PR #166, opened 2017-04-25
Object assign not lodash merge

kshen0:
**Jira:** n/a

**Overview:** uses `Object.assign` to merge Wizard state rather than `_.merge`, which is a deep merge and doesn't work with arrays

**Screenshots/GIFs:** n/a

**Testing:**
Confirmed correct behavior when merging objects with arrays locally


ahsojar:
LGTM

PR #169, opened 2017-04-27
README: use absolute links for documentation links

peggyl:
Currently they're treated as relative links to https://github.com/Clever/components/blob/master/clever.github.io/components/#/intro and https://github.com/Clever/components/blob/master/clever.github.io/components/#/getting-started, which do not exist.


osdiab:
LGTM!

PR #170, opened 2017-04-27
DateInput

cozmo:
This PR does several things to simplify and empower more date related functions

- [x] Change `DatePicker` to be a simple card that flows with content (instead of being overlaid over the content). This allows us to use it in our own popovers or as part of a page, etc. This also means it's a very very thin wrapper around `react-datepicker`. 
- [x] Create a new `DateInput` input that follows our normal input styles, and pops open an datepicker to pick dates. 

This allows more flexibility but also more standardization in how we handle different cases of date picking.  It also allows use to leverage the accessibility of react-datepicker, which is actually pretty good (see more in screenshot below). 

Screenshot gif (for some reason fails to embed) -
 https://cl.ly/0e1q2v311K2U/Screen%20Recording%202017-04-27%20at%2012.34%20AM.gif


src/DatePicker/DatePicker.jsx
```diff
@@ -1,78 +1,26 @@
...
-  value: PropTypes.oneOfType([
-    PropTypes.string,
-    PropTypes.instanceOf(Date),
-  ]),
-  onClose: PropTypes.func.isRequired,
-  onChange: PropTypes.func,
-  footer: PropTypes.node,
-  datePickerProps: PropTypes.object,
-};
+const stringOrDate = PropTypes.oneOfType([
```
kofi-clever: should this include PropTypes.string as well, if it supports it?

src/DateInput/DateInput.less
```diff
@@ -0,0 +1,83 @@
...
+      .padding--top--xs;
+      line-height: @size_2xl;
+      min-height: @size_2xl;
+      color: @neutral_dark_gray;
+    }
+  }
+
+  &.DateInput--hasError {
+    box-shadow: inset @size_2xs 0 @alert_red;
+    transition: box-shadow .25s ease-out;
```
kofi-clever: nit: can put the transition on the base `.DateInput` style and it will apply to both `--hasError` and `--inFocus`

src/DateInput/DateInput.less
```diff
@@ -0,0 +1,83 @@
+@import (reference) "../less/colors";
+@import (reference) "../less/border";
+@import (reference) "../less/sizing";
+@import (reference) "../less/type-size";
+@import (reference) "../less/type-utilities";
+@import (reference) "../less/spacing";
```
kofi-clever: opt: you can all get these by importing the single `../less/utilities`

brettcvz:
Is there a way to "Clear" the datepicker?

cozmo:
@brettcvz it's a textinput so you can use your keyboard.

cozmo:
You can try this in the react datepicker demos

docs/components/DateInputView.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+  render() {
+    return (
+      <View title="DateInput" sourcePath="src/DateInput/DateInput.jsx">
+        <p>
+          DateInput is an input that allows the user to select dates from a DatePicker.
+        </p>
+        <Example>
+          <DateInput
+            label="label"
```
CarlNelson: The label is not getting associated with the `for={name}` of the `input` field for some reason.

Might want to double check that it works correctly.

CarlNelson:
LGTM

src/DateInput/DateInput.less
cozmo: For now keeping this the same as the `.TextInput` for my own sanity.

PR #172, opened 2017-05-02
first draft at adding a panel element

Christopher-Bradshaw:
**Jira:**

**Overview:**
Add a panel element with various types. This is my first component so probably a bit rough. Will write tests once I get a preliminary once over. Open questions for reviewer:

1) `Panel` might be too generic a name. Any suggestions? (Changed to AlertBox)
2) Are we OK that this requires the app to have access to font awesome?
3) This doesn't show up in left nav of the docs. I can't work out where I need to add it. Any suggestions?

**Screenshots/GIFs:**
panel
image

**Testing:**
- [x] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


src/Panel/Panel.less
```diff
@@ -0,0 +1,45 @@
+@import (reference) "../less/index";
+
+.Panel--container {
+  border-style: solid;
+  border-width: @size_3xs @size_3xs @size_3xs @size_s;
+  border-radius: @size_2xs;
```
CarlNelson: You can use the .borderRadius mixin here.

src/Panel/Panel.less
```diff
@@ -0,0 +1,45 @@
+@import (reference) "../less/index";
+
+.Panel--container {
+  border-style: solid;
+  border-width: @size_3xs @size_3xs @size_3xs @size_s;
```
CarlNelson: You can use the .border mixins here.

src/Panel/Panel.less
```diff
@@ -0,0 +1,45 @@
...
+  border-radius: @size_2xs;
+  padding: @size_xs;
```
CarlNelson: Can use spacing mixins here, and all the places you directly reference the sizing variables

e.g.:  .padding--xs

src/Panel/Panel.jsx
```diff
@@ -0,0 +1,48 @@
...
+/*
+ * Panel is various types of panel.
+*/
+export default function Panel({children, type, title}) {
+  const {Icons, cssClass} = Panel;
+
+  return (
+    <div className={classnames(`Panel--${type}`, cssClass.CONTAINER)}>
+      <div className={cssClass.HEADER}>
+        <i className={classnames("fa", Icons[type])} />
```
CarlNelson: Would add an 'alertIcon--[type]' here to style directly rather than have them nested in the LESS.

src/Panel/Panel.less
```diff
@@ -0,0 +1,45 @@
+@import (reference) "../less/index";
+
+.Panel--container {
```
CarlNelson: nit: should be PascalCase for the child object.

See http://clever.github.io/components/#/less/less-style-guide for explanation

CarlNelson:
These are generally called Alerts or Alert Boxes. Would rename the component to that.

Should also have an `canDismiss` prop that is a boolean to close/dismiss the alert box.

src/Panel/Panel.less
Christopher-Bradshaw: For the left border, none of the mixins are large enough. Can I use the `@size_s` there or is there a better way to do it.

src/Panel/Panel.less
Christopher-Bradshaw: Can you elaborate on what you mean here? I see
`.TabBar--Tab--alert` in that page and thought container would be like alert?

src/Panel/Panel.jsx
Christopher-Bradshaw: Does nesting not make sense here? The nested syling only makes sense/is applied when inside the relevant box? Happy to change if we want to avoid all nesting but I feel it is easier to understand like this.

CarlNelson:
Totally can

`alert` in the example is a modifier. While `container` is an element.

For AlertBox `info|error|warning|success` are modifiers.

CarlNelson:
Generally avoiding nesting of any kind makes it overall more readable.

docs/components/AlertBoxView.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  render() {
+    const {AlertBoxOptions} = AlertBoxView;
+    const {type} = this.state;
+
+    return (
+      <View title="AlertBox" sourcePath="src/AlertBox/AlertBox.jsx">
+        <p>
+          A set of alert boxes. N.B. this requires the app you use it in to have
```
osdiab: It would be better if it didn't, since it requires fontawesome assets which may not be available in the client app; so you can get the SVG versions of the icons, and copy them into components like so:

```
function MyIcon() {
  return <svg>...</svg>; // copy the svg contents here
}
```

src/Panel/Panel.jsx
osdiab: browsers build indexes of classnames to elements, and specifying elements via nesting can often be less efficient and brittle to structural changes in the HTML than if you just gave the element a CSS class itself. He's saying to keep the current nesting of the HTML, but to refer to his element in CSS by a classname.

src/Panel/Panel.less
osdiab: I think he means that if `container` is a component, this should be `.Panel--Container` to denote that `Container` is a component; but if it's not I don't think you need to capitalize it, correct me if I'm wrong @CarlNelson

docs/components/AlertBoxView.jsx
```diff
@@ -0,0 +1,69 @@
...
+          access to Font Awesome.
+        </p>
+        <CodeSample>
+          {`
+            import {AlertBox} from "clever-components";
```
osdiab: Code sample could use more love

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,70 @@
...
+}
+
+AlertBox.Type = {
+  WARNING: "warning",
+  SUCCESS: "success",
+  ERROR: "error",
+  INFO: "info",
+};
+
+AlertBox.Icons = {
```
osdiab: yep, I'd just include the SVG assets as components as described above.

src/AlertBox/AlertBox.less
```diff
@@ -0,0 +1,70 @@
...
+  .padding--xs;
+}
+
+.AlertBox--Title {
+  display: inline;
+  .margin--left--2xs;
+}
+
+.AlertBox--warning {
+  .tint(background-color, @alert_orange, 6);
```
osdiab: maybe make `6` a variable

docs/components/AlertBoxView.jsx
Christopher-Bradshaw: Can do. Also https://en.wikipedia.org/wiki/Nota_bene

CarlNelson:
Let's do a quick design pass in person before we land this :)

Christopher-Bradshaw:
LGTM in person by carl

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,66 @@
+import _ from "lodash";
+import classnames from "classnames";
+import React, {PureComponent, PropTypes} from "react";
+
+import "./AlertBox.less";
+import "./AlertBoxIcons.less";
+
+const closeIcon = <svg viewBox="0 0 1792 1792" xmlns="http://www.w3.org/2000/svg"><path d="M1490 1322q0 40-28 68l-136 136q-28 28-68 28t-68-28l-294-294-294 294q-28 28-68 28t-68-28l-136-136q-28-28-28-68t28-68l294-294-294-294q-28-28-28-68t28-68l136-136q28-28 68-28t68 28l294 294 294-294q28-28 68-28t68 28l136 136q28 28 28 68t-28 68l-294 294 294 294q28 28 28 68z" /></svg>;
```
osdiab: Can you make this a function that returns this and then use it as a component? Like:

```
function CloseIcon() {
 return <svg>...</svg>
};

//...
render() {
  return <CloseIcon />...
}
```

In the current case, all `AlertBox`'es on the screen will use the same instance of the component if you have multiple of them onscreen, I'm not so sure how React would deal with that especially if the Javascript enacts any modification to the SVG styling / elements via js. more robust to just create a new instance via the above code.

src/AlertBox/AlertBoxIcons.less
```diff
@@ -0,0 +1,24 @@
...
...Izc7bbTXSwhmmp6AkZBlFcbKflVqK5J7WvpTa6sgiDJsZRvOB7xEyugR5tJTUmysYw6RBpLn2mVzw08apXuJs3mXM3r8nnqxZWLcvSfK1+8cL6K8nGCyVDeqd3n4pc2xGciYKfr25UPGVmLadyqBi90+i/kV/0Qo2bszuLbnjgaPTyz4ZVdyNYDVf9XtxA18lXTrh+uOXagVfRlUDn3OXKmnvK39Qeyzmi/y+onpX6AAAAeJxjYGRgYADi9RGugvH8Nl8ZuJlfAEUYrhqICsLo/1//ZzF/YQ4CcjkYmECiACUsCv4AAHicY2BkYGAO+p8FJF/8//r/C/MXBqAICmAFALW2B48AA+gAAAPo//UDWf/9A+gAAAOgAAAAAAAAAHQA7AFAAeUAAQAAAAUAVgADAAAAAAACABIAIgBzAAAAXwtwAAAAAHicdZDdasIwGIbfzJ9tCtvYYKfL0VDG6g8MQRAEh55sJzI8HbXWtlIbSaPgbewedjG7iV3LXts4hrKWNM/35MuXrwFwjW8I5M8TR84CZ4xyPsEpepYL9M+Wi+QXyyVU8Wa5TP9uuYIHBJaruMEHK4jiOaMFPi0LXIlLyye4EHeWC/SPlovknuUSbsWr5TK9Z7mCiUgtV3EvvgZqtdVREBpZG9Rlu9nqyOlWKqoocWPprk2odCr7cq4S48excjy13PPYD9axq/fhfp74Oo1UIltOc69GfuJr1/izXfV0E7SNmcu5Vks5tBlypdXC94wTGrPqNhp/z8MACitsoRHxqkIYSNRo65zbaKKFDmnKDMnMPCtCAhcxjYs1d4TZSsq4zzFnlND6zIjJDjx+l0d+TAq4P2YVfbR6GE9IuzOizEv25bC7w6wRKcky3czOfntPseFpbVrDXbsuddaVxPCghuR97NYWNB69k92Koe2iwfef//sB5m6EUQB4nGNgYoAALgbsgJWRiZGZkYWRlZGNgb08sSgvMy+dJTMvLZ81tagov4i9uDQ5ObW4mIEBAKA1CgcAAAB4nGPw3sFwIihiIyNjX+QGxp0cDBwMyQUbGVidNjEwMmiBGJu5mBg5ICw+BjCLzWkX0wGgNCeQze60i8EBwmZmcNmowtgRGLHBoSNiI3OKy0Y1EG8XRwMDI4tDR3JIBEhJJBBs5mFi5NHawfi/dQNL70YmBhcADHYj9AAA') format('woff'),
```
osdiab: @CarlNelson I wonder if it's better to use this inline icon font or to instead use [the SVG icons](https://github.com/encharm/Font-Awesome-SVG-PNG/tree/master/black/svg) directly in the code—I think the latter is clearer what it does and makes the SVG's more easily swappable, as this is arcane and more difficult to customize later on. Also may not be a pattern we can use for icons in the future if we ever need to transform the icons themselves, as SVG's rendered into the DOM give full use of CSS and JavaScript selectors and transformations on the icons themselves.

CarlNelson: Ah yeah. I would inline SVGs instead of using the icon font. Good catch.

Christopher-Bradshaw: cc @kofi-clever ^ is how we want to do it rather than using the fontello stuff

kofi-clever: yeah, i'm probably in the minority here on this issue, but my thoughts are:
- svg fonts are better suited for icons that don't necessarily have a pre-determined size or color and can be adjusted easily to fit various contexts (grow with text, colored for different state like enabled/disabled, etc)
- plain svg icons are great for icons/images with pre-determined size/color settings (like our `Icon` component) but can be more restrictive WRT customization/adapting to surrounding context, since they have to be styled differently

in this specific case, no strong feelings either way, but figured this probably won't be the last time we need these icons in Dewey (e.g. toast notifications) and would be nice if they were flexible enough to be used in different situations

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,70 @@
...
+    const {children, className, type, title, isClosable} = this.props;
+    const {isOpen} = this.state;
+    const {Icons, cssClass} = AlertBox;
+    if (!isOpen) {
+      return null;
+    }
+    return (
+      <div className={classnames(`AlertBox--${type}`, cssClass.CONTAINER, className)}>
+        <div className={cssClass.HEADER}>
+          <span>{Icons[type]()}</span>
```
osdiab: Only one change I'd make here—just to make clear that this is actually creating a React element, can you in the code above say:

```
const Icon = Icons[type];
```

and then invoke it here as:

```
<span><Icon /></span>
```

That way it's obvious that `Icon` is a React component.

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,70 @@
...
+            <button className={cssClass.CLOSE} onClick={() => this.closeBox()}><CloseIcon /></button>
+          }
+        </div>
+        {children}
+      </div>
+    );
+  }
+}
+
+AlertBox.Type = {
```
osdiab: No need for `Type` to be uppercased, maybe call this `AlertBox.types` or capitalize it entirely, and call it `TYPES`. Additionally it doesn't need to be a member of `AlertBox`, actually. So how about just

```
const TYPES = {
  //...
};
```

If you want them to be accessible outside of `AlertBox.jsx`, then you can `export` this as a named export as well.

src/AlertBox/AlertBox.jsx
osdiab: And here, instead of capitalizing it, call it `const ICONS = ...`

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,70 @@
...
+};
+
+AlertBox.Icons = {
+  warning: WarningIcon,
+  success: SuccessIcon,
+  error: ErrorIcon,
+  info: InfoIcon,
+};
+
+AlertBox.cssClass = {
```
osdiab: to keep with the plural/non-member naming scheme above, `const CLASSES` or `CLASS_NAMES`; I wouldn't append `css` since classnames are as much a part of the HTML spec as they are in CSS.

src/AlertBox/AlertBox.jsx
```diff
@@ -0,0 +1,70 @@
+import _ from "lodash";
+import classnames from "classnames";
+import React, {PureComponent, PropTypes} from "react";
+
+import "./AlertBox.less";
+import {CloseIcon, WarningIcon, SuccessIcon, ErrorIcon, InfoIcon} from "./Icons";
```
osdiab: since `Icons` isn't actually a React component, but just a file that exports other react components, I'd lowercase it and call the file `icons.jsx`.

PR #174, opened 2017-05-04
add style to modalbutton's button to match height of container div

osdiab:
**Jira:**
none

**Overview:**
Add 100% height styling to modal button's nested button, so that it matches height of parent. Also adds `nowrap` to `Button` after talking with @mischelle-m , confirming that we shouldn't be having multi-line buttons anyway.

**Screenshots/GIFs:**

NOTE: the real version of the example page used below doesn't actually use the primary type links in an infopanel footer; it's just to illustrate the true size of the buttons.

### Buggy behavior
In a flex parent, the modalbutton is smaller than the rest of them. its wrapping div matches the correct height, but not the child button.

screen shot 2017-05-04 at 10 31 27

### With the fix
screen shot 2017-05-04 at 10 31 41

### With expanded parent height
screen shot 2017-05-04 at 10 31 55

### In its natural documentation habitat
screen shot 2017-05-04 at 10 31 15

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Button/Button.less
+++ b/src/Button/Button.less
@@ -34,6 +34,7 @@ a, button {
     font-weight: 600;
     text-decoration: none;
     transition: all 0.25s ease-out;
+    white-space: nowrap;
 
     &[disabled],
     &[disabled]:hover,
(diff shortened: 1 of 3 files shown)


osdiab:
Actually i'm thinking this is not a good change to make, as it makes buttons not their true size in the flex context. Closing.

PR #175, opened 2017-05-05
Vendor in a version of datepicker to fix a bug

cozmo:
React datepicker has a bug - https://github.com/Hacker0x01/react-datepicker/issues/743

There's a PR to fix said bug here - https://github.com/Hacker0x01/react-datepicker/pull/825

Sadly we can't just use that github repo as the source in our `package.json` since there's a build step run before publishing and that repo isn't published. 

Instead I just build the code locally and committed it into components. Hopefully we can revert this soon once the PR gets merged and the fix gets released. Until then this approach seems to work fine.


cozmo:
I tried to roll back to `0.40.0` but it was different enough that it wasn't a drop in replacement (at the very least had to re-skin it as I think some class names changed). This seemed easier.

PR #177, opened 2017-05-05
Select: add creatable option

jonahkagan:
**Jira:** none

**Overview:** Adds a `creatable` prop to allow users to create new options. Implemented using the ReactSelect Creatable component as recommended in the docs: https://github.com/JedWatson/react-select#user-created-tags. Also adds a `creatableProps` prop to allow configuration of new option creation. I chose this because I didn't want to have an opinion about what kinds of customization is worthwhile - I figure ReactSelect got the way it did for a reason so we should just use their behavior.

**Screenshots/GIFs:**
screenshot 2017-05-05 15 40 59

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.24.5",
+  "version": "0.25.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/Select/Select.jsx
+++ b/src/Select/Select.jsx
@@ -37,6 +37,8 @@ export function Select({
   placeholder = "",
   readOnly,
   searchable,
+  creatable,
+  creatableProps,
   value,
   className,
 }) {
@@ -52,12 +54,15 @@ export function Select({
     reactSelectClasses += ` ${cssClass.READ_ONLY}`;
   }
 
+  const SelectComponent = creatable ? ReactSelect.Creatable : ReactSelect;
+  const extraProps = creatable && creatableProps;
+
   // The label container must be returned after the ReactSelect otherwise it does not get displayed
   // in the browser.
   return (
     <div className={classnames(cssClass.CONTAINER, className)}>
       <div id={id}>
-        <ReactSelect
+        <SelectComponent
           className={reactSelectClasses}
           clearable={clearable}
           disabled={disabled || readOnly}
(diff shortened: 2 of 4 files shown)


src/Select/Select.jsx
```diff
@@ -105,6 +111,8 @@ Select.propTypes = {
   placeholder: React.PropTypes.string,
   readOnly: React.PropTypes.bool,
   searchable: React.PropTypes.bool,
+  creatable: React.PropTypes.bool,
+  creatableProps: React.PropTypes.object,
```
kofi-clever: this makes it a little difficult to discover what props are available when using a creatable (and also leaks the internals of this component - wrt the fact that it's currently dependent on react-select)

i'd recommend listing out explicit creatable props that we want to support in the top-level `Select.propTypes` and passing them through explicitly when we use the `Creatable` component. wdyt?

jonahkagan: It's hard for me to know what creatable props we want to support. The fact that react-select decided to implement this behavior with a separate component makes me think that they probably did that in order to avoid having a ton of props that only applied in the "creatable" case. I was basically trying to follow their design pattern without actually making a new component.

kofi-clever: yeah, we shouldn't need a separate component. hoping we can just copy the list of proptypes over from the `Creatable` and list them here instead of requiring users to look them up in react-select docs. interface for a normal select and a create-enabled one would be the same; the user can just include the additional props.
```jsx
<Select
  options={[...]}
  creatable
  newOptionCreator={() => {...}}
  isValidNewOption={() => {...}}
/>
```

jonahkagan: Fixed in aa5f30b

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.24.5",
+  "version": "0.25.0",
```
kofi-clever: ah, missed this earlier - this can be `0.24.6` since it's not a breaking change

jonahkagan: oops. i figured it's a minor by normal semver since it expands the interface

PR #178, opened 2017-05-09
Inline icons

osdiab:
**Jira:**
https://clever.atlassian.net/browse/FEE-64

**Overview:**
Use the `react-svg-loader` webpack middleware to convert required svgs into react components.

Chose [`react-svg-loader`](https://github.com/boopathi/react-svg-loader) over [`svg-react-loader`](https://github.com/jhamlet/svg-react-loader) since the documentation is better and it looks more active.

Doesn't work as smoothly using `require`'s since the loader provides the SVGs as the default export, which `require` doesn't default to; so I just used the ES6 `import` keyword.

**Screenshots/GIFs:**
Icons look the same as before.

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


osdiab:
Wait, so I thought the dist folder was the distributable we make available, and that it was produced by webpack; is that not the case?

osdiab:
This whole "webpack only used for the docs" thing seems silly, I'll make the webpack config always be used tomorrow so we can be done with this confusion.

kofi-clever:
FWIW, webpacking everything before consumers import could also be tricky. if we do it in the simplest way, we'd lose the ability to eventually enable tree-shaking in consumers, since they'd just be pulling in one blob of js.

might require getting really complicated and having a separate webpack entry point for each component and configuring them to not pull in other components - that has the risk of increasing the overhead/complexity of adding new components or adding redundant bloat to clever-components imports.

this must be a solved problem. we can see what other open source components libraries are doing and see what works best for our use case

osdiab:
I'm pretty sure this is a solved problem, in the form of Rollup:

https://medium.com/webpack/webpack-and-rollup-the-same-but-different-a41ad427058c

I'll do some investigation on what the hubbub around that is.

PR #179, opened 2017-05-09
Wizard: support opening wizard at a specific step

peggyl:
**Overview:**

The wizard always opens at the first step by default, but sometimes we want to resume the wizard at a specific later step, e.g. the furthest step the user has reached so far.

**Screenshots/GIFs:**

https://cl.ly/321i0X294610

**Testing:**
- [ ] Unit tests - I think the demo (see above) is more useful, since this essentially only calls an existing method when a prop value is set.
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10 - no, but I don't think this is necessary since it just conditionally updates state in an existing component to an already-possible value.

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/Wizard/Wizard.jsx
+++ b/src/Wizard/Wizard.jsx
@@ -18,10 +18,21 @@ export class Wizard extends React.Component {
     this.state = _.assign(INITIAL_STATE, {
       data: props.initialWizardData || {},
     });
+
     this.reset = this.reset.bind(this);
     this.prevStepHandler = this.prevStepHandler.bind(this);
     this.nextStepHandler = this.nextStepHandler.bind(this);
     this.calculatePercentComplete = this.calculatePercentComplete.bind(this);
+
+    // Launch the wizard at a specific step (e.g. to resume at furthest step)
+    if (this.props.initialStep) {
+      const idx = this.props.initialStep;
+      Object.assign(this.state, {
+        currentStep: idx,
+        stepsVisited: _.range(idx + 1), // include the step itself
+        percentComplete: this.calculatePercentComplete(),
+      });
+    }
   }
 
   reset() {
@@ -214,6 +225,7 @@ Wizard.propTypes = {
     props: PropTypes.object,
     onStepComplete: PropTypes.func,
   })).isRequired,
+  initialStep: PropTypes.number,
   nextButtonValue: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   prevButtonValue: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   seekable: PropTypes.bool,
--- a/docs/components/WizardView.jsx
+++ b/docs/components/WizardView.jsx
@@ -72,6 +72,13 @@ export default class WizardView extends PureComponent {
               optional: true,
             },
             {
+              name: "initialStep",
+              type: "Number",
+              description: "The index of the step to open the wizard at",
+              defaultValue: 0,
+              optional: true,
+            },
+            {
               name: "initialWizardData",
               type: "Object",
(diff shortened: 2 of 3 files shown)


src/Wizard/Wizard.jsx
```diff
@@ -18,12 +18,23 @@ export class Wizard extends React.Component {
     this.state = _.assign(INITIAL_STATE, {
       data: props.initialWizardData || {},
     });
+
+    if (this.props.initialStep) {
+      this.setState({currentStep: this.props.initialStep});
```
samfishman: You should't call `this.setState` in the constructor - just mutate `this.state` directly

samfishman:
Why doesn't it work to just set `this.state.currentStep` in the constructor?

peggyl:
`jumpToStep` sets some other state in addition to `currentStep`, at which point it made more sense to reuse the existing function than implement it all in the constructor.

samfishman:
:thinking: I'm a little worried that there could be some jumpiness since the initial render will be on the first step, and it seems like it's pretty easily avoidable by doing things in the constructor - can we factor out a function that returns the state update (e.g. returns the object `{currentStep: idx, stepsVisited: _.union(this.state.stepsVisited, [idx]), percentComplete: this.calculatePercentComplete()}`, but use `Object.assign` in the constructor and `this.setState` in `jumpToStep`?

peggyl:
Actually, from a quick glance at the implementation, we may not need to initialize `stepsVisited` and `percentComplete` in the constructor for this. I'll try it out…

src/Wizard/Wizard.jsx
```diff
@@ -18,22 +18,31 @@ export class Wizard extends React.Component {
...
       data: props.initialWizardData || {},
     });
+
     this.reset = this.reset.bind(this);
     this.prevStepHandler = this.prevStepHandler.bind(this);
     this.nextStepHandler = this.nextStepHandler.bind(this);
     this.calculatePercentComplete = this.calculatePercentComplete.bind(this);
+
+    if (this.props.initialStep) {
+      this.state = Object.assign({}, this.state, this._getStateAtStep(this.props.initialStep));
```
samfishman: This line can just be written `Object.assign(this.state, this._getStateAtStep(this.props.initialStep))` – `Object.assign` mutates the first parameter.

src/Wizard/Wizard.jsx
```diff
@@ -18,22 +18,31 @@ export class Wizard extends React.Component {
...
     this.setState(_.assign({}, INITIAL_STATE, {data: {}}));
   }
 
-  jumpToStep(idx) {
-    this.setState({
+  _getStateAtStep(idx) {
+    return {
       currentStep: idx,
-      stepsVisited: _.union(this.state.stepsVisited, [idx]),
+      stepsVisited: _.range(idx + 1), // include the step itself
```
samfishman: I know this goes counter to the discussion we just had in person, but we probably don't want to break the pre-existing behavior of `jumpToStep`, which is used when `seekable` is true. So maybe we just need to keep `jumpToStep` and the constructor for `initialStep` totally separate? Up to you so long as there isn't a breaking change.

src/Wizard/Wizard.jsx
```diff
@@ -31,7 +42,7 @@ export class Wizard extends React.Component {
   jumpToStep(idx) {
     this.setState({
       currentStep: idx,
-      stepsVisited: _.union(this.state.stepsVisited, [idx]),
+      stepsVisisted: _.union(this.state.stepsVisited, [idx]),
```
samfishman: typo 😬

peggyl: linter, you had one job… 😡

PR #183, opened 2017-05-11
Sticky wizard sidebar

samfishman:
**Overview:**
Adds an option to make the sidebar sticky on a wizard. Uses `react-sticky` since `position: fixed` messes up wildly when you can scroll horizontally.

**Screenshots/GIFs:**
wizard-sticky

**Testing:**
- [n/a] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Wizard/Wizard.less
+++ b/src/Wizard/Wizard.less
@@ -20,6 +20,10 @@
   }
 }
 
+.Wizard--stepsDisplay--stepTitle {
+  -ms-flex: 0 1 auto;
+}
+
 @content_width: 37.5rem;
 @help_min_width: 18.75rem;
 .Wizard--contentGroup {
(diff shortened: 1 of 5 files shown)


samfishman:
Also, when testing in IE10, noticed that the step labels are broken. According to StackOverflow, the fix is to add `-ms-flex: 0 1 auto` to the affected node.

src/Wizard/Wizard.less
```diff
@@ -20,6 +20,10 @@
   }
 }
 
+.Wizard--stepsDisplay--stepTitle {
+  -ms-flex: 0 1 auto;
+}
+
```
potluck: linter caught that this `.Wizard--stepsDisplay--stepTitle` selector is already there on line 115, so just add it there?

samfishman: ah nice

PR #188, opened 2017-05-19
Table: Prevent lazy table corruption/race condition.

samfishman:
Without this, it's possible that if there are subsequent calls to `.lazyReset` and the first fetch finishes after the second fetch, the first fetch will end up displaying in the table even though it should be the second. This change tracks calls to `.lazyReset` so that this race condition can't occur.

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Table/Table.jsx
+++ b/src/Table/Table.jsx
@@ -48,6 +48,10 @@ export class Table extends Component {
       pageLoading: false,
       allLoaded: false,
     };
+
+    // the epoch is incremented on calls to this.lazyReset, so we know to throw
+    // out in-progress fetches if the table gets reset (see this._fetchLazy).
+    this._epoch = 0;
   }
 
   componentWillMount() {
@@ -91,6 +95,8 @@ export class Table extends Component {
       return;
     }
 
+    const startEpoch = this._epoch;
+
     this.setState({pageLoading: true});
 
     const newPages = lazyPages.slice(0);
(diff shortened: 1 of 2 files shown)


src/Table/Table.jsx
```diff
@@ -107,6 +113,12 @@ export class Table extends Component {
       newPages.push(pageData);
     }
 
+    if (this._epoch !== startEpoch) {
+      // the table was reset while we were fetching, so we don't want to
+      // actually insert our data
+      return;
+    }
```
kshen0: @jefff and I had discussed the best way to handle outdated request promises. One more direct way to do this is to track the promise itself rather than a counter

```
async _fetchLazy(pageIdx) {
  ...
  const getDataPromise = getData(query);
  this.currentPromise = getDataPromise;
  const pageData = await getDataPromise;
  if (this.currentPromise !== getDataPromise) {
    return;
  }
  ...
}

lazyReset() {
  this.currentPromise = null;
  ...
}
```

samfishman: It doesn't really work here since there are multiple requests being made in a loop.

samfishman: Plus, I think this way is a lot more understandable anyway.

kshen0: Ah true we'd have to track all in-flight promises then, which is harder to reason about.

PR #193, opened 2017-06-02
Select: Support lazy-loading options.

samfishman:
**Overview:**
Lets table fetch options from a backend server. Useful if there are lots of options.

**Screenshots/GIFs:**
select-async

**Testing:** N/A? Just passing options to a 3rd party lib.
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/SelectView.jsx
+++ b/docs/components/SelectView.jsx
@@ -229,8 +229,14 @@ export default class SelectView extends Component {
             },
             {
               name: "loadOptions",
-              type: "(input: string) => Promise<{options: {label: string; value: string}[]}>",
-              description: "Function to load options for a given search input. Required if `lazy`.",
+              type: "(input: string) => Promise<{options: {label: string; value: string}[], complete: boolean}>",
+              description: "Function to load options for a given search input. " +
+                "Required if `lazy`. When the component is mounted, this function " +
+                "will be called with the empty string as input to seed the default " +
+                "set of options. Results are cached. If you set `complete` to `true` " +
+                "in the return value to indicate that the set of options is the complete " +
+                "set for that query, then more specific queries will not incur new " +
+                "(unnecessary) calls to `loadOptions`.
               optional: true,
             },
             {
--- a/docs/components/SelectView.jsx
+++ b/docs/components/SelectView.jsx
@@ -236,7 +236,7 @@ export default class SelectView extends Component {
                 "set of options. Results are cached. If you set `complete` to `true` " +
                 "in the return value to indicate that the set of options is the complete " +
                 "set for that query, then more specific queries will not incur new " +
-                "(unnecessary) calls to `loadOptions`.
+                "(unnecessary) calls to `loadOptions`.",
               optional: true,
             },
             {
--- a/docs/components/SelectView.jsx
+++ b/docs/components/SelectView.jsx
@@ -20,6 +20,7 @@ export default class SelectView extends Component {
       readOnly: false,
       searchable: false,
       creatable: false,
+      lazy: false,
       selectValue: null,
     };
   }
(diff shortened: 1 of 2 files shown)


src/Select/Select.jsx
```diff
@@ -44,6 +44,22 @@ export function Select({
 }) {
   const {cssClass} = Select;
 
+  if (!lazy) {
```
osdiab: Alternatively, you can rewrite propTypes to behave a little differently to achieve the same thing without explicit prop checking code, like so:

```
Select.propTypes = {
  // ...
  options: React.PropTypes.oneOfType([
    React.PropTypes.arrayOf(selectValuePropType),
    React.PropTypes.shape({
      // custom validator to ensure that lazy must be set to true for `loadOptions` to be required
      lazy: (v) => v === true ? null : new Error("`lazy` must be `true` to use lazy loading"),
      loadOptions: React.PropTypes.func.isRequired,
    }),
  ]).isRequired,
  // ...
}
```

This way existing usage of options directly as an array is supported, but if it's an object of the lazy format it would use that form.

docs/components/SelectView.jsx
```diff
@@ -199,8 +216,21 @@ export default class SelectView extends Component {
...
+              name: "lazy",
+              type: "boolean",
+              description: "Set to true to fetch options asynchronously using the `loadOptions` function.",
+              defaultValue: "false",
+              optional: true,
+            },
+            {
+              name: "loadOptions",
+              type: "(input: string) => Promise<{options: {label: string; value: string}[]}>",
+              description: "Function to load options for a given search input. Required if `lazy`.",
```
osdiab: There's some nuance to the way react-select does this that we should probably expose, like how unless we supply `autoload=false` to the React Select component it will default the data to whatever `loadOptions('')` returns; and that options for previously searched inputs are cached (unless explicitly set not to in the React Select component). At a minimum, this behavior ought to be documented, and at a maximum we should consider if this sort of stuff are things we'd like to make configurable, perhaps via a passthrough `reactSelectOptions` property.

src/Select/Select.jsx
```diff
@@ -44,6 +44,22 @@ export function Select({
...
+    if (!options) {
+      console.error("Select: prop \"options\" must be set if not \"lazy\"");
```
osdiab: If you do stay with this, pretty sure this should be `console.warn`

src/Select/Select.jsx
samfishman: Hmm I find that schema pretty confusing (array or object with lazy attribute set to true and a loadOptions function). I'll change to console.warn, but I'm going to keep it as is.

docs/components/SelectView.jsx
samfishman: Personally, I think that behavior is really reasonable. Until we see an explicit need to configure it, I think we should leave it as-is (but I'll add the appropriate documentation). And even then, I think we should limit the configuration surface rather than just exposing a blanket `reactSelectOptions` prop.

src/Select/Select.jsx
osdiab: An alternative, in that case, is to do it at the type checking phase with custom validators:

```
import { all } from "prop-types-extra";
// ...
// validator to ensure props are not truthy at the same time
function numTruthy(num, ...propNames) {
  return (props) => 
    propNames.filter(name => !!props[name]) !== num ?
      new Error(`Exactly ${num} of props ${JSON.stringify(propNames)} must be specified") : null
}

// custom validator to ensure props are all truthy, or are all falsey
function allOrNoneTruthy(...propNames) {
  return (props) => {
      [0, propNames.length].indexOf(propNames.filter(name => !!props[name])) !== -1 ?
        null : new Error(`Props ${JSON.stringify(propNames)} must be specified together") :
}

Select.propTypes = {
  // ...
  options: all(
    React.PropTypes.arrayOf(selectValuePropType),
    numTruthy(1, "lazy", "options"),
  ),
  lazy: all(
    React.PropTypes.bool,
    numTruthy(1, "lazy", "options"),
    allOrNoneTruthy("lazy", "loadOptions"),
  ),
  loadOptions: all(
    React.PropTypes.func,
    allOrNoneTruthy("lazy", "loadOptions"),
  )
  // ...
}
```

But if this is something you don't like, then the current approach is OK.

PR #195, opened 2017-06-09
Adds a RichText component

cozmo:
Adds a component that renders text while preserving newlines and linkifying links and emails

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/RichTextView.jsx
+++ b/docs/components/RichTextView.jsx
@@ -1,0 +1,52 @@
+import React from "react";
+
+import Example from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+import {RichText} from "src";
+
+export default function DateInputView() {
+  return (
+    <View title="RichText" sourcePath="src/RichText/RichText.jsx">
+      <p>
+        RichText is a view that improves rendering of raw text - Specifically by linkifying urls and not discarding linebreaks.
+      </p>
+      <i>Note - Rendering large blobs of text with urls isn't the best from an accessibility perspective, when possible links should have readable titles, instead of just a url.</i>
+      <h2>Without using RichText</h2>
+      <Example>
+        <p>{`
+          Do not go gentle into that good night,
+          Old age should burn and rave at close of day;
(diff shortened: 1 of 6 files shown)


docs/components/RichTextView.jsx
```diff
@@ -0,0 +1,58 @@
...
+            type: "String",
+            description: "The text to render",
+            optional: false,
+          },
+          {
+            name: "newTab",
+            type: "Boolean",
+            description: "Should links open in a new tab?",
+            optional: true,
+            defaultValue: "false",
```
potluck: what do you think about defaulting to true instead of false for this one? I feel like links opening in a new tab is the more expected / user-friendly behaviort

src/RichText/RichText.jsx
```diff
@@ -0,0 +1,17 @@
+import React from "react";
+import Linkify from "react-linkify";
+
+export default function RichText(props) {
+  let properties = {};
+  if (props.newTab) {
+    properties = {target: "_blank"};
+  }
+  return (<Linkify properties={properties}>
+    {props.text.split("\n").map((item, key) => <span key={key}>{item}<br /></span>)}
```
potluck: @cozmo what will happen if the user inputs multiple new lines in their input text? will we end up having empty spans?

potluck: also, how do you think about using separate `<p>` elements rather than line breaks?

cozmo: Check out the example screenshot.

docs/components/RichTextView.jsx
cozmo: Yea that's true. Upon thinking about it more I just removed the option since I think they should always open in new tabs. If we see a case where we need it it's easy to add but don't want to give people an option they should never set.

cozmo:
@potluck I removed the newTab option and added a line of text from Carl - Are we good to go here?

PR #197, opened 2017-06-13
Add provided className to children of Wizard

osdiab:
**Jira:**
none, but this will help solve https://clever.atlassian.net/browse/CROC-902.

Discussion on Slack regarding why this is useful here: https://clever.slack.com/archives/C0ERS5ABB/p1497294991024954

**Overview:**
Use both the custom class name provided to the wizard as well as the `Wizard` classname for all classes internal to the wizard, so they can be styled without compound selectors

**Screenshots/GIFs:**
screen shot 2017-06-13 at 2 45 08 pm

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/WizardExample.jsx
+++ b/docs/WizardExample.jsx
@@ -240,6 +240,7 @@ export default class WizardExample extends React.Component {
       </fieldset>
 
       <Wizard
+        className="ExampleWizard"
         title="Delivery setup"
         description="Ensure that your delivery will come on time"
         stickySidebar={this.state.sticky}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.25.14",
+  "version": "0.25.15",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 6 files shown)


docs/WizardExample.jsx
```diff
@@ -11,13 +11,20 @@ class AddressStep extends React.Component {
   render() {
     const {setWizardState, wizardState, recipient} = this.props;
     return (
-      <form ref={(f) => { this.form = f; }} onSubmit={(e) => e.preventDefault()}>
+      <form
+        ref={f => {
+          this.form = f;
+        }}
+        onSubmit={e => e.preventDefault()}
+      >
```
kofi-clever: lot of noisy unrelated changes in here - is that intentional?

osdiab: Ah must be the eslint/vscode auto fix on save setting I was using :P i'll try to go through and undo some of this stuff

osdiab: Ah it was triggering since I set a max width of 80 chars on my own editor, I think it blindly applied it to everythign

osdiab: Yep, totally was VSCode's auto format on save thing. Disabled that, repushed

PR #203, opened 2017-07-10
Table: Wrap header cells.

samfishman:
**Before:**
screen shot 2017-07-10 at 11 43 24 am

**After:**
screen shot 2017-07-10 at 11 42 42 am

### DIFF
--- a/src/Table/HeaderCell.jsx
+++ b/src/Table/HeaderCell.jsx
@@ -4,6 +4,7 @@ import React, {PropTypes} from "react";
 import * as tablePropTypes from "./tablePropTypes";
 import Cell from "./Cell";
 import SortIcons from "./SortIcons";
+import {FlexBox, ItemAlign} from "../flex";
 
 require("./HeaderCell.less");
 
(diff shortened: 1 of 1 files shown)


src/Table/HeaderCell.jsx
```diff
@@ -25,14 +26,19 @@ export default function HeaderCell({
...
       )}
-      noWrap
       onClick={() => sortable && onSortChange()}
       width={width}
     >
-      <div className={cssClass.LABEL}>
-        {children}
-      </div>
-      {sortable && <SortIcons direction={activeSortDirection} className={cssClass.SORT} />}
+      <FlexBox alignItems={ItemAlign.CENTER} >
```
kofi-clever: fwiw, table cells already have built-in vertical alignment, so in theory, you should be able to do `vertical-align: middle` at the cell level to avoid having to use a flexbox

samfishman: The flexbox is to keep the sort arrows on the same line as the label. Is there an easy way to do that without flex?

kofi-clever: yeah, if the label is the only thing that needs to wrap, that can separately be changed to `white-space: normal` (while the cell stays `nowrap`)

PR #208, opened 2017-07-26
update selector-no-id

johnhuangclever:
Deprecation Warning: 'selector-no-id' has been deprecated and in 8.0 will be removed. Instead use 'selector-max-id' with '0' as its primary option. See: https://stylelint.io/user-guide/rules/selector-no-id/


potluck:
lgtm!

PR #209, opened 2017-07-26
InfoPanel: Allow any node for title.

samfishman:
So you can put HTML in InfoPanel titles without react yelling at you.


potluck:
Makes sense, LGTM

PR #211, opened 2017-07-29
Force tooltip to use Proxima Nova

alexandersoto:
ReactBootstrapTooltip forces the font family to Helvetica. This overrides that and forces the tooltip to Proxima Nova. The font size increases a little bit too...I had a choice of medium or small text and the small looked too small. We can change that if others think a smaller text size makes more sense.

Before:
image

After:
image

(Ignore the dark background, that's just Chrome's debugger in the paused state).

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.25.25",
+  "version": "0.25.26",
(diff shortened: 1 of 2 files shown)


potluck:
prob want a patch?

alexandersoto:
Yup! Patched. Thanks for reviewing :)

PR #212, opened 2017-08-01
Fix logic for next step skipping

amir-clever:
**Jira:**

**Overview:**
The wizard component did not correctly handle skipping the final step in the wizard and it also did not handle skipping multiple steps.

**Screenshots/GIFs:**
image

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [n/a] Safari
  - [n/a] IE10

**Roll Out:**
- Before merging:
  - [n/a] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.25.26",
+  "version": "0.25.27",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.25.26",
+  "version": "0.25.28",
```
kofi-clever: fyi: this is bumping +2 (unless there's another PR going in for 0.25.27)

amir-clever: Oh. No. Must have run it twice. Sorry. Will fix. Thanks for the review.

PR #213, opened 2017-08-04
Add a List component for lists of things

kofi-clever:
**Jira:** https://clever.atlassian.net/browse/FEE-41
**Layout Guide:** https://docs.google.com/document/d/14Qs75OMztK4YzxlJdiqLZ5VbmFKQA5SDe99Ws5u87FU

**Overview:**
Add a `<List />` components for all your "1-column table" needs!

After consultation with Mischelle, setting defaults to bordered list with plain rows.

**Screenshots/GIFs:**
screen shot 2017-08-03 at 5 39 29 pm

https://gyazo.com/eabf23ab5db8037fe859a427b653dc2c

**Testing:**
- [N/A] Unit tests -- Component's a little too simple for tests. Demo's sufficient for now.
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 125
+LINT_MAX_LESS_PROBLEMS := 322
 lint:
 	@echo "Linting files..."
 	@$(LINT) $(JS_FILES) $(JSX_FILES)
--- a/docs/components/ListView.jsx
+++ b/docs/components/ListView.jsx
@@ -43,9 +43,6 @@ export default class ListView extends Component {
     showBorder: true,
   };
 
-  componentWillMount() {
-  }
-
   renderConfig() {
     const {cssClass} = ListView;
     const {itemsKey, rowType, showBorder, showCustomEmptyMessage, showTitle} = this.state;
@@ -165,7 +162,6 @@ export default class ListView extends Component {
               name: "children",
               type: "List.Item | List.Item[]",
               description: "List items to display.",
-              optional: true,
             },
             {
               name: "className",
--- a/docs/components/ListView.jsx
+++ b/docs/components/ListView.jsx
@@ -1,0 +1,231 @@
+import _ from "lodash";
+import React, {Component} from "react";
+import {Link} from "react-router";
+
+import Example, {CodeSample, ExampleCode} from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+import {List, FlexBox, ItemAlign, SegmentedControl} from "src";
+
+import "./ListView.less";
+
+const Items = {
+  empty: [],
+  nonEmpty: [
+    "Raindrops on roses",
+    "Whiskers on kittens",
+    "Bright copper kettles",
+    "Warm woolen mittens",
+    "Cream-colored ponies",
+    "Crisp apple strudels",
+    "Door bells",
+    "Sleigh bells",
+    "Schnitzel with noodles",
+  ],
+};
+
+export default class ListView extends Component {
+  static cssClass = {
+    CONFIG_CONTAINER: "ListView--configContainer",
+    CONFIG_OPTIONS: "ListView--configOptions",
+    CONFIG: "ListView--config",
+    CONTAINER: "ListView",
+    INTRO: "ListView--intro",
+    LIST: "ListView--list",
+    PROPS: "ListView--props",
+  };
+
+  state = {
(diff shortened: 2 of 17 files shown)


Makefile
```diff
@@ -47,7 +47,7 @@ border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 125
+LINT_MAX_LESS_PROBLEMS := 322
```
CarlNelson: This seems like a lot of added problems...

bitmoji

src/List/List.jsx
```diff
@@ -0,0 +1,67 @@
...
+const RowType = {
+  PLAIN: "plain",
+  BORDERED: "bordered",
+  ZEBRA: "zebra",
+};
+
+const cssClass = {
+  CONTAINER: "List",
+  EMPTY: "List--empty",
+  HEADER: "List--header",
```
CarlNelson: This doesn't match our Dewey guidelines for components for naming. Only modifiers like noBorder or empty are camelCased, otherwise PascalCased.

src/Table/Table.jsx
```diff
@@ -271,6 +271,13 @@ export class Table extends Component {
     const {lazy, numRows} = this.props;
     const {currentPage, sortState, pageLoading, allLoaded} = this.state;
 
+    const columns = lodash.compact(React.Children.toArray(children));
+    if (columns.length < 2) {
+      throw new Error(
+        "Table requires at least 2 columns. Consider using the List component instead."
```
CarlNelson: 💯

Makefile
kofi-clever: haha yeah...we had a broken lint rule that's been deprecated. upgraded to the current one and these popped up. they're mostly from the auto-generated files that don't have any new lines between rulesets, so we should be able to bring the count down a bit in a follow-up if we update the script

src/List/List.jsx
kofi-clever: ah, the idea was to have PascalCase for top-level components (basically, match the javascript class name) - so far, we've been using camelCase for sub-elements and modifiers (i.e. `.List--Item` is a (sub-)component used with `List`, `.List--header` is a sub-element directly within `List`'s markup, `.List--header--small` would be a modified version of the header sub-element, etc.)

arsalansufi: TIL: Didn't realize we had this convention -- good to know!

docs/components/ListView.jsx
```diff
@@ -0,0 +1,231 @@
...
+import Example, {CodeSample, ExampleCode} from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+import {List, FlexBox, ItemAlign, SegmentedControl} from "src";
+
+import "./ListView.less";
+
+const Items = {
+  empty: [],
+  nonEmpty: [
```
peggyl: 💯 This example is amazing.

docs/components/ListView.jsx
```diff
@@ -0,0 +1,231 @@
...
+        </Example>
+
+        <PropDocumentation
+          title="<List /> Props"
+          availableProps={[
+            {
+              name: "children",
+              type: "List.Item | List.Item[]",
+              description: "List items to display.",
+              optional: true,
```
peggyl: Shouldn't this be required even if an empty list is accepted?

docs/components/ListView.jsx
```diff
@@ -0,0 +1,231 @@
...
+
+  state = {
+    itemsKey: "nonEmpty",
+    rowType: List.defaultProps.rowType,
+    showCustomEmptyMessage: false,
+    showTitle: true,
+    showBorder: true,
+  };
+
+  componentWillMount() {
```
peggyl: Why is this needed? The entire component lifecycle is invoked even without an empty function, right?

src/List/List.jsx
CarlNelson: We should clarify in the docs then, my assumption was that all elements were PascalCased and only modifiers were camelCased.

CarlNelson:
LGTM

docs/components/ListView.jsx
kofi-clever: oops, good catch - i copied this file from another components docs and didn't fully cleanup

docs/components/ListView.jsx
kofi-clever: yeah, fair point. functionally, it's optional, but in practice, there's no reason for anyone to provide null/undefined children

PR #216, opened 2017-08-07
TabBar: Align tabs to bottom.

samfishman:
From:
image

To:
image


potluck:
@samfishman curious if you explored other ways of achieving this / whether you see any tradeoffs to using flexbox here!

samfishman:
AFAIK, there's not a way to achieve this without flexbox, and all of the browsers we target support flexbox, so I don't think there's a real tradeoff

PR #218, opened 2017-08-08
App icon border radius

johnhuangclever:
10% for app icons

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/genBorderRadius.js
+++ b/genBorderRadius.js
@@ -10,6 +10,7 @@ const borderRadii = {
   m: 0.1875, // 3px
   l: 0.25, // 4px
   xl: 0.5, // 8px
+  "10percent": "10%", // for apps
 };
 
 const constants = [];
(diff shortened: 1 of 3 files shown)


genBorderRadius.js
```diff
@@ -10,16 +10,22 @@ const borderRadii = {
   m: 0.1875, // 3px
   l: 0.25, // 4px
   xl: 0.5, // 8px
+  "10percent": "10%", // for apps
```
kofi-clever: can just leave out the comment for this one

kofi-clever:
also, feel free to prepend a `// stylelint-disable max-line-length` to the file generation, since we don't really need to worry about that rule for generated code

PR #228, opened 2017-08-31
Fix header alignment in AlertBox

arsalansufi:
**Overview:**

The icon in the `AlertBox` header sits slightly higher than the title. Verified that this was not an intentional decision. This PR fixes the alignment!

**Screenshots/GIFs:**

*Before:*

before

*After:*

after-1

after-2

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**

- Before merging:
  - [ ] ~Updated docs~ *N/A*
    - Backward compatible change? Run `npm version patch` <--
- After merging:

### DIFF
--- a/src/AlertBox/AlertBox.jsx
+++ b/src/AlertBox/AlertBox.jsx
@@ -42,8 +42,8 @@ export default class AlertBox extends PureComponent {
     return (
       <div className={classnames(`AlertBox--${type}`, CLASSNAMES.CONTAINER, className)}>
         <div className={CLASSNAMES.HEADER}>
-          <span><Icon /></span>
-          <strong className={CLASSNAMES.TITLE}> {title} </strong>
+          <Icon />
+          <strong className={CLASSNAMES.TITLE}>{title}</strong>
           {isClosable &&
             <button className={CLASSNAMES.CLOSE} onClick={() => this.closeBox()}><CloseIcon /></button>
           }
--- a/src/AlertBox/AlertBox.jsx
+++ b/src/AlertBox/AlertBox.jsx
@@ -3,7 +3,7 @@ import classnames from "classnames";
 import React, {PureComponent, PropTypes} from "react";
 
 import {CloseIcon, WarningIcon, SuccessIcon, ErrorIcon, InfoIcon} from "./icons";
-import {FlexBox, ItemAlign} from "../flex";
+import {FlexBox, FlexItem, ItemAlign} from "../flex";
 
 import "./AlertBox.less";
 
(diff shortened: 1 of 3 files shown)


src/AlertBox/AlertBox.less
```diff
@@ -10,13 +10,15 @@
 }
 
 .AlertBox--title {
-  display: inline;
-  .margin--left--2xs;
+  .margin--left--xs;
   .text--semi-bold;
+  flex-grow: 1;
```
arsalansufi: This `flex-grow` makes sure that the close button sits at the right end of the header. (This wasn't necessary earlier because we weren't using flex styling for the header.)

src/AlertBox/AlertBox.jsx
```diff
@@ -42,8 +42,8 @@ export default class AlertBox extends PureComponent {
     return (
       <div className={classnames(`AlertBox--${type}`, CLASSNAMES.CONTAINER, className)}>
         <div className={CLASSNAMES.HEADER}>
```
amir-clever: Q. Why don't clever components use themselves? It seems like instead of using CSS we should use https://clever.github.io/components/#/components/flex-box ? Thoughts?

peggyl:
🎉 👍 I'm so excited about this. Thanks!!

src/AlertBox/AlertBox.jsx
arsalansufi: Fair point! I've generally just viewed the `FlexBox` component as a convenience. And this case felt simple enough to just use CSS directly

amir-clever: Let's FlexBox it up! I think it'll be cleaner code. :)

arsalansufi: That said, I'm realizing that, even if I don't use the `<FlexBox>` component, I should use the [flex mixins](https://github.com/Clever/components/blob/25fc1c1cef0d16391baebee1200aabc321b81ac9/src/less/flex.less#L11-L16) to ensure compatibility

amir-clever: Or if you think the lift is too much.... then LGTM

amir-clever: @arsalansufi if you're going down that route... then I would definitely use FlexBox. Way more maintainable.

src/AlertBox/AlertBox.jsx
```diff
@@ -41,13 +43,15 @@ export default class AlertBox extends PureComponent {
...
-        <div className={CLASSNAMES.HEADER}>
-          <span><Icon /></span>
-          <strong className={CLASSNAMES.TITLE}> {title} </strong>
-          {isClosable &&
-            <button className={CLASSNAMES.CLOSE} onClick={() => this.closeBox()}><CloseIcon /></button>
-          }
-        </div>
+        <FlexBox className={CLASSNAMES.HEADER} alignItems={ItemAlign.CENTER}>
+          <Icon />
+          <FlexBox className={CLASSNAMES.TITLE} grow>{title}</FlexBox>
```
amir-clever: Should this be FlexItem? Any reason to make it a FlexBox
http://clever.github.io/components/#/components/flex-box

arsalansufi: ^ Discussed offline, changing to a FlexItem

PR #235, opened 2017-10-18
Modal a11y

prime-time:
**Jira:**
https://clever.atlassian.net/browse/FEE-95

**Overview:**
Incorporated focus-lock-react library to help improve the accessibility of the modal. Focus is placed in the modal when it is opened instead of remaining in the background. Focus stays on elements within the modal when the user tabs and shift + tabs while the modal is open. Closing the modal returns focus to the element that had focus before the modal was opened. 

This feature is proposed to be toggleable via a prop in Modal.

**Screenshots/GIFs:**
screen recording 2017-10-17 at 04 08 pm

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


src/Modal/Modal.jsx
```diff
@@ -65,8 +74,10 @@ Modal.propTypes = {
   title: React.PropTypes.string.isRequired,
   closeModal: React.PropTypes.func.isRequired,
   children: React.PropTypes.node.isRequired,
+  isFocusTrap: React.PropTypes.bool,
```
kofi-clever: nit: s/`isFocusTrap`/`focusTrapped`/ or `focusLocked`?
since usage will usually look like `<Modal title="foo" focusLocked>...</Modal>`

prime-time: Good point. I like the latter, so let's go with that.

prime-time:
I clicked revert by accident

PR #236, opened 2017-10-18
Modal Accessibility

prime-time:
**Jira:**
https://clever.atlassian.net/browse/FEE-95

**Overview:**
The focus locking feature in the modal window is now the default behavior in order to help promote accessibility standards. Accessibility labels have been added to the modal to improve the experience for users with screen readers. 

**Screenshots/GIFs:**
screen shot 2017-10-18 at 2 19 45 pm

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Modal/Modal.jsx
+++ b/src/Modal/Modal.jsx
@@ -46,7 +46,7 @@ export class Modal extends React.Component {
     let contentStyle = {maxHeight: (this.state.windowHeight * 0.9) - 60};
     const modalContent = (
       <div className={classnames("Modal", this.props.className)}>
-        <div className="Modal--background" onClick={this.props.closeModal} />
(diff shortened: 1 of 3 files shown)


src/Modal/Modal.jsx
```diff
@@ -79,5 +79,5 @@ Modal.propTypes = {
 
 Modal.defaultProps = {
   width: DEFAULT_WIDTH,
-  focusLocked: false,
+  focusLocked: true,
```
kofi-clever: note: if we do default to true, this should be a major bump to `0.26.0` so that clients don't pull it in automatically

prime-time: Thanks, I'll make the change

PR #240, opened 2017-11-20
SegmentedControl fixes/updates.

kofi-clever:
**Jira:** N/A

**Overview:**
- Fix bug where the control is unresponsive to clicks when the explicit value is changed by the parent and the internal state goes out of sync.
- Also, prevent the buttons from wrapping - ends up looking weird in some UIs that are using flex styling.

**Before:**
Screenshot from Gyazo

**After:**
Screenshot from Gyazo

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [N/A] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.26.1",
+  "version": "0.26.2",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


kofi-clever:
thanks!

> Even though it might cause weirdness in other places, we can make the change and be on the lookout for weirdness imo.

true - considering maybe adding an explicit vertical mode, if we see a need for it, so that clients can choose to switch when they're space-constrained. then we can arrange and style the segments appropriately for that format

PR #243, opened 2017-12-01
fix Chrome breaking design choices

CarlNelson:
**Overview:**
Chrome broke the way we render buttons by making bad default choices. This fixes that.

**Screenshots/GIFs:**
Before:
image

Fixed:
image

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


package-lock.json
```diff
@@ -0,0 +1,11930 @@
+{
```
kofi-clever: can leave this file out - we're not currently checking in the lock files AFAIK

src/Button/Button.less
```diff
@@ -22,6 +22,7 @@ a, button {
   &.Button {
     .border--s(transparent);
     .borderRadius--top--m;
+    .borderRadius--bottom--0;
```
kofi-clever: could you also all a `.borderRadius--0` to the `.button--reset()` mixin? that will also fix other buttons like the ones in the left nav: https://github.com/Clever/components/blob/master/src/less/resets.less#L4

we'll need the same fix on the segmented control as well: https://github.com/Clever/components/blob/master/src/SegmentedControl/SegmentedControl.less#L7
(it should probably eventually be updated to use `.button--reset()` as well)

CarlNelson: Smart... will add that to this PR

src/less/resets.less
```diff
@@ -1,9 +1,9 @@
-@import (reference) "./animations";
-
+@import (reference) "../less/index";
```
kofi-clever: i think this could end up in circular references - can remove, since it's unused

PR #251, opened 2018-01-05
Make InfoPanel optionally collapsible

venalen:
**Overview:**
Right now `InfoPanel` is a static area to view text. This change allows it to collapse into the header (I'm using the `react-accessible-accordion` and `react-fontawesome` packages.

I've also added checkboxes so that we can preview the`InfoPanel` with the collapsible feature as well as what it looks like with the footer.

**Screenshots/GIFs:**

infopanel2

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


src/InfoPanel/InfoPanel.jsx
```diff
@@ -1,45 +1,79 @@
...
+import React, {Component, PropTypes} from "react";
+import FontAwesome from "react-fontawesome";
+import ReactCollapsible from "react-collapsible";
 
 import "./InfoPanel.less";
 
 
 /**
- * Base presentational component for the displaying information in  paneled format.
+ * Base presentational component for the displaying information in paneled format.
```
brollin: 💯

src/InfoPanel/InfoPanel.less
```diff
@@ -21,25 +22,47 @@
   }
 }
 
-.InfoPanel--header {
-  .border--bottom--s(@neutral_silver);
+.InfoPanel--collapsible {
```
brollin: Can probably remove this?

src/InfoPanel/InfoPanel.less
```diff
@@ -21,25 +22,47 @@
...
+}
+
+.InfoPanel--headerOpen {
   background-color: @neutral_off_white;
+  .border--bottom--s(@neutral_silver);
   padding: @size_s @size_m;
+  .text--regular;
+  color: @neutral_dark_gray;
+  font-family: "Proxima Nova";
+  margin: 0;
```
brollin: Nit: can use `.margin--none`
https://github.com/Clever/components/blob/master/src/less/spacing.less#L12

test/InfoPanel_test.jsx
```diff
@@ -12,9 +12,11 @@ describe("InfoPanel", () => {
...
-  it("renders title", () => {
+  it("renders title in the Collapsible trigger", () => {
+    const title = <div>Panel Title</div>;
     const panel = shallow(<InfoPanel title="Panel Title" />);
-    assert.equal(panel.find(`.${cssClass.TITLE}`).text(), "Panel Title");
+    assert.deepEqual(panel.find(`.${cssClass.COLLAPSIBLE}`).props().trigger,
+      title);
   });
 
   it("renders content", () => {
```
brollin: I'm thinking it would be good to do another test here like `it("renders no content when defaultOpen is false")`.  I think it would look just like this test but include a prop of `defaultOpen={true}`, and then `assert.equal` that the `.panel.find()` returns nothing.  At least I'm pretty sure that's the way to test this case.  Happy to pair if it becomes tricky!

src/InfoPanel/InfoPanel.less
venalen: Ah yeah, I added that in so we could find this class in the unit test, otherwise I felt like I was being too specific on how to drill down into the DOM. I'm open to suggestions on a better way to do this though!

brollin: Gotcha - so the class's existence makes sense but if you delete these two lines, 25 and 26, does it pose a problem?  I _think_ you shouldn't need them?

brollin: Huh - when I load your branch locally, these lines are gone.  I may be missing something but in any case, the lines that I originally commented on and seemed extraneous were:

```
.InfoPanel--collapsible {
}
```

src/InfoPanel/InfoPanel.jsx
```diff
@@ -1,45 +1,79 @@
...
+      collapseArrow = <FontAwesome name="caret-down" />;
+    } else if (collapsible && !isCollapsed) {
+      collapseArrow = <FontAwesome name="caret-right" />;
+    }
+
+    return (
+      <div className={classnames(cssClass.CONTAINER, className)}>
+        <ReactCollapsible trigger={collapsible ?
+          (<div>
+            <div className={cssClass.COLLAPSE_ARROW}>{collapseArrow}</div>
```
CarlNelson: The collapse toggle is inaccessible as a div without aria roles.

I'd recommend taking a look at accessible accordions to model after like this one: https://github.com/springload/react-accessible-accordion#accessibility

test/InfoPanel_test.jsx
```diff
@@ -13,15 +13,27 @@ describe("InfoPanel", () => {
...
+      "Panel Title");
+  });
+
+  it("has is-closed class from react-collapsible when defaultOpen is false", () => {
+    const content = <div>Panel Content</div>;
+    const panel = render(<InfoPanel title="Panel Title" collapsible defaultOpen={false}>{content}</InfoPanel>);
+    assert(panel.find(`.${cssClass.HEADER_CLOSED}`).hasClass("is-closed"));
+  });
+
+  it("has is-open class from react-collapsible when defaultOpen is false", () => {
```
brollin: Should read `when defaultOpen is true`

src/InfoPanel/InfoPanel.jsx
brollin: Thanks for the find, Carl!  I'm looking into this with Vicky and wondering if we get keyboard accessibility and the aria tags/role set up correctly, think we're fine to use `react-collapsible`?  Or do you think we should switch to something that was built with accessibility in mind, ie `react-accessible-accordion`?

CarlNelson: We would have to test it to make sure, I'm uncertain as to how react-collapsible deals with accessibility.

src/InfoPanel/InfoPanel.jsx
```diff
@@ -1,48 +1,41 @@
...
     this.state = {
-      isCollapsed: props.defaultOpen,
+      isCollapsed: !props.defaultOpen,
+      collapseArrow: (props.defaultOpen ? "caret-down" : "caret-right"),
     };
   }
 
+  toggleArrow(keys) {
+    console.log(this.state.isCollapsed);
+    console.log(keys);
```
brollin: You can remove these logs 🌲

src/InfoPanel/InfoPanel.less
```diff
@@ -22,17 +21,30 @@
...
+  .border--bottom--s(@neutral_silver);
+  background-color: @neutral_off_white;
   padding: @size_s @size_m;
 }
 
+.InfoPanel--title {
+  .text--regular;
+  color: @neutral_dark_gray;
+  font-family: "Proxima Nova";
+  margin: 0;
```
brollin: Nit: `.margin--none`, `.padding--none`

package.json
```diff
@@ -23,10 +23,12 @@
     "moment": "^2.18.1",
```
brollin: Oh also we'll want to bump the package version

PR #254, opened 2018-01-11
Allow the option for textareas to autogrow

cozmo:
Adds an option to allow textareas to autogrow, defaults to false

### DIFF
--- a/src/TextArea/TextArea.jsx
+++ b/src/TextArea/TextArea.jsx
@@ -1,5 +1,6 @@
 import React from "react";
 import classnames from "classnames";
+import TextareaAutosize from "react-autosize-textarea";
 
 import "./TextArea.less";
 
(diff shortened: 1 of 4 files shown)


src/TextArea/TextArea.jsx
```diff
@@ -80,28 +81,35 @@ export class TextArea extends React.Component {
...
+      readOnly: this.props.readOnly,
+      ref: "input",
+      required: this.props.required,
+      spellCheck: this.props.spellCheck,
+      value: this.props.value,
+    };
+
+    let textarea = <textarea {...props} />;
+    if (this.props.autoResize) {
+      textarea = <TextareaAutosize rows={2} {...props} />;
```
ahsojar: suggestion: let the user pass in a `rows` value

PR #262, opened 2018-02-02
Add support for a date time picker in DateInput component

peternga:
**Jira:**
https://clever.atlassian.net/browse/IL-160

**Overview:**
As part of IL-160, we're allowing users to select DateTime for start and end. This change allows us to use DateTime in the DateInput component. The problem with the library we currently use is two-fold:
* We're working off a branched version so we can't use functionality introduced in newer versions that somewhat support DateTime input
* It doesn't work in Modals because it moves the calendar portion to the top of the DOM instead of being a child of the Modal

This change introduces the react-datetime npm library: https://github.com/YouCanBookMe/react-datetime which doesn't suffer the problems listed above.

**Screenshots/GIFs:**
http://g.recordit.co/KBVS1CNzvz.gif

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10
  - [x] Firefox

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/components/DateInputView.jsx
+++ b/docs/components/DateInputView.jsx
@@ -16,6 +16,7 @@ export default class DateInputView extends Component {
       readOnly: false,
       hasError: false,
       required: false,
+      useTime: false,
       value: null,
     };
   }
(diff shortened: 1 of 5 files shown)


peternga:
Planning on `npm version patch`

PR #266, opened 2018-02-10
add onClick to list item

johnhuangclever:
screenshot 2018-02-09 20 22 50

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/List/Item.tsx
+++ b/src/List/Item.tsx
@@ -13,6 +13,8 @@ export default class Item extends React.PureComponent {
 
   static cssClass = {
     CONTAINER: "List--Item",
+    CLICKABLE_CONTAINER: "List--Item--Clickable",
+    CLICKABLE_ITEM: "List--Item--ClickableItem",
 
     type: type => `List--Item--${type}`,
   };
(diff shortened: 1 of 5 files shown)


src/List/Item.tsx
```diff
@@ -18,10 +19,10 @@ export default class Item extends React.PureComponent {
 
   render() {
     const {cssClass} = Item;
-    const {children, className} = this.props;
+    const {children, className, onClick} = this.props;
 
     return (
-      <li className={classnames(cssClass.CONTAINER, className)}>
+      <li className={classnames(cssClass.CONTAINER, className)} onClick={onClick}>
```
kofi-clever: for a11y, maybe we should wrap `{children}` in a <button> if an onClick is specified so that the interaction's more standard (pointer shows up, user can tab through items, etc)

and then we should probably also add a hover state - darken the background a bit for that row?

we can move the padding that's currently on `cssClass.CONTAINER` onto another class and assign that to either the <button> or a plain <div> depending on whether or not onClick is specified

johnhuangclever:
So I'm not sure if my css/html is organized in the best way. Feels kind of ugly, but I'm not sure how to clean it up. Also because of the transition on the item and the padding changing between elements, toggling the onClick makes it look really jumpy.

components-list-onclick

 Would love some guidance here. Thanks!

johnhuangclever:
Also added a optional prop for classname to the `<ul>` - 393a3b5

PR #270, opened 2018-02-27
Update react-select, make div transparent

brollin:
**Overview:**
There's a styling bug that occurs when repos like `sd2` use `react-select@1.2.1`, where a white div appears over text. This PR upgrades the `react-select` dependency to `1.2.1` and then makes that div transparent.  It appears the div does not display anything in our usages of the component, so this is a safe change.

**Screenshots/GIFs:**
Previously:
screen shot 2018-02-26 at 6 00 29 pm

Now:
screen shot 2018-02-26 at 6 05 18 pm

**Testing:**
- [x] Local testing

**Roll Out:**
- Before merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "react-linkify": "^0.2.1",
     "react-onclickoutside": "^5.11.1",
     "react-overlays": "^0.7.0",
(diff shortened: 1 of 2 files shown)


cozmo:
praise 🙌

PR #272, opened 2018-03-13
Added function type to Wizard Step

venalen:
**Overview:**
This PR changes the `nextButtonValue` of the `WizardStep` to be able to accept a function as well as the already accepted node and string. The reasoning behind this is to allow for dynamically changing button values based on the `wizardState` which can only be accessed from within the wizard.

I also removed `nextButtonValue` from the `WizardStep` file because we're generating the buttons in the actual wizard instead.

**Screenshots/GIFs:**
Note how in Step 3 the `Next` button changes to `Save and exit`
badgewizard

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/Wizard/Wizard.jsx
+++ b/src/Wizard/Wizard.jsx
@@ -226,7 +226,6 @@ export class Wizard extends React.Component {
             help={curStep.help ? curStep.help : help}
             validate={curStep.validate}
             prevButtonValue={curStep.prevButtonValue}
-            nextButtonValue={curStep.nextButtonValue}
           />
 
           <div
@@ -247,7 +246,12 @@ export class Wizard extends React.Component {
               className={classNameFor(baseClasses, "nextButton")}
               onClick={this.nextStepHandler}
               disabled={nextDisabled} type="primary"
-              value={curStep.nextButtonValue || nextButtonValue || "Next"}
+              value={
+                typeof curStep.nextButtonValue === "function" ?
+                curStep.nextButtonValue(this.state.data) : curStep.nextButtonValue ||
+                nextButtonValue ||
+                "Next"
+              }
             />
           </div>
         </div>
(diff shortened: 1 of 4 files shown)


src/Wizard/Wizard.jsx
```diff
@@ -276,6 +280,7 @@ Wizard.propTypes = {
       PropTypes.func,
       PropTypes.instanceOf(React.Component),
     ]).isRequired,
+    nextButtonValue: PropTypes.oneOfType([PropTypes.string, PropTypes.node, PropTypes.func]),
```
potluck: Should the `nextButtonValue` prop on line 291 (top-level prop on `Wizard`) also accept a function?

src/Wizard/WizardStep.jsx
```diff
@@ -116,7 +116,6 @@ WizardStep.propTypes = {
   help: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   Component: PropTypes.oneOfType([PropTypes.func, PropTypes.instanceOf(React.Component)]).isRequired,
   componentProps: PropTypes.object,
-  nextButtonValue: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
   prevButtonValue: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
potluck: For consistency, would it make sense to move `prevButtonValue` into `Wizard` as well, while we're making this change, since that's also generated in the wizard?

src/Wizard/Wizard.jsx
venalen: Great question! We were thinking about that, but we couldn't think of a use-case where globally changing the button values would be helpful. That being said, if you think it would be best to keep it consistent then let me know!

src/Wizard/WizardStep.jsx
venalen: Sounds good. The button's already rendered in the Wizard so I'll remove these unused props. Thanks!

PR #283, opened 2018-04-04
Upgrade react

prime-time:
**Jira:**
https://clever.atlassian.net/browse/FEE-108

**Overview:**
This change upgrades react to the latest version. Most of the work to make components react 16 compatible was done in previously merged pull requests. One unit test did cause a console warning due to react 16's new error handling of mounted components. Shallow rendering the component in the test removes the warning.

Version 1.0.0?

https://reactjs.org/blog/2017/09/26/react-v16.0.html

Relevant to the quest for React 16: 
https://github.com/Clever/components/pull/274
https://github.com/Clever/components/pull/275
https://github.com/Clever/components/pull/281
https://github.com/Clever/minimal-react-typings/pull/6

**Screenshots/GIFs:**

**Testing:**
I tested this by running components locally, and I also pulled it into one of the dashboards running locally.

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


prime-time:
Sure, I'm excited to discuss this in more detail with the guild.

prime-time:
The new version of React warns about `setState` calls on unmounted components. Updated the `Table` component to bail out in the async fetch data function if the component unmounts.
https://stackoverflow.com/questions/39767482/is-there-a-way-to-check-if-the-react-component-is-unmounted
screen shot 2019-01-24 at 2 03 18 pm

kofi-clever:
would be curious what others in the guild think, but i'd be down with this kicking off v1.0.0

package.json
```diff
@@ -87,9 +87,8 @@
     "mocha": "^2.4.5",
     "postcss-loader": "^1.1.1",
     "prismjs": "^1.5.1",
-    "react": "~15.3.0",
-    "react-addons-test-utils": "~15.3.2",
-    "react-dom": "~15.3.0",
+    "react": "^16.7.0",
```
kofi-clever: let's also add a peer dependency for `"react": ">=16.0.0 <17.0.0"` so that clients at least get a warning if they're not on react16. (even though we're not using any strict v16 features here yet, that'll make sure that we can if we choose)
e.g. https://github.com/Clever/goals-components/blob/master/package.json#L35

prime-time:
Thanks for taking another look at this. I was just going to bump it to 0.38.0, but I checked with the frontend guild and everyone is fine about going to v1.0.0

PR #287, opened 2018-04-20
make the TextArea component space better

schimmy:
**Overview:**
The TextArea component kept:
1) Jumping around in size when going from placeholder to input text
2) Overlapping with the OPTIONAL / REQUIRED keywords on the top

**Screenshots/GIFs:**
before - jumping around sizes:
before

after:
after

before - interference from top bar on placeholder text:
screen shot 2018-04-19 at 16 56 16

after:
screen shot 2018-04-19 at 18 16 30

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/TextArea/TextArea.less
+++ b/src/TextArea/TextArea.less
@@ -2,6 +2,7 @@
 
 .TextArea {
   .padding--left--s;
+  .padding--y--3xs;
   .text--regular;
   .border--s(@neutral_silver);
   position: relative;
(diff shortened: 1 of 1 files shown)


src/TextArea/TextArea.less
```diff
@@ -9,6 +9,9 @@
   background: @neutral_white;
   box-sizing: border-box;
   width: 100%;
+  min-height: 3.5rem;
+  padding-top: 0.125rem;
+  padding-bottom: 0.125rem;
 }
```
ahsojar: I would use size vars whenever possible to make sure we stick to our design system guidelines: https://github.com/Clever/components/blob/828fb17feae2ab568f95e3c14088c463004a6b9c/src/less/sizing.less

And for things like padding top/bottom, you can use `.padding--y`: https://github.com/Clever/components/blob/828fb17feae2ab568f95e3c14088c463004a6b9c/src/less/spacing.less#L193

PR #291, opened 2018-05-24
TextTruncate component

johnhuangclever:
**Jira:**
Basic component to toggle truncating text

**Screenshots/GIFs:**
truncate-text

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/TextTruncate/TextTruncate.less
+++ b/src/TextTruncate/TextTruncate.less
@@ -1,0 +1,1 @@
+@import (reference) "../less/utilities";
--- a/src/TextTruncate/TextTruncate.tsx
+++ b/src/TextTruncate/TextTruncate.tsx
@@ -1,0 +1,79 @@
+import * as classnames from "classnames";
+import * as PropTypes from "prop-types";
+import * as React from "react";
+
+import {Button} from "../Button/Button";
+
+import "./TextTruncate.less";
+
+const propTypes = {
+  className: PropTypes.string,
+  text: PropTypes.string.isRequired,
+  more: PropTypes.string,
+  less: PropTypes.string,
+  lines: PropTypes.number,
+  characters: PropTypes.number,
+};
+
+const defaultProps = {
+  characters: 300,
+  more: "Show more",
+  less: "Show less",
+};
+
+const cssClass = {
+  CONTAINER: "TextTruncate",
+  TEXT: "TextTruncate--text",
+};
+
+/**
+ * Truncate long text
+ * Comes with read more/less toggle
+ * Adapted from https://www.npmjs.com/package/react-truncate#integrated-example-for-toggling-read-more-text
+ */
+export default class TextTruncate extends React.PureComponent {
+  static propTypes = propTypes;
+  static defaultProps = defaultProps;
+  static cssClass = cssClass;
+
+  state = {
+    truncated: true,
+  };
+
+  toggleTruncation(event) {
+    event.preventDefault();
+
+    this.setState({truncated: !this.state.truncated});
+  }
+
+  truncate(text) {
+    return text.substring(0, this.props.characters);
+  }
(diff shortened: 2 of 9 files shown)


src/TextTruncate/TextTruncate.less
```diff
@@ -0,0 +1 @@
+@import (reference) "../less/utilities";
```
brollin: Seems like this isn't needed?

src/TextTruncate/TextTruncate.tsx
```diff
@@ -0,0 +1,79 @@
...
+const defaultProps = {
+  characters: 300,
+  more: "Show more",
+  less: "Show less",
+};
+
+const cssClass = {
+  CONTAINER: "TextTruncate",
+  TEXT: "TextTruncate--text",
+};
```
brollin: Looks like these aren't used?

src/TextTruncate/TextTruncate.tsx
```diff
@@ -0,0 +1,79 @@
...
+          </div> :
+          <div className={classnames(cssClass.CONTAINER, className)}>
+            {text} <Button type="linkPlain" onClick={(e) => this.toggleTruncation(e)} value={less} />
+          </div>
+        }
+      </div>
+    );
+  }
+
+  _handlePerformAction = () => this.props.onPerformAction("action performed");
```
kofi-clever: can you pattern-match this for the toggleTruncation method? that way you can pass the function in as is to `onClick` without having to create the anonymous function on every render
(and also reminder to remove this!)

src/TextTruncate/TextTruncate.tsx
```diff
@@ -0,0 +1,79 @@
...
+        {text}
+      </div>);
+    }
+
+    return (
+      <div>
+        {truncated ?
+          <div className={classnames(cssClass.CONTAINER, className)}>
+            {this.truncate(text)}
+            <Button type="linkPlain" onClick={(e) => this.toggleTruncation(e)} value={`... ${more}`} />
```
kofi-clever: nit: the ellipsis look a little weird as part of the button - should they just trail the text and then put the space between the ellipsis and the button itself?

src/TextTruncate/TextTruncate.tsx
```diff
@@ -0,0 +1,79 @@
...
+
+import {Button} from "../Button/Button";
+
+import "./TextTruncate.less";
+
+const propTypes = {
+  className: PropTypes.string,
+  text: PropTypes.string.isRequired,
+  more: PropTypes.string,
+  less: PropTypes.string,
```
kofi-clever: nit: let's be more explicit with these prop names, so it's obvious what they do: like `showMoreLabel` and `showLessLabel` or something

src/TextTruncate/TextTruncate.tsx
```diff
@@ -0,0 +1,79 @@
...
+  lines: PropTypes.number,
+  characters: PropTypes.number,
```
kofi-clever: opt: `maxCharacters` or `maxChars`?

johnhuangclever: Changed to `maxCharsShown`

src/TextTruncate/TextTruncate.tsx
johnhuangclever: Yeah sounds good!

src/TextTruncate/TextTruncate.tsx
johnhuangclever: Yup. Thanks!

PR #303, opened 2018-08-31
Don't let other elements poke through the calendar popup

prime-time:
**Jira:**
http://clever.atlassian.net/browse/soc-1767

**Overview:**
`react-popper` is used by `react-datepicker` to render the calendar popup in the `DateInput` component. [The z-index is set to 1](https://github.com/Hacker0x01/react-datepicker/blob/d25d0d7eb617800e943ef749d7f31cd6daec6fed/src/stylesheets/datepicker.scss#L43), so other components with higher z-index values can poke through. `SegmentedControl` uses z-index 2 for example. This change sets a higher z-index to make sure that the calendar popup is always in the forefront of the page.

I used `.zIndex--3` which is actually 300. Is this preferable over setting the z-index to 3?
https://github.com/Clever/components/blob/master/src/less/z-index.less

**Screenshots/GIFs:**
screen shot 2018-08-30 at 8 04 35 pm
screen shot 2018-08-30 at 8 04 23 pm

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -27,7 +27,7 @@
     "react-autosize-textarea": "^3.0.1",
     "react-bootstrap": "0.32.1",
     "react-copy-to-clipboard": "^4.2.1",
-    "react-datepicker": "1.2.2",
+    "react-datepicker": "^1.6.0",
     "react-datetime": "^2.14.0",
     "react-dropzone": "^3.11.0",
     "react-fontawesome": "^1.6.1",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.30.7",
+  "version": "0.30.8",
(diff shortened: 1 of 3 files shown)


johnhuangclever:
z-index always causes trouble :(

I wonder if we could make segmentedcontrol not have to rely on z-index, but this is fine

prime-time:
There's a comment about the z-index in `SegmentedControl` [here](https://github.com/Clever/components/blob/9a6c373e41cbca6b8519fe9fd9557e4def8b044d/src/SegmentedControl/SegmentedControl.less#L50). Since it was intentional, I thought it would be best to leave it.

PR #306, opened 2018-09-04
Optionally hide labels on the FileInput component

jish:
**Overview:**

We would like to reuse the functionality of the `FileInput` component
without having the labels present.

This adds a new option `displayLabels` which can be set to `false` to
hide the labels on the `FileInput` component. The default value of
`displayLabels` is `true`. This allows all of the existing code using
the `FileInput` component to remain unchanged.

**Screenshots/GIFs:**

upload-component-no-labels

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/components/FileInputView.jsx
+++ b/docs/components/FileInputView.jsx
@@ -5,6 +5,28 @@ import PropDocumentation from "./PropDocumentation";
 import View from "./View";
 import {FileInput} from "src";
 
+import "./FileInputView.less";
+
+function isError() {
+  return false;
+}
+
+function storeFn(_file, { progress, success }) {
+  let currentProgress = 0;
+  progress(currentProgress);
+
+  const interval = setInterval(() => {
+    currentProgress += 2;
+
+    if (currentProgress < 100) {
+      progress(currentProgress);
+    } else {
+      clearInterval(interval);
+      success();
+    }
+  }, 32);
+}
+
 export default function FileInputView() {
   const {cssClass} = FileInputView;
 
@@ -25,7 +47,7 @@ export default function FileInputView() {
                   if (progressPercent !== 100) {
                     progressPercent += 5;
                     progress(progressPercent);
-                  } else if (this.state.error) {
+                  } else if (isError()) {
                     error("An error occurred");
                     clearInterval(intervalID);
                   } else {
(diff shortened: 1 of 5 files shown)


jish:
The most recent build errored out

> "Too long with no output (exceeded 10m0s)"

When I pressed the "Rebuild" button, it passed: https://circleci.com/gh/Clever/components/1049

docs/components/FileInputView.jsx
```diff
@@ -65,13 +102,21 @@ export default function FileInputView() {
...
             description: "Function to be called when the user has seleceted a file to be uploaded",
             defaultValue: "False",
           },
+          {
+            name: "displayLabels",
+            type: "Boolean",
+            description: "Display text labels inside of the component",
+            defaultValue: "true",
+            optional: true,
+          },
```
ahsojar: curious, what was the reasoning behind having this prop as well, rather than relying on `label` existing or not?

docs/components/FileInputView.jsx
```diff
@@ -25,7 +47,7 @@ export default function FileInputView() {
                   if (progressPercent !== 100) {
                     progressPercent += 5;
                     progress(progressPercent);
-                  } else if (this.state.error) {
+                  } else if (isError()) {
```
ahsojar: if this is always false, do do we need this `else if` here?

docs/components/FileInputView.jsx
jish: We thought it would be too confusing to have `label=""` cause many side-effects inside of the component. It hides multiple HTML elements, and seemed too magical.

docs/components/FileInputView.jsx
jish: The existing example `FileInput` demo page is currently throwing an error on this line https://clever.github.io/components/#/components/file-input

screen shot 2018-09-05 at 11 42 46 am

I believe this function exists mostly as documentation. You can see it in the code example on the page, and when you mock "upload" a file, this function fills the progress indicator over time.

Unfortunately, in combination with the `setInterval()` call and the `if (progressPercent !== 100)` check, it means that the page will continuously throw errors forever in a loop, and the success icon will never be displayed.

I thought I would fix up this error while I was at it, but to be honest, this change is not even related to my changes nor the second example I created down lower on the page. I can back this out if you like.

ahsojar: oh i see, it's purely to show example code. Fine to leave it as is!

If you wanted to, you could add a toggle on the documentation to 'show error state' so they can see what happens in this example, though it wouldn't be too exciting haha

PR #308, opened 2018-09-06
Make the ReactDatePicker usage fixed height by default

natebrennand:
Gifs coming

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.30.9",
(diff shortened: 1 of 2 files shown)


cozmo:
@natebrennand Although up to you if you want to apply it to the `DateTimePicker` or not?

natebrennand:
1. the affected usage is not for `useTime=true`
2. I don't see a prop for it in their docs - https://github.com/YouCanBookMe/react-datetime#api

PR #309, opened 2018-09-07
Textarea default rows

johnhuangclever:
Error from console:
screenshot 2018-09-07 11 40 09

Stepping through old code:
textarea-broken

Stepping through new code:
textarea-fixed

Verified that visual sizing of textarea is the same with the new code

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/TextArea/TextArea.jsx
+++ b/src/TextArea/TextArea.jsx
@@ -82,7 +82,7 @@ export class TextArea extends React.Component {
       inputNote = <span className="TextArea--error">{this.props.error}</span>;
     }
 
-    const props = {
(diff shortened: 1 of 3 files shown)


src/TextArea/TextArea.jsx
```diff
@@ -145,3 +145,7 @@ TextArea.propTypes = {
   autoResize: PropTypes.bool,
   rows: PropTypes.number,
 };
+
+TextArea.defaultProps = {
```
brollin: TIL about `defaultProps`! Surprised I hadn't run into this before...

PR #323, opened 2018-10-31
fix customIcon propType

aastein:
**Jira:**

**Overview:**
customIcon is a react node, which is an object not a function.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.33.5",
+  "version": "0.33.6",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


cozmo:
Fun fact nodes can actually be either. In the case of pure components (props => <div>hello</div>) they _will_ be functions.

Luckily you can use PropTypes.node for this exact case.

cozmo:
Mispoke - Even with pure components the output is still an object 🤦‍♂️ - that said the node proptype is still the more specific/correct one here so might as well use it

aastein:
TIL, didn't know Proptypes.node was a thing. Made a change to the class name as well.

PR #328, opened 2018-12-04
Add a WithKeyboardNav wrapper component

kofi-clever:
**Jira:** https://clever.atlassian.net/browse/GOAL-206

**Overview:**
Needed for upcoming RadioGroup component.

Enables stepping through a defined list of items using keyboard nav keys, for a11y and convenience purposes.

We've had to do this in a few places across Clever, so trying to consolidate in a single reusable component.

**Screenshots/GIFs:**
[Screenshot from Gyazo](https://gyazo.com/8fa00d2d95fa359fc3927ac124862f6b)

**In RadioGroup context:**
[Screenshot from Gyazo](https://gyazo.com/50fd0d1f43ee0d581f3325b0fefa2d1c)

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


prime-time:
Did you mean to import `WithKeyboardNav.less` ? I don't see it imported anywhere.

src/WithKeyboardNav/WithKeyboardNav.tsx
```diff
@@ -0,0 +1,100 @@
...
+  }
+
+  _handleKeyDown = event => {
+    const {backKeys, forwardKeys, forceTabKey} = this.props;
+
+    if (!_.isString(event.key)) {
+      return;
+    }
+
+    if (new Set(forwardKeys).has(event.key)) {
```
prime-time: Is there an advantage of using a set here instead of just using the array?

src/WithKeyboardNav/WithKeyboardNav_test.tsx
```diff
@@ -0,0 +1,183 @@
...
+    simulateKeyDownEvent(wrapper, {key: KeyCode.K});
+    expect(onChangeMock).toHaveBeenCalledWith(ITEM_IDS[2]);
+  });
+});
+
+function simulateKeyDownEvent(wrapper, {key, shiftKey = false}) {
+  wrapper
+    .find("div")
+    .getDOMNode()
+    .dispatchEvent(new KeyboardEvent("keydown", {key, shiftKey}));
```
prime-time: There's an enzyme function for this too https://airbnb.io/enzyme/docs/api/ShallowWrapper/simulate.html#simulateevent-args--self

src/WithKeyboardNav/WithKeyboardNav.tsx
kofi-clever: ah no, i was storing the set for something else before and didn't change it in the refactoring.

src/WithKeyboardNav/WithKeyboardNav_test.tsx
kofi-clever: that one doesn't quite do what i need in this case - it emits a React synthetic event which would trigger the `onKeyDown` prop of the div, but since we're listening for pure browser events, we need to dispatch the event directly on the DOM node

prime-time: ah gotcha

PR #332, opened 2018-12-05
Add RadioGroup component

kofi-clever:
[Might be easier reviewing commits separately]

**Jira:** https://clever.atlassian.net/browse/GOAL-206

**Spec:** https://invis.io/6YPCS27X2JP#/333969277_radio_Buttons_In_Dewey

**Overview:**
Adding a `RadioGroup` component for single-choice selection from a small list of options.

**Screenshots/GIFs:**
[Screenshot from Gyazo](https://gyazo.com/fda852c5e89cef47e1d371f691b4907b)

**Testing:**
- [n/a] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/RadioGroupView.jsx
+++ b/docs/components/RadioGroupView.jsx
@@ -1,0 +1,198 @@
+import * as React from "react";
+
+import Example, {CodeSample, ExampleCode} from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+import {RadioGroup, FlexBox, ItemAlign} from "src";
+
+import "./RadioGroupView.less";
+
+const cssClass = {
+  CONFIG_CONTAINER: "RadioGroupView--configContainer",
+  CONFIG_OPTIONS: "RadioGroupView--configOptions",
+  CONFIG: "RadioGroupView--config",
+  CONFIG_TOGGLE: "RadioGroupView--configToggle",
+  CONTAINER: "RadioGroupView",
+  INTRO: "RadioGroupView--intro",
+  PROPS: "RadioGroupView--props",
+};
+
+export default class RadioGroupView extends React.PureComponent {
+  static cssClass = cssClass;
+
+  state = {
+    disableAll: false,
+    selectedCity: null,
+    selectedFood: null,
+    requireSelection: false,
+  };
+
+  render() {
+    const {disableAll, requireSelection, selectedCity, selectedFood} = this.state;
+
+    return (
+      <View
+        className={cssClass.CONTAINER}
+        title="RadioGroup"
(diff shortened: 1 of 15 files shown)


docs/components/RadioGroupView.jsx
```diff
@@ -0,0 +1,198 @@
...
+        </header>
+
+        <Example title="Basic Usage:">
+          <p>
+            Use the <code>TAB</code> key to move between radio groups and use the arrow keys to move
+            within radio groups.
+          </p>
+          <ExampleCode>
+            <RadioGroup
+              label="Favourite City"
```
arsalansufi: That British spelling tho :laughing:

src/RadioGroup/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+
+  componentDidMount() {
+    this._element = ReactDOM.findDOMNode(this);
+  }
+
+  render() {
+    const {checked, children, className, disabled, tabIndex} = this.props;
+
+    return (
+      <button
```
johnhuangclever: Why is this a button instead of an input type=radio?

src/RadioGroup/Radio.tsx
```diff
@@ -0,0 +1,91 @@
...
+  _labelID = getNextLabelID();
+
+  focus() {
+    if (this._element) {
+      this._element.focus();
+    }
+  }
+
+  componentDidMount() {
+    this._element = ReactDOM.findDOMNode(this);
```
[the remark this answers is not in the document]
kofi-clever: oh yeah, we're still blocked from using React 16-specific features in clever-components directly, since not all dependent repos have been migrated to 16 yet. we'll need to do an update across the repo once we've done that (i think it's just HM remaining 🤞)

src/RadioGroup/Radio.tsx
kofi-clever: it's not possible to arbitrarily style certain native browser elements like the radio button and checkbox, so a common pattern is to use custom elements and assign a11y roles to them so they're still understood by tools like screen readers.

some custom implementations will often create a hidden radio input in addition to this and set its value under the covers to enable native form submit behaviour (i think material does this) but not really necessary in our case, since we're exclusively using AJAX requests to submit form data

src/RadioGroup/Radio.tsx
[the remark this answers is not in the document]
arsalansufi: Ah right totally forgot that components weren't on React 16 yet

PR #338, opened 2018-12-14
switch component

aastein:
**Jira:**

**Overview:**
https://projects.invisionapp.com/d/main/default/#/console/16080002/335819621/inspect

Switch component. Has three independent display states: checked, disabled, and focused. Focusing is only possible by tabbing to the element. When the component is disabled it cannot be interacted with.

**Screenshots/GIFs:**
screen recording 2018-12-14 at 02 30 pm

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/CloseIcon/CloseIcon.jsx
+++ b/src/CloseIcon/CloseIcon.jsx
@@ -1,0 +1,30 @@
+import * as React from "react";
+import classnames from "classnames";
+import * as PropTypes from "prop-types";
+
+
+const cssClass = {
+  CONTAINER: "CloseIcon--container",
+};
+
+CloseIcon.propTypes = {
+  className: PropTypes.string,
+};
+
+export function CloseIcon(props) {
+  const {className} = props;
+  return (
+    <svg
+      focusable="false"
+      className={classnames(cssClass.CONTAINER, className)}
+      viewBox="0 0 10 10"
+      xmlns="http://www.w3.org/2000/svg"
+    >
+      <path
+        d="M9.934 7.916c.04.052.066.105.066.171a.278.278 0 0 1-.066.172L8.259 9.934a.278.278 0 0 1-.172.066.278.278 0 0 1-.171-.066L5 7.018 2.084 9.934a.278.278 0 0 1-.171.066.278.278 0 0 1-.172-.066L.066 8.259A.278.278 0 0 1 0 8.087c0-.066.026-.119.066-.171L2.982 5 .066 2.084A.278.278 0 0 1 0 1.913c0-.066.026-.119.066-.172L1.741.066A.278.278 0 0 1 1.913 0c.066 0 .119.026.171.066L5 2.982 7.916.066A.278.278 0 0 1 8.087 0c.066 0 .119.026.172.066l1.675 1.675c.04.053.066.106.066.172a.278.278 0 0 1-.066.171L7.018 5l2.916 2.916z"
+        fill="#474C5E"
+        fillRule="evenodd"
+      />
+    </svg>
+  );
+}
--- a/src/CloseIcon/CloseIcon.jsx
+++ b/src/CloseIcon/CloseIcon.jsx
@@ -15,8 +15,9 @@ export function CloseIcon(props) {
   const {className} = props;
   return (
     <svg
-      focusable="false"
       className={classnames(cssClass.CONTAINER, className)}
+      // Prevent IE11 from focusing on the element.
+      focusable="false"
       viewBox="0 0 10 10"
       xmlns="http://www.w3.org/2000/svg"
     >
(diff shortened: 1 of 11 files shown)


src/Switch/Switch.tsx
```diff
@@ -0,0 +1,147 @@
...
+        bgStyle = {
+          ...bgStyle,
+          border: "2px solid #474C5E",
+          borderRadius: "4px",
+        };
+      }
+    }
+
+    return (
+      <div className={classnames(cssClass.CONTAINER, className)}>
```
kofi-clever: some structural feedback:
we can think of the switch as a toggle button, essentially, so i think we should make the container component a `<button>` instead.

that way, we don't have to have the custom mouseenter, mouseleave, etc handling and can use the regular button `onClick` listener.

it also increases the clickable/tappable area of the switch.

and then it provides a11y benefits, because:
- it automatically has the ability to receive keyboard focus
- we get all the button click, focus, hover, disabled semantics for free (and a11y tools can interact with it in a standard way)

src/Switch/Switch.tsx
```diff
@@ -0,0 +1,147 @@
...
+          onMouseDown={() => this._onMouseDown()}
+          onMouseUp={() => this._onMouseUp()}
+        >
+          <FlexBox justify={Justify.BETWEEN} grow>
+            <FlexBox column justify={Justify.CENTER}><Checkmark className={cssClass.CHECKED_ICON} /></FlexBox>
+            <FlexBox column justify={Justify.CENTER}><CloseIcon className={cssClass.UNCHECKED_ICON} /></FlexBox>
+          </FlexBox>
+        </FlexBox>
+        <div
+          aria-checked={checked}
```
kofi-clever: these `aria` tags should also go on the top-level element, since that will be the clickable element

src/Switch/Switch.less
```diff
@@ -0,0 +1,133 @@
...
+  -moz-transition: background-color 0.25s, transform 0.25s, box-shadow 0.15s;
+  transition: background-color 0.25s, transform 0.25s, box-shadow 0.15s;
+}
+
+.Switch--handle.Switch--checked {
+  transform: translateX(2em);
+}
+
+.Switch--handle.Switch--disabled {
+  cursor: default;
```
kofi-clever: we usually use the `not-allowed` option for disabled controls, so that it's more obvious

src/Switch/Switch.less
```diff
@@ -0,0 +1,133 @@
...
+  background-color: @neutral_white;
+  cursor: pointer;
+  display: inline-block;
+  border-radius: 0.125em;
+  position: absolute;
+  top: 0.3em;
+  outline: 0;
+  border: 0;
+  -webkit-transition: background-color 0.25s, transform 0.25s, box-shadow 0.15s;
+  -moz-transition: background-color 0.25s, transform 0.25s, box-shadow 0.15s;
```
kofi-clever: you can leave out these -webkit, -moz versions, since all the browsers we officially support can handle the `transition` property: https://caniuse.com/#feat=css-transitions

src/Switch/Switch.less
```diff
@@ -0,0 +1,133 @@
...
+  transition: background-color 0.25s, transform 0.25s, box-shadow 0.15s;
```
kofi-clever: fyi, we have pre-defined transition speeds in `css/animations.less` - check to see if any of those are close to what you need here (and we can always add more if we need them, but better to standardise)

src/CloseIcon/CloseIcon.jsx
```diff
@@ -0,0 +1,30 @@
...
+
+CloseIcon.propTypes = {
+  className: PropTypes.string,
+};
+
+export function CloseIcon(props) {
+  const {className} = props;
+  return (
+    <svg
+      focusable="false"
```
kofi-clever: can you add comments to these to explain why they're needed?

src/Switch/Switch.less
```diff
@@ -0,0 +1,133 @@
+@import (reference) "../less/utilities";
+
+@grey: #71738A;
+@grey-light: #D6D7DE;
+@grey-dark: #474C5E;
+@green: #49D2A2;
+@green-dark: #018657;
```
kofi-clever: we also have pre-defined colour variables in `css/colors.less` that we've standardised on: https://clever.github.io/components/#/design/colors
instead of these variables, you can use those directly and pick ones that are close to what you need

teacamperneil:
Sooooo Gooood!

PR #341, opened 2018-12-19
Switch: display focus on click

aastein:
**Jira:**

**Overview:**
Display switch focus when focus is trigged by clicking.

**Screenshots/GIFs:**
screen recording 2018-12-18 at 04 13 pm

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Switch/Switch.tsx
+++ b/src/Switch/Switch.tsx
@@ -49,12 +49,6 @@ export default class Switch extends React.PureComponent {
   static cssClass = cssClass;
   static defaultProps = defaultProps;
 
-  state = {
-    hasOutline: false,
-    hover: false,
-    mouseDown: false,
-  };
-
   render() {
     const {
       ariaLabelledby,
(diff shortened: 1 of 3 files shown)


src/Switch/Switch.tsx
```diff
@@ -64,32 +58,22 @@ export default class Switch extends React.PureComponent {
...
         aria-label={ariaLabel}
-        className={classnames(cssClass.CONTAINER, className)}
-        onMouseDown={() => this._onMouseDown()}
-        onMouseUp={() => this._onMouseUp()}
-        onFocus={this._onFocus}
-        onBlur={this._onBlur}
+        className={classnames(cssClass.CONTAINER, className, "flexbox", cssClass.BG,
+          checked && cssClass.CHECKED,
+          disabled && cssClass.DISABLED,
+        )}
```
kofi-clever: also `disabled={disabled}` so that we get native disabled state treatment (like taking it out of the tab order and disabling events)

src/Switch/Switch.tsx
```diff
@@ -64,32 +58,22 @@ export default class Switch extends React.PureComponent {
...
         role="checkbox"
```
kofi-clever: nit: we can switch to using the `switch` role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Switch_role

src/Switch/Switch.tsx
kofi-clever: then you could potentially use the `.Switch[disabled]` selector instead of manually managing the disabled class

src/Switch/Switch.tsx
```diff
@@ -64,32 +58,22 @@ export default class Switch extends React.PureComponent {
...
         aria-checked={checked}
         aria-disabled={disabled}
         aria-labelledby={ariaLabelledby}
         aria-label={ariaLabel}
-        className={classnames(cssClass.CONTAINER, className)}
-        onMouseDown={() => this._onMouseDown()}
-        onMouseUp={() => this._onMouseUp()}
-        onFocus={this._onFocus}
-        onBlur={this._onBlur}
+        className={classnames(cssClass.CONTAINER, className, "flexbox", cssClass.BG,
```
kofi-clever: recommend using `.flexbox` as a mixin in the less file instead of adding it here directly, since this assumes that the flex styles are being fully imported (not just as reference) and isn't a pattern we should encourage

PR #343, opened 2018-12-22
Stepper component

vynmeister:
(Review by commit!)

**Jira:**
https://clever.atlassian.net/browse/SYNC-949

**Spec:**
https://projects.invisionapp.com/share/6YPCS27X2JP#/screens/337902518_stepper_In_Dewey

**Overview:**
Adding a Stepper component as part of the effort to abstract logic out of the Wizard.

**To do:**
- testing
- make steps tabbable
- (in the future) horizontal stepper
- (in the future) responsiveness

**Screenshots/GIFs:**
stepper_hover
stepper_focus

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


assets/images/check.svg
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
prime-time: We could use react components here instead of SVGs. I would be interested in hearing Kofi's thoughts on this, and if it would be worth doing.

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+            We ran into an error on this step
+            </span>
+          )}
+          {showMessage && optional && (
+            <span className={cssClass.MESSAGE_OPTIONAL}>
+            Optional
+            </span>
+          )}
+          {showDescription && (
+          _.isString(description)
```
prime-time: Since this is the only use of lodash, we could use `typeof description === "string"` to avoid importing it. It's a really big library so we should try to use builtins when possible.

src/Stepper/Stepper.less
```diff
@@ -0,0 +1,25 @@
...
+  .padding--l;
+}
+
+.Stepper--stepsDisplay {
+  list-style-type: none;
+  .padding--none;
+
+  li {
+    &:not(:last-child)::after {
+      display: block;
```
prime-time: I recently discovered that there is a `display--block` mixin in clever-components
https://github.com/Clever/components/blob/bb7ea958db87b896b6ec1a225dbb0cc2c4caa347/src/less/display.less#L3

It doesn't save any code, so I don't see an advantage to changing this.

prime-time:
I would like to advocate for adding some Kofi magic by including css transitions to make hovering over each step feel buttery smooth.

Here are a few examples

https://github.com/Clever/components/blob/bb7ea958db87b896b6ec1a225dbb0cc2c4caa347/src/TabBar/Tab.less#L19

https://github.com/Clever/components/blob/fc710b1818b1bd547fde399400490bf414291455/src/RadioGroup/Radio.less#L6

Search bar background color transition in apps-dashboard:
  `transition: background-color @timingQuickly ease-out, box-shadow @timingPromptly ease-out;`

teacamperneil:
Awesome work, everyone. Looking great!!

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,88 @@
...
+        </ul>
+      </div>
+    );
+  }
+
+  render() {
+    const {className, style, stickySidebar} = this.props;
+
+    return (
+      <StickyContainer className={classnames(cssClass.CONTAINER, className)} style={style}>
```
kofi-clever: i think this component is doing too much. the stepper should probably just render the steps and decide whether it's rendering vertically, or horizontally and handle size changes.

i'd tying in a sidebar and stickiness belongs in a higher-order component that wraps this. e.g. imagine a generic sidebar container component that can be given any content and told whether or not to be sticky, etc.

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+import "./Step.less";
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+  showTitle: PropTypes.bool,
+  showMessage: PropTypes.bool,
+  showDescription: PropTypes.bool,
+  index: PropTypes.number,
```
kofi-clever: let's not make this an index - recommend calling it `id` instead and making it a string - that way, users aren't limited to numbers and i could, for example, use semantic IDs like `"intro"`, `"emailSettings"`, etc

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+          )}
+        </div>
+      </div>
+    );
+    return (
+      seekable
+        ? <Button
+          className={cssClass.BUTTON}
+          type="linkPlain"
+          onClick={() => onClick(index)}
```
kofi-clever: this will create a new function every time this `Step` component is rendered - instead you can just assign `onClick`
```tsx
class Step {
  render() {
    return (
      <button className={...} onClick={this._onClick}>
        {content}
      </button>
    );
    onClick={onClick
  }

  _onClick = () => this.props.onClick(this.props.id);
}
```

src/Stepper/Step.less
```diff
@@ -0,0 +1,205 @@
...
+  }
+}
+
+.Step:hover.Step--seekable.Step--current:not(:active) {
+  .Step--icon {
+    background-color: @primary_blue_shade_2;
+    border: @iconBorderWidth solid @primary_blue_shade_2;
+    color: @neutral_white;
+  }
+}
```
kofi-clever: the Step css looks a little complicated - are there ways we can simplify it?
a couple things to call out:
- i see a lot of `:not` selectors. in css, that's usually a sign that you could probably re-arrange your rules a bit to take advantage of the cascading nature of rule precedence
- i think i see some selectors that have been assigned either the same or very similar rules - helps to think about what's common and can be grouped into a single class selector higher up and then the rules that vary the ui can appear later and be much simpler because they only specify the rules that are different.

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+    const stepNumber = index + 1;
+
+    const listValue = (
+      <div className={stepClassName}>
+        <div className={cssClass.ICON}>
+          {stepNumber.toString()}
+        </div>
+        <div>
+          {showTitle && (
+            <span className={cssClass.TITLE}>
```
kofi-clever: are these meant to be `<span>`s? that would make then appear inline with each other.
fwiw, if you're setting them to `display: block` after the fact, that's a sign that you should probably just go with a `div` or similar block element

src/Stepper/Step.less
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+  .Step--title, .Step--description, .Step--message--optional {
+    color: @neutral_dark_gray;
+  }
+}
+
+.Step--success:not(.Step--current) {
+  .Step--icon {
+    background-image: url("../../assets/images/check.svg");
```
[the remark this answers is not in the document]
prime-time: 👍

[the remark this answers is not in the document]
kofi-clever: it's possible, but if you need the content to change on mouse interaction (from the mocks, it looks like we only need that for the `:active` state) then it's simpler to just treat the svg as an overlay and fade it out on `:active` (basically absolutely position it on top of the number text).

IMO, we should just push back on revealing the number on mouse down, since it doesn't add much, if anything at all. supposedly, if you click the step button, you're moving to that step anyway and you'll see the number, so i can't see much motivation for also revealing on mouse down.

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,112 @@
...
+    );
+    return (
+      seekable
+        ? <Button
+          className={cssClass.BUTTON}
+          type="linkPlain"
+          onClick={() => onClick(index)}
+          value={listValue}
+        />
+        : listValue
```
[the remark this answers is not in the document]
vynmeister: It's because the Wizard had it and I copied the name 🙈

src/Stepper/Stepper.less
```diff
@@ -0,0 +1,25 @@
...
+      content: "";
```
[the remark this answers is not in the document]
vynmeister: I think this is another copy 🍝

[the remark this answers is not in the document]
aastein: @kofi-clever are you referring to line 21 or the the ` &:not(:last-child)::after {` rule?

src/Stepper/Step.tsx
aastein: the step number (index) is what is rendered in the step icon. Long semantics ID's could not be used, but possibly steps like "A", "B", "C"?

src/Stepper/Step.less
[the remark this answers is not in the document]
kofi-clever: i think we might be miscommunicating - just to be sure, i was saying that we *shouldn't* bother with revealing the number on mouse down (which is the way is spec'd out in https://projects.invisionapp.com/share/6YPCS27X2JP#/screens/337902518_stepper_In_Dewey)

instead, just reveal the number when the page for that step is currently visible.

is that in line with what you were asking?

src/Stepper/Stepper.less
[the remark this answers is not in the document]
kofi-clever: basically the whole ruleset, since it seems to exist just to add the empty string

src/Stepper/Step.tsx
kofi-clever: ohh, i think those should be separate, conceptually and semantically.

i think baking the notion of an "index" into the `Step` might be a bit much, since a `Step` doesn't really care or need to know where it falls in the order. it's more of a `label` than an `index` (and i agree, i could imagine a case where we label them `"A"`, `"B"`, `"C"` instead. the `Stepper`, on the other hand, can take care of ordering/orchestration and deciding what label is rendered in each `Step`

i think the `id` on the other hand should be required and should identify each step uniquely (and that's what i would include instead as the argument of the `onClick()`)

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+
+.Step--description {
+  color: @neutral_gray;
+  .text--medium;
+  .text--regular;
+  transition: color @timingQuickly ease-out;
+}
+
+.Step--message--optional {
+  overflow: hidden;
```
kofi-clever: why does the message need an `overflow: hidden`? just making sure we cut down on the noise with unnecessary rules

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+  font-size: @titleFontSize;
+  .text--semi-bold;
+  .margin--top--2xs;
+  -ms-flex: 0 1 auto;
+  transition: color @timingQuickly ease-out;
+}
+
+.Step--description {
+  color: @neutral_gray;
+  .text--medium;
```
kofi-clever: this is repeated a lot through this file. if it's common to most of the component, can move up to `.Step`. then it can be overridden wherever that's necessary (same for `.text--regular` below)

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+  .justifyContent(center);
+  .circle(@stepIconWidth);
+  background-color: @neutral_white;
+  border: @iconBorderWidth solid @neutral_gray;
+  .text--medium;
+  .text--regular;
+  .margin--right--s;
+  color: @neutral_gray;
+  box-shadow: 0 0 0 @iconFocusShadowThickness transparent;
+  transition: background-color @timingQuickly ease-out, border-color @timingQuickly ease-out;
```
kofi-clever: lot simpler and more readable to go with `all @timingQuickly ease-out` to avoid listing properties out.

that's usually the go-to unless we explicitly need to animate only specific properties. otherwise, someone could add a property change that doesn't get animated and looks weird in relation to the one that's animating.

https://github.com/Clever/components/search?q=transition%3A+all&unscoped_q=transition%3A+all

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+}
+
+.Step--success {
+  .Step--icon {
+    background-color: @neutral_dark_gray;
+    border: @iconBorderWidth solid @neutral_dark_gray;
```
kofi-clever: nit: as mentioned above, clean up using standard mixins (and sorry those aren't more visible)
```less
.Step--icon {
  .border--s(@neutral_gray);
}
```

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+    color: @neutral_dark_gray;
+  }
+
+  .Step--description, .Step--message--optional {
+    color: @neutral_dark_gray;
+  }
+}
+
+.Step:disabled {
+  cursor: default;
```
kofi-clever: `cursor: not-allowed` to match our treatment of other buttons throughout Dewey

src/Stepper/Step.less
```diff
@@ -0,0 +1,174 @@
...
+}
+
+.Step:focus:not(.Step--current), .Step:hover.Step--seekable:not(.Step--current) {
+  .Step--title {
+    color: @primary_blue;
+  }
+}
+
+.Stepper--CheckMark {
+  width: .75em;
```
kofi-clever: `width: @size_s;` - same below

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+import Exclamation from "./Exclamation";
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.oneOfType([PropTypes.node]),
+  showTitle: PropTypes.bool,
+  showMessage: PropTypes.bool,
+  showDescription: PropTypes.bool,
+  id: PropTypes.number,
```
kofi-clever: as mentioned earlier, 
- `id: string`
- `label: string`
- `label` gets rendered, `id` is sent with `onClick` event
- `Stepper` takes care of any ordering concerns based on the list of steps it's given

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+      cssClass.CONTAINER,
+      cssClass.BUTTON,
+      success && cssClass.SUCCESS,
+      current && cssClass.CURRENT,
+      warning && cssClass.WARNING,
+      seekable && cssClass.SEEKABLE,
+    );
+
+    return (
+      <button
```
kofi-clever: we shouldn't be rendering this as a button if it's not seekable. out of sync, semantically, and incorrect from an a11y perspective.

pattern we've been using elsewhere is: 
```tsx
const content = (
  <div>
    <some />
    <content />
  </div>
);

if (!onClick) {
  return (
    <div className={stepClassName}>
      {content}
    </div>
  );
}

return (
  <button className={stepClassName} foo={bar} onClick={this._onClick)>
    {content}
  </button>
);
```

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.oneOfType([PropTypes.node]),
+  showTitle: PropTypes.bool,
+  showMessage: PropTypes.bool,
+  showDescription: PropTypes.bool,
+  id: PropTypes.number,
+  seekable: PropTypes.bool,
```
kofi-clever: up to you, based on preference, but this prop is essentially redundant, since we can rely on the presence of an onClick to decide whether or not we render a working button.

we don't want to render a button with no onClick handler anyway, so better to just use an `onClick` check as the deciding factor

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import "./Step.less";
+import CheckMark from "./CheckMark";
+import {FlexBox} from "../flex";
+import Exclamation from "./Exclamation";
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.oneOfType([PropTypes.node]),
```
kofi-clever: `description: PropTypes.node,`

"Drop the `PropTypes.oneOfType`. It's cleaner." - Sean Parker

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+  showTitle: PropTypes.bool,
+  showMessage: PropTypes.bool,
+  showDescription: PropTypes.bool,
```
kofi-clever: missed from last feedback round - let's drop the `show*` props and use the presence of the props to decide what's rendered. these just bloat the API with no added benefit

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+  id: PropTypes.number,
+  seekable: PropTypes.bool,
+  success: PropTypes.bool,
+  current: PropTypes.bool,
+  warning: PropTypes.bool,
```
kofi-clever: should `warning: true` and `success: true` be mutually exclusive?
i'm a little confused by the API on first glance, which is a sign it could be tightened up a bit.

from reading the props, i *think* `warning` enables warning styling (colour change, icon change, etc) and `success` does similar for a successful state. if that's the case, we can go with the pattern we've been using elsewhere and provide an enum of state values - `INCOMPLETE`, `SUCCESS`, `WARNING` - and require that the client specify one of those states

related: we're using this to decide whether or not the `message` is rendered as warning text - that's conflating 2 concepts. i'd recommend adding a `warning: PropTypes.string` instead along with `message: PropTypes.string` and displaying either/both depending on which is specified (unless `message` is always a warning, in which case, might as well leave out `message`)

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+  static propTypes = propTypes;
+  static defaultProps = defaultProps;
+  static cssClass = cssClass;
+
+  _onClick = () => this.props.onClick(this.props.index);
+
+  _iconContent = () => {
+    const {index, success, current, warning} = this.props;
+    const stepNumber = index + 1;
+    if (current) return stepNumber.toString();
```
kofi-clever: let's stick to expanded if clause formatting. clearer and more consistent.

also, if we switch to a state enum (see prop comments above), we can just have a mapping of state to icon content and use that if `!current`

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <FlexBox className={cssClass.ICON}>{this._iconContent()}</FlexBox>
+        <div>
+          {title && (
+            <div className={cssClass.TITLE}>
+              {title}
+            </div>
+          )}
+          {warning && (
+            <div className={cssClass.MESSAGE_WARNING}>
+            We ran into an error on this step
```
kofi-clever: i'd recommend giving clients control over what the error message is. it's fine if we have a default for the prop, but the current implementation is fairly inflexible.

also, looks like we're not using the `showMessage` and `message` props at all

src/Stepper/Stepper.less
```diff
@@ -0,0 +1,24 @@
...
+  max-width: 18rem;
+  .padding--l;
+}
+
+.Stepper--stepsDisplay {
+  list-style-type: none;
+  .padding--none;
+
+  li {
+    &:not(:last-child)::after {
```
kofi-clever: can we do without the `::after` selector? in almost all cases, it's unnecessary

src/Stepper/Stepper.less
```diff
@@ -0,0 +1,24 @@
+@import "../less/fonts.less";
+@import (reference) "../less/index";
+
+.Stepper {
+  .flexbox();
+}
+
+.Stepper--sidebar {
```
kofi-clever: related to last review - move sidebar stuff into a higher-order component if needed.

ideally, we'd just have some generic sticky sidebar container component that's very simple and just renders the content it's given and handles positioning with a `maxWidth` prop

src/Stepper/Stepper.less
```diff
@@ -0,0 +1,24 @@
...
+  .flex(0, 1, 18rem);
+  max-width: 18rem;
+  .padding--l;
+}
+
+.Stepper--stepsDisplay {
+  list-style-type: none;
```
kofi-clever: let's switch to the `.list--reset()` mixin for consistency:
e.g. https://github.com/Clever/components/search?q=list-reset&unscoped_q=list-reset

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,82 @@
...
+  static propTypes = propTypes;
+  static defaultProps = defaultProps;
+  static cssClass = cssClass;
+
+  _renderSidebar(style) {
+    const {currentStep, steps, seekable, onStepClick, className} = this.props;
+
+    return (
+      <div style={style}>
+        <ul className={cssClass.STEPS_DISPLAY}>
```
kofi-clever: `<div>` wrapper here is redundant - the `<div><ul></ul></div>` can be collapsed into a single `<ul>`

src/Stepper/Stepper.tsx
```diff
@@ -0,0 +1,82 @@
...
+import Step from "./Step";
+
+import "./Stepper.less";
+
+const propTypes = {
+  className: PropTypes.string,
+  style: PropTypes.object,
+  steps: PropTypes.arrayOf(PropTypes.shape({
+    title: PropTypes.string.isRequired,
+    description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
```
kofi-clever: `description: PropTypes.node,`
(since PropTypes.node is a superset of PropTypes.string)

src/Stepper/Step.less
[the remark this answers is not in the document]
aastein: yes, thanks

src/Stepper/Step.less
```diff
@@ -117,42 +108,48 @@
   }
 }
 
-.Step:disabled {
-  cursor: default;
+.Step:not(.Step--current):not(.Step--seekable) {
+  cursor: not-allowed;
 }
 
 .Step:focus.Step--warning {
```
kofi-clever: lower specificity: `.Step--warning:focus`

src/Stepper/Step.less
```diff
@@ -117,42 +108,48 @@
...
+  }
+}
+
+.Step.Step--current.Step--warning {
+  .Step--icon {
+    background-color: @primary_blue;
   }
 }
 
 .Step:focus.Step--current {
```
kofi-clever: `.Step--current:focus`

src/Stepper/Step.less
```diff
@@ -117,42 +108,48 @@
...
   .Step--icon {
-    border: @iconBorderWidth solid @primary_blue;
+    .border--s(@primary_blue);
     color: @neutral_white;
   }
 }
 
 .Step:hover.Step--seekable.Step--success {
```
kofi-clever: `.Step--seekable.Step--success` - similar thing for the others

src/Stepper/Step.tsx
```diff
@@ -49,49 +45,47 @@ const defaultProps = {
...
+    const icon = (<FlexBox className={cssClass.ICON}>{this._iconContent()}</FlexBox>);
+    const content = (
+      <div>
           {title && (
             <div className={cssClass.TITLE}>
               {title}
             </div>
           )}
-          {warning && (
+          {state === "WARNING" && (
```
kofi-clever: why not just render the warning message if one is specified?

src/Stepper/Step.tsx
```diff
@@ -38,9 +35,8 @@ const cssClass = {
 };
 
 const defaultProps = {
-  showTitle: true,
-  showMessage: true,
-  showDescription: true,
+  state: "INCOMPLETE",
+  warning: "We ran into an error on this step",
```
kofi-clever: i think we should just leave out this default. we should encourage clients to have more useful error messages than this

src/Stepper/Step.less
```diff
@@ -2,16 +2,16 @@
...
 .Step {
   .flexbox();
   .alignItems(flex-start);
-  width: 100%;
   .padding--y--2xs;
+  .text--medium;
+  .text--regular;
 }
 
 .Step--button {
```
kofi-clever: can this be consolidated with `.Step`? couldn't tell why they're separate.

src/Stepper/Step.tsx
```diff
@@ -105,7 +99,25 @@ export default class Step extends React.PureComponent {
...
+    if (onClick) {
+      return (
+      <button
+        tabIndex={0}
+        className={classnames(stepClassName, cssClass.BUTTON)}
+        onClick={this._onClick}
+      >
+        {icon}
+        {content}
+      </button>);
```
kofi-clever: nit: please run a formatter on this - formatting's little off.

src/Stepper/Step.tsx
```diff
@@ -105,7 +99,25 @@ export default class Step extends React.PureComponent {
             </div>
           )}
         </div>
-      </button>
+    );
+
+    if (onClick) {
+      return (
+      <button
+        tabIndex={0}
```
kofi-clever: this is already the default, so no need to specify

src/Stepper/Stepper.tsx
```diff
@@ -12,19 +12,18 @@ const propTypes = {
   style: PropTypes.object,
   steps: PropTypes.arrayOf(PropTypes.shape({
     title: PropTypes.string.isRequired,
-    description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    description: PropTypes.node,
```
kofi-clever: this is missing `id` and `state` (and `label`, if you want to support that)

src/Stepper/Stepper.tsx
```diff
@@ -12,19 +12,18 @@ const propTypes = {
...
     success: PropTypes.bool,
-    warning: PropTypes.bool,
+    warning: PropTypes.string,
     optional: PropTypes.bool,
   })).isRequired,
-  currentStep: PropTypes.number,
-  seekable: PropTypes.bool,
+  currentStep: PropTypes.string,
```
kofi-clever: opt: `currentStepID` to be clearer?

src/Stepper/Stepper.tsx
```diff
@@ -39,29 +38,27 @@ export default class Stepper extends React.PureComponent {
...
-        </ul>
-      </div>
+      <ul style={style} className={cssClass.STEPS_DISPLAY}>
+        {steps.map(step =>
+          <li key={step.id}>
+            <Step
+              className={className}
+              title={step.title}
+              description={step.description}
+              label={step.label}
```
kofi-clever: this can just be the string version of the step's index + 1 (or if you want the labels to be customisable, let's at least default to `index + 1` if the client hasn't specified a label)

src/Stepper/Stepper.tsx
```diff
@@ -12,19 +12,18 @@ const propTypes = {
...
-  currentStep: PropTypes.number,
-  seekable: PropTypes.bool,
+  currentStep: PropTypes.string,
   stickySidebar: PropTypes.bool,
   onStepClick: PropTypes.func,
 };
 
 const defaultProps = {
-  currentStep: 0,
+  currentStep: "0",
```
kofi-clever: not going to work if a client is using semantic IDs - you can pick `steps[0].id` as the default in the `render` method if no `currentStepID` is specified.
better yet, just make `currentStepID` required, so things are clear and explicit. the client inevitably has to track step changes anyway, so we might as well require the initial to be specified to ensure we're always in sync with the client.

src/Stepper/Stepper.tsx
```diff
@@ -12,19 +12,18 @@ const propTypes = {
   style: PropTypes.object,
   steps: PropTypes.arrayOf(PropTypes.shape({
     title: PropTypes.string.isRequired,
-    description: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    description: PropTypes.node,
     success: PropTypes.bool,
```
kofi-clever: rm

src/Stepper/Stepper.tsx
```diff
@@ -37,8 +33,8 @@ export default class Stepper extends React.PureComponent {
   static defaultProps = defaultProps;
   static cssClass = cssClass;
 
-  _renderSidebar(style) {
-    const {currentStep, steps, onStepClick, className} = this.props;
+  render() {
+    const {currentStep, style, steps, onStepClick, className} = this.props;
```
kofi-clever: supporting `style` was only necessary before, when we expected `Sticky` to manage the component's style. can remove now to encourage people to keep styling in LESS for now

src/Stepper/Step.less
aastein: not sure what you mean by consolidated.

aastein: not totally. .button--reset() cannot be used when Step is a <div> because it changes the pointer.

src/Stepper/Step.less
```diff
@@ -0,0 +1,171 @@
...
+  .padding--y--2xs;
+  .text--medium;
+  .text--regular;
+  .padding--none;
+  text-align: left;
+  .text--line-height-3;
+}
+
+button.Step {
+  .button--reset();
```
kofi-clever: this will override any intersecting styles in `.Step` above - i believe one way to solve this and make sure the button and non-button styling stay in sync is to add a reference to `.Step` as a mixin:

```less
button.Step {
  .button--reset();
  .Step();
}
```

src/Stepper/Step.less
```diff
@@ -0,0 +1,171 @@
...
+
+@stepIconWidth: 1.75rem; // 28px
+@iconFocusShadowThickness: @size_2xs; // 4px
+@titleFontSize: 1.25rem; // 20px
+@alertSize: 0.375rem; // 6px
+
+.Step {
+  .flexbox();
+  .alignItems(flex-start);
+  .padding--y--2xs;
```
kofi-clever: just fyi; no need to change this right now, since we need to do a mass change in this repo anyway, but this syntax for mixins is getting deprecated, IIRC.
need to start always including the parentheses now: `.padding--y--2xs()`, etc

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,119 @@
...
+import CheckMark from "./CheckMark";
+import {FlexBox} from "../flex";
+import Exclamation from "./Exclamation";
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.node,
+  id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  label: PropTypes.number,
```
kofi-clever: update `label` to `PropTypes.node`, since it's not limited to just numbers

src/Stepper/Step.tsx
```diff
@@ -0,0 +1,119 @@
...
+import "./Step.less";
+import CheckMark from "./CheckMark";
+import {FlexBox} from "../flex";
+import Exclamation from "./Exclamation";
+
+const propTypes = {
+  className: PropTypes.string,
+  title: PropTypes.string,
+  description: PropTypes.node,
+  id: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
```
kofi-clever: 1 - we can keep this simple and just make it a string (i wouldn't encourage numeric IDs anyway)
2 - make this prop required

src/Stepper/Stepper.tsx
kofi-clever: ^still relevant

docs/components/StepperView.jsx
```diff
@@ -113,13 +126,14 @@ export default class StepperView extends PureComponent {
...
               <Row grow>
                 <Col span={4}>
-                <Stepper
-                  className="ExampleStepper"
-                  currentStepID={currentStepID}
-                  steps={steps}
-                  sticky={sticky}
-                  onStepClick={seekable ? (id) => this.jumpToStep(id) : null }
-                />
+                <StickyContainer className={classnames(cssClass.STICKY_CONTAINER)}>
```
kofi-clever: we have Dewey set up so that whatever appears within the `<ExampleCode>` tags gets output verbatim as the code sample that people will refer to.

the way this is currently written, we're going to end up with only the `<Grid>` and `<StickyContainer>` stuff showing, but not the stepper itself.

let's clean this up a bit:
- no need to include the sticky container - doesn't add value to the demo code at this point
- move the `{this._renderConfig()}` part outside of the `<ExampleCode>` element
- [opt] do away with `<Grid>` and `<Row>` - the Grid component can be very noisy and, in most cases, we can accomplish the same thing with regular divs. in this case, would also be fine just to use the `<Col>` elements on their own and drop the excess Grid/Row code

src/Stepper/Stepper.less
```diff
@@ -1,18 +1,10 @@
...
-
-.Stepper--stickyContent {
-  .flex(0, 1, 18rem);
-  max-width: 18rem;
-}
-
 .Stepper--stepsDisplay {
   .list-reset;
   .padding--l;
+  max-width: 18rem;
```
kofi-clever: i think we should leave this out for now - this is something that can be determined by the client that's rendering the stepper

src/Stepper/Stepper.tsx
```diff
@@ -23,55 +20,38 @@ const propTypes = {
   })).isRequired,
   currentStepID: PropTypes.string.isRequired,
   onStepClick: PropTypes.func,
-  sticky: PropTypes.bool,
+  style: PropTypes.object,
```
kofi-clever: this prop was only needed to support the sticky container, so can drop it now - we want to encourage styling to happen via LESS at this point

PR #347, opened 2019-01-07
Add animation to progress bar

emilyhou:
**Jira:**
[SYNC-947](https://clever.atlassian.net/browse/SYNC-947)

**Overview:**
Add animation to when the fill of the progress bar changes. These animations are due to adjustments in the future but these are the numbers I found to be ok.

Unfortunately the gif doesn't show the rendering as smoothly as it appears in browser :(

**Screenshots/GIFs:**
kapture 2019-01-07 at 12 24 16

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Backward compatible change? Run `npm version patch` <--
- After merging:

### DIFF
--- a/src/ProgressBar/Progress.less
+++ b/src/ProgressBar/Progress.less
@@ -5,6 +5,8 @@
   height: 100%;
   width: 100%;
   position: absolute;
+  transition: width @timingSlowly ease-out @timingPromptly;
+  -webkit-transition: width @timingSlowly ease-out @timingPromptly;
 }
 
 .ProgressBar--Progress--none {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "0.36.8",
+  "version": "0.36.9",
   "description": "A library of helpful React components and less styles",
   "repository": {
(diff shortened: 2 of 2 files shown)


src/ProgressBar/Progress.less
```diff
@@ -5,6 +5,8 @@
   height: 100%;
   width: 100%;
   position: absolute;
+  transition: width 0.5s ease-out 0.25s;
+  -webkit-transition: width 0.5s ease-out 0.25s;
```
prime-time: Are the transitions too quick with these timing variables?
https://github.com/Clever/components/blob/c257f5b1c5cf7fdfe9e8a5f25a28c305829c6107/src/less/animations.less#L6-L10

emilyhou: Oh cool I didn't realize these were already defined. 0.5s is 500ms so what I had set is actually slower than everything else! I'll probably do @timingSlowly for the transition duration and @timingPrompty for the delay.

PR #348, opened 2019-01-07
Check bad fill values

emilyhou:
**Jira:**
[SYNC-947](https://clever.atlassian.net/browse/SYNC-947)

**Overview:**
Check that updated `fill` prop after first render has valid values.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/ProgressBar/ProgressBar.tsx
+++ b/src/ProgressBar/ProgressBar.tsx
@@ -53,6 +53,12 @@ const cssClass = {
 };
 
 export default class ProgressBar extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {fillPercentage: props.fill};
+  }
+
   static propTypes = propTypes;
   static cssClass = cssClass;
   static defaultProps = {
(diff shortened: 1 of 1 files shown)


src/ProgressBar/ProgressBar.tsx
```diff
@@ -53,6 +53,12 @@ const cssClass = {
 };
 
 export default class ProgressBar extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.state = {fillPercentage: props.fill};
```
kofi-clever: usually not a good idea to mirror props in state and have them go out of sync - seems like it's a client bug if an invalid fill value is passed in so we can either just use a default value (0?) when rendering with an invalid fill (maybe extract the logic to a `_getFillPercentage()` method if that was the intent behind storing in `state`) or not render the progress bar at all.

emilyhou: The worry is that the user entered a correct `fill` prop on initial render of the component, but when updating the component with a new `fill` value they miscalculated or something and ended up entering a NaN or some value that is too small or too big. In this case I would want the progress bar to keep what it has instead of updating the value. If we use a default of 0 then the progress bar would actually go through an animation and display the default which I think could be confusing to the end user.

emilyhou:
Closing as this check shouldn't be done on the component side. Client is responsible for passing in a valid percentage value!

PR #350, opened 2019-01-09
Get that bread! (aka add toast notifications)

arsalansufi:
> Commit by commit reviewing recommended

# JIRA

[IL-928](https://clever.atlassian.net/browse/IL-928)

# Overview

This PR adds toast notifications to Dewey in the form of a `<ToastStack>` component!

### All the toasts

light-to-dark

### Enter/leave animations

animations-higher-res

### Docs

docs

props

# Testing

- [x] Tested all options/props and added them to the docs
- [x] Tested on Chrome
- [x] Tested on Firefox
- [x] Tested on Safari
- [x] Tested on Edge
- [x] Tested on IE

# Rollout

- [x] Bump version
- [x] Merge
- [ ] Deploy updated docs

### DIFF
--- a/docs/components/ToastStackView.jsx
+++ b/docs/components/ToastStackView.jsx
@@ -1,0 +1,267 @@
+import * as React from "react";
+
+import {Button, FlexBox, ItemAlign, SegmentedControl, ToastStack, ToastType} from "src";
+import Example, {CodeSample, ExampleCode} from "./Example";
+import PropDocumentation from "./PropDocumentation";
+import View from "./View";
+
+import "./ToastStackView.less";
+
+const cssClass = {
+  CONFIG_CONTAINER: "ToastStackView--configContainer",
+  CONFIG_OPTIONS: "ToastStackView--configOptions",
+  CONFIG: "ToastStackView--config",
+  CONFIG_TOGGLE: "ToastStackView--configToggle",
+  CONTAINER: "ToastStackView",
+  INTRO: "ToastStackView--intro",
+  PROPS: "ToastStackView--props",
+  TOAST_STACK: "ToastStackView--toastStack",
+};
+
+const messageMap = {
+  [ToastType.ERROR]: "burnt",
+  [ToastType.INFO]: "high in fiber",
+  [ToastType.PROCESSING]: "toasting...",
+  [ToastType.SUCCESS]: "ready!",
+  [ToastType.WARNING]: "slightly burnt",
+};
+
+const actionType = `{
+  text: string,
+  href?: string,
+  onClick?: Function,
+}`;
+
+export default class ToastStackView extends React.PureComponent {
+  static cssClass = cssClass;
+
+  state = {
+    durationOverride: "No override",
+    id: 0,
+    includeAction: false,
+    notifications: [],
+    showCloseButton: true,
+    toastType: ToastType.SUCCESS,
+  };
+
+  render() {
+    const {
+      durationOverride,
+      id,
+      includeAction,
+      notifications,
+      showCloseButton,
+      toastType,
+    } = this.state;
+
+    return (
+      <View
+        className={cssClass.CONTAINER}
+        title="ToastStack"
+        sourcePath="src/ToastStack/ToastStack.tsx"
+      >
+        <header className={cssClass.INTRO}>
+          <p>
+            Toast notifications can be used to provide feedback in response to user interactions.
+            The <code>ToastStack</code> renders toast notifications and handles their enter/leave
+            animations.
+          </p>
+          <CodeSample>{`
+            import {ToastStack, ToastType} from "clever-components";
+          `}</CodeSample>
+        </header>
+
+        <Example title="Basic Usage:">
+          <ExampleCode>
+            <ToastStack
+              className={cssClass.TOAST_STACK}
+              notifications={notifications}
+              setNotifications={ns => this.setState({notifications: ns})}
+              defaultNotificationDurationMS={5000}
+            />
+            <Button
+              value="Create notification"
+              onClick={() => this.setState({
+                id: id + 1,
+                notifications: [...notifications, {
+                  id,
+                  type: toastType,
+                  content: <span>Your toast is {messageMap[toastType]}</span>,
+                  action: includeAction ? {
+                    text: "Butter it",
+                    onClick: () => console.log("Buttering"),
+                  } : undefined,
+                  showCloseButton,
+                  durationMS: durationOverride === "No override" ? undefined : durationOverride,
+                }],
+              })}
+            />
+            {" "}
+            <Button
+              value="Clear all notifications"
+              onClick={() => this.setState({notifications: []})}
+            />
+          </ExampleCode>
+          {this._renderConfig()}
+        </Example>
+
+        {this._renderToastStackProps()}
+        {this._renderNotificationDefinition()}
+      </View>
+    );
+  }
+
+  _renderConfig() {
+    const {
+      durationOverride,
+      includeAction,
+      showCloseButton,
+      toastType,
+    } = this.state;
+
+    return (
+      <FlexBox alignItems={ItemAlign.CENTER} className={cssClass.CONFIG_CONTAINER} wrap>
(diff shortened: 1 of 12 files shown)


src/ToastStack/ToastStack.tsx
```diff
@@ -0,0 +1,111 @@
...
+      return notification.durationMS !== Infinity;
+    }
+    return defaultNotificationDurationMS !== Infinity;
+  }
+
+  _clearNotification(id) {
+    const {notifications, setNotifications} = this.props;
+
+    setNotifications(notifications.filter(n => n.id !== id));
+  }
```
arsalansufi: I decided to have the `<ToastStack>` take in a `setNotifications` prop instead of a `clearNotification` prop for a few reasons:

1. This makes the `<ToastStack>`'s interface similar to other Dewey components. `<TextInput>` for example takes a `value` prop and an `onChange` prop to update `value`.
2. Constructing a `setNotifications` prop as a consumer is easier than constructing a `clearNotification` prop as a consumer. In fact, I initially planned to have this component take a `clearNotification` prop and constructed a buggy one when creating the `<ToastStack>` docs. Happy to describe the bug in person! (It involved outdated object references.)

arsalansufi: ^ Actually, after trying to use this new `<ToastStack>` in an actual Clever repo, I've had a change of heart. Switched back from `setNotifications` to `clearNotification` in https://github.com/Clever/components/pull/350/commits/aec5da33fdc2d738651276b09bef5e57504eae97.

Regarding the reasoning for this:

1. Even with this new component, we'll still need to set up `clearNotification` Redux actions in our repos. Supporting a `setNotifications` API would require us to set up an additional, somewhat redundant Redux action.
2. The bugs that I'm concerned about are avoided in the Redux contexts that our `<ToastStack>`s will be used in.

I still went ahead and added a word of caution to the prop docs

docs/components/ToastStackView.jsx
```diff
@@ -76,7 +82,10 @@ export default class ToastStackView extends React.PureComponent {
             <ToastStack
               className={cssClass.TOAST_STACK}
               notifications={notifications}
-              setNotifications={ns => this.setState({notifications: ns})}
+              clearNotification={
+                // Internally, `this._clearNotification` performs:
+                // this.setState({notifications: notifications.filter(n => n.id !== id)});
```
arsalansufi: Added this comment since this is the example code displayed in the Dewey docs

src/ToastStack/ToastStack.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+  componentDidUpdate(prevProps) {
+    const {defaultNotificationDurationMS} = this.props;
+    const prevNotifications = prevProps.notifications || [];
+    const newNotifications = this.props.notifications || [];
+
+    // If a notification was just added to the stack and it has a finite duration, set up a timer
```
mkmathur: Could multiple new notifications be added to the stack at once?

docs/components/ToastStackView.jsx
```diff
@@ -0,0 +1,267 @@
...
+            <ToastStack
+              className={cssClass.TOAST_STACK}
+              notifications={notifications}
+              setNotifications={ns => this.setState({notifications: ns})}
+              defaultNotificationDurationMS={5000}
+            />
+            <Button
+              value="Create notification"
+              onClick={() => this.setState({
+                id: id + 1,
```
mkmathur: id is a little confusing here. Maybe call it something like notificationsCount or nextNotificationID?

docs/components/ToastStackView.jsx
mkmathur: Instead of adding a comment, it might be more clear if you inline the function.

docs/components/ToastStackView.jsx
```diff
@@ -183,6 +192,13 @@ export default class ToastStackView extends React.PureComponent {
             description: "An additional CSS class name to apply to the container.",
             optional: true,
           },
+          {
+            name: "clearNotification",
+            type: "(id: number) => void",
+            description: "A function that clears the specified notification. Important gotcha: " +
+              "This function may be run with a timer, so make sure that the function doesn't " +
```
mkmathur: I'm not sure if I understand the gotcha / how to avoid it.

src/ToastStack/ToastNotification.tsx
```diff
@@ -0,0 +1,99 @@
...
+import * as PropTypes from "prop-types";
+import * as React from "react";
+
+import {Button} from "../Button/Button";
+import {FlexBox, ItemAlign} from "../flex";
+import {ToastType} from "./ToastType";
+
+import "./ToastNotification.less";
+
+export const actionPropType = PropTypes.shape({
```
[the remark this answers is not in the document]
arsalansufi: Unfortunately not yet :cry: This repo still uses `eslint` and the minimal react typings:

https://github.com/Clever/components/blob/179c6cfee99127ce2c5043b50523c64a065aa587/package.json#L52

both of which don't play well with component types

src/ToastStack/ToastStack.tsx
arsalansufi: Ooh yes great point. This is technically possible, though in practice unlikely given how we use toast notifications. I'll still update the code to handle this case for completeness

docs/components/ToastStackView.jsx
arsalansufi: 👍
I'll go with `nextNotificationID`

docs/components/ToastStackView.jsx
arsalansufi: So inlining the function will actually reproduce the bug that I've been referring to 😅

```tsx
const { notifications } = this.state;
...
clearNotification={(id) => this.setState({
  notifications: notifications.filter(n => n.id !== id)
})
```

Here, I'm creating four notifications in quick succession, each set to expire after a second. Note the (unexpected and repeated) resurfacing of n - 1 notifications without any user interaction :bug:

bug

The following tweak fixes the bug while still allowing for inlining in the example code. Key difference: Using `this.state.notifications` instead of the earlier extracted `notifications`.

```tsx
const { notifications } = this.state;
...
clearNotification={(id) => this.setState({
  notifications: this.state.notifications.filter(n => n.id !== id)
})
```

So I could use this. But I'm hesitant for a few reasons.

1. Someone looking at the example code might get confused as to why we sometimes use `this.state.notifications` and other times just `notifications`.
2. I could change everything in the example code to access state fields via `this.state.*` instead of a destructuring `const { ... } = this.state`. But that feels fragile. I could see someone switching to destructuring, not realizing the implications.
3. It's good practice to bind function props like this outside `render` for components that you expect to re-render frequently to avoid repeated function allocations. (More details about this [here](https://medium.freecodecamp.org/react-binding-patterns-5-approaches-for-handling-this-92c651b5af56)!) And I figure the example code should try to demonstrate best practices.

Regarding the bug itself, I would type up an explanation for it, but this comment has already gotten to be pretty lengthy 😅 So let's chat in person tomorrow! tl;dr: `notifications` when extracted from `this.state` can go out of date. Whenever we update the `notifications` state value, `this.state.notifications` becomes a new reference.

src/ToastStack/ToastStack.tsx
arsalansufi: Left some initial notes about it (short of an actual explanation) in response to another comment. Let's chat tomorrow!

docs/components/ToastStackView.jsx
arsalansufi: After we sync up tomorrow on the bug itself, I'd love to hear your thoughts on how to concisely guide against it here

arsalansufi: @mkmathur Thanks for you help coming up with a better guideline! Added in https://github.com/Clever/components/pull/350/commits/ff016cf9721e0db529e677b9342e04d0e4de400b

ayy

mkmathur: Perfect!

mkmathur:
LGTM!

PR #352, opened 2019-01-16
SYNC-949: sticky sidebar HOC

aastein:
**Jira:**
https://clever.atlassian.net/browse/SYNC-949

**Overview:**
A sticky sidebar. There's nothing about this that makes it a sidebar. It's just a sticky container HOC. Maybe this should be renamed?

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/Sidebar/Sidebar.tsx
+++ b/src/Sidebar/Sidebar.tsx
@@ -1,0 +1,32 @@
+import * as React from "react";
+import * as classnames from "classnames";
+import * as PropTypes from "prop-types";
+import {Sticky, StickyContainer} from "react-sticky";
+
+import "./Sidebar.less";
+
+const cssClass = {
+  CONTAINER: "Sidebar",
+  SIDEBAR: "Sidebar--sidebar",
+};
+
+
+export default function sidebar(sticky, content) {
+  return class extends React.PureComponent {
+    propTypes = {
+      style: PropTypes.object,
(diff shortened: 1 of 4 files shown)


src/Sidebar/Sidebar.tsx
```diff
@@ -0,0 +1,32 @@
...
+    propTypes = {
+      style: PropTypes.object,
+    }
+    render() {
+      return (
+        <StickyContainer className={classnames(cssClass.CONTAINER)}>
+          <div className={cssClass.SIDEBAR}>
+            {!sticky ? content :
+              <Sticky>
+                {({style}) => <div style={style}>{content}</div>}
```
kofi-clever: yeah, i don't think it's necessary for this to be a Dewey component if this is all it does currently. especially since we have a LeftNav component already and this could be confusing.

can we just do this `Sticky` wrapping where it's needed? it's very little code. looks like this is in support of a secure-sync task?

aastein: I'm just using the same ticket as https://github.com/Clever/components/pull/343. I agree, it doesn't make sense making this its own component.

PR #357, opened 2019-02-04
Set up pre-specified form element sizes

arsalansufi:
# JIRA

https://clever.atlassian.net/browse/IL-1008

# Overview

Throughout our product, we have unbounded form elements. These elements look a bit strange and also aren't user friendly.

oh-my

This PR introduces a `size` prop

props-documentation

to all of the following form elements:

`CopyableInput`
`DateInput`
`FileInput`
`Select`
`TextArea`
`TextInput`

For backwards compatibility, the default `size` is still unbounded. Once we've done some auditing, we can switch the default to be one of the bounded sizes.

## GIFs

copyableinput

dateinput

fileinput

select

textarea

textinput

# Testing

See the above GIFs! Also browser tested.

- [x] Chrome
- [x] Firefox
- [x] Safari
- [x] Edge
- [x] IE

# Rollout

As noted in the Overview, the default size is still unbounded. So Dewey consumers won't experience any automatic changes.

### DIFF
--- a/docs/components/CopyableInputView.jsx
+++ b/docs/components/CopyableInputView.jsx
@@ -106,7 +106,7 @@ export default class CopyableInputView extends Component {
                 {content: "small", value: FormElementSize.SMALL},
                 {content: "medium", value: FormElementSize.MEDIUM},
                 {content: "large", value: FormElementSize.LARGE},
-                {content: "unbounded", value: FormElementSize.UNBOUNDED},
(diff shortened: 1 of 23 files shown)


jish:
How attached are you to the word "unbounded"? Do you think we could come up with something else like "full-width", "stretch", "100 percent", other? 🤔

arsalansufi:
Ooh not attached at all, I like full-width / `FormElementSize.FULL_WIDTH`! Will update

PR #362, opened 2019-03-07
SYNC-1098: make save & exit button optional

aastein:
**Jira:**
https://clever.atlassian.net/browse/SYNC-1098

**Overview:**
For content mapping the save & exit button should be hidden. This PR makes its rendering conditional based on if the onSaveAndExit prop is provided to the component.

**Screenshots/GIFs:**
image

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/WizardLayout/WizardLayout.tsx
+++ b/src/WizardLayout/WizardLayout.tsx
@@ -24,7 +24,7 @@ const propTypes = {
   nextStepButtonText: PropTypes.string,
   onNextStep: PropTypes.func.isRequired,
   onPrevStep: PropTypes.func.isRequired,
-  onSaveAndExit: PropTypes.func.isRequired,
+  onSaveAndExit: PropTypes.func,
   prevStepButtonDisabled: PropTypes.bool,
   prevStepButtonText: PropTypes.string,
   stepper: PropTypes.node.isRequired,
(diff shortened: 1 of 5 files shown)


src/WizardLayout/WizardLayout.tsx
```diff
@@ -110,11 +110,11 @@ export default class WizardLayout extends React.PureComponent {
           </FlexBox>
         </FlexBox>
         <FlexBox className={cssClass.FOOTER}>
-          <Button
+          { onSaveAndExit && <Button
```
brollin: Nit: This is subtle, and I see that this file has done it inconsistently, but there should be no spacing inside JSX brackets, and there should be inside of JS object brackets.

So in this example it would look like:
```jsx
{onSaveAndExit && <Button ... />}
```

Sometimes you get object brackets inside of JSX brackets which is when you get the funny thing like:
```jsx
<Select
  value={{ label: "label", value: "value" }}
  ...
/>
```

Just a heads up so you can be the change whenever you see this done incorrectly around our codebases!

PR #365, opened 2019-03-28
add class to next button

aastein:
**Jira:**

**Overview:**
For non-impersonators pendo renders a Clever logo at the bottom right of all pages. This PR adds a css class to the next button so that clients can move the bottom button group out of the way of the pendo clever logo.

**Screenshots/GIFs:**
image

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "1.7.2",
+  "version": "1.8.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


johnhuangclever:
I think we would prefer to pass a className prop through for the button rather than rely on css selectors in the caller

aastein:
> I think we would prefer to pass a className prop through for the button rather than rely on css selectors in the caller

that feels off to me because no other sub-components of the wizard get a passed through class.

PR #370, opened 2019-05-01
Fix menu clicking behavior

cozmo:
The menu misbehaved when you clicked the trigger. 

This was because it was triggering the `focusOut` event, which was closing it, and then the click event was re-opening it. This makes it so we only trigger on `focusOut` when another element is being focused. This preserves the a11y behavior while fixing the clicking issue. 

I also removed some hacky reactDOM usage in favor of just using the built in `onBlur`. All the a11y stuff continues to work with these changes and the UX is certainly better with a mouse. 

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

### DIFF
--- a/src/Menu/Menu.tsx
+++ b/src/Menu/Menu.tsx
@@ -2,6 +2,7 @@ import * as _ from "lodash";
 import * as classnames from "classnames";
 import * as PropTypes from "prop-types";
 import * as React from "react";
+import * as RootCloseWrapper from "react-overlays/lib/RootCloseWrapper";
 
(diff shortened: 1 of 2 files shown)


src/Menu/Menu.tsx
```diff
@@ -183,14 +170,15 @@ export default class Menu extends React.PureComponent {
 
   _handleFocusOut = e => {
```
brollin: Nit: maybe we can rename `_handleFocusOut` to `_handleBlur`?

PR #374, opened 2019-05-09
fixed menu item select bug

dhspaeth12:
**Jira:**
https://clever.atlassian.net/browse/DISC-455

**Overview:**
The first menu item was always highlighted when using the mouse to hover over buttons because the focusIndex state param was only modified on arrow key press. Now mouse enter triggers the focus change and hover styling is removed

**Screenshots/GIFs:**
Before:
Screen Recording 2019-05-09 at 01 40 PM

After:
Screen Recording 2019-05-09 at 01 12 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "1.11.1",
+  "version": "1.11.2",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "1.11.1",
+  "version": "1.11.2",
```
brollin: You can bump this to: `1.12.1` since I just merged `1.12.0`

PR #383, opened 2019-05-23
Floating button

dhspaeth12:
**Jira:**
https://clever.atlassian.net/browse/DISC-444

**Overview:**
Implements fixed position floating button. Future version will add color groups to Button component so background color doesn't break hover/select button styling

**Screenshots/GIFs:**
Screen Recording 2019-05-23 at 11 58 AM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,195 @@
...
+        }}
+      >
+        <div className={classnames(cssClass.SUB_ELEMENT, className)}>
+          <Button
+            className={cssClass.BUTTON}
+            onClick={this.mainButtonHandler}
+            value={
+              active ? (
+                <div>
+                  <span className="fa fa-times" /> Close
```
aastein: Drive-by opinion: I think it would be nice for the label in the active state to be optionally user defined.

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+                ? closeLabel || (
+                    <div>
+                      <span className="fa fa-times" /> Close
+                    </div>
+                  )
+                : label
+            }
+            size={size || "regular"}
+            style={{
+              borderRadius: "500px",
```
arsalansufi: Since this style is always applied, it should just live in the stylesheet. Also, rather than using an arbitrarily large value, I'd just use `border-radius: 50%`

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+              ...(bgColor
```
arsalansufi: Rather than allowing users to specify any background color, we should define an enum for colors and internally map those to Dewey colors.

```js
export const Color = {
  BLUE: "blue",
  GREEN: "green",
  ...
};

const cssClass = {
  colorCssClass: (color: string) => `FloatingButton--color--${color}`,
  ...
};

const propTypes = {
  color: PropTypes.oneOf(Object.values(Color)),
  ...
};
```

```less
.FloatingButton--color--blue {
  background-color: @primary_blue;
}

.FloatingButton--color--green {
  background-color: @alert_green;
}
```

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,198 @@
...
+                  ...(bgColor
+                    ? {
+                        backgroundColor: bgColor,
+                        borderColor: bgColor,
+                      }
+                    : {}),
+
+                  // With mutliple clicks within transition window, we may reach active=true, display=false edge case.
+                  // We should only remove from DOM when inactive and not persisting display for transition out.
+                  display: active || display ? "inline" : "none",
```
arsalansufi: We can likely define this style in the stylesheet as well. We just need to apply a CSS class when the active prop is specified and then style accordingly. Happy to chat through what this might look like in person!

In short, I don't think you need the `style` prop for anything in this PR except for the `positionX` and `positionY`

arsalansufi:
Also! For a good example of how to take a prop and apply styles based on that prop, `<ToastNotification>` is a good example. It takes a `type` prop and applies styles based on that.

ToastNotification.tsx
-----

https://github.com/Clever/components/blob/8a2ee0ad411cbe63211bd26327f21ef864f02260/src/ToastStack/ToastNotification.tsx#L62-L63

ToastNotification.less
-----

https://github.com/Clever/components/blob/8a2ee0ad411cbe63211bd26327f21ef864f02260/src/ToastStack/ToastNotification.less#L16-L34

You can even style sub elements using the top-level class name:

https://github.com/Clever/components/blob/8a2ee0ad411cbe63211bd26327f21ef864f02260/src/ToastStack/ToastNotification.less#L49-L99

`.ToastNotification--info button.ToastNotification--action` will select every `button` with class name `ToastNotification--action` that lives inside an element with class name `ToastNotification--info`. This approach is really clean because you only have to apply the special `type` class name once to the top-level element

-----

In your case, I think `color` and `active` could both work this way, with a single class name addition to the top-level element

src/FloatingButton/FloatingButton.tsx
dhspaeth12: I tried to place it in the stylesheet but the styles from the Button component override it. I think that's partially why that component exposes the "style" prop.

dhspaeth12: To your point about border-radius, this is what border-radius: 50% looks like: 
Screen Shot 2019-05-24 at 3 17 31 PM

src/FloatingButton/FloatingButton.tsx
dhspaeth12: It's my intention to do exactly this except in the Button component. I talked with Neil about it. It's a little hacky the way I have it but I'm hoping it's ok to do it this way temporarily.

src/FloatingButton/FloatingButton.tsx
dhspaeth12: I played around with it and couldn't find a way to work without using the style prop. Putting it in the css class caused it to not animate. My guess is this has something to do with the order of operations of style application (seems the style prop takes precedence). In general, what I found is that transition does not mix well with display. Most resources I found online suggested using visibility: none instead of display, but I wasn't satisfied with that because I don't people clicking on buttons accidentally. Open to suggestions for better ways to do this but this is the best I've got.

arsalansufi: Ah yes, the `style` prop takes highest precedence. But there are still ways to get your CSS to take precedence over the `<Button>`'s internal styles. I've found that, to get styles to work on the `<Button>`, you often need to use a compound selector:

```
button.FloatingButton--button
```

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+
+  bgColor: PropTypes.string,
+  className: PropTypes.string,
+  closeLabel: PropTypes.node,
+  label: PropTypes.node,
+
+  onClick: PropTypes.func,
+
+  offsetX: PropTypes.number,
+  offsetY: PropTypes.number,
```
jish: Why do we have offsets if the developer positions the menu with screen locations `positionX` and `positionY` ("top", "left", "bottom", "right")?

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+  }
+
+  mainButtonHandler = () => {
+    const { additionalButtons, animate, onClick } = this.props;
+    const { active } = this.state;
+    if (additionalButtons && additionalButtons.length) {
+      // main button activates additional buttons
+      if (active) {
+        // stop displaying when transition to inactive state complete
+        this.setState({ active: false, display: true });
```
jish: Why do we need both an `active` state and a `display` state? Aren't these tracking the same thing? Can we use CSS transitions to handle the animations so we can avoid the race conditions you mention below?

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+        setTimeout(() => this.setState({ display: false }), animate ? 200 : 0);
+      } else {
+        this.setState({ active: true, display: true });
+      }
+    }
+    if (onClick && !active) {
```
jish: The name `onClick` implies that this event will fire every time our component is clicked. If this function only fires to open the menu, can we rename it to `onOpen` or something that more clearly illustrates that it will not actually fire on every click?

src/FloatingButton/FloatingButton.tsx
arsalansufi: Got it! Makes perfect sense to push this into the `<Button>` long term.

In the meantime though, I'm still hesitant to expose a generic color prop instead of a pre-specified enum color prop. Especially because transitioning from the generic color prop to the enum will be a breaking change. And having a generic prop allows consumers of the component to break from Design-approved colors.

Even when we do add color enum logic to the `<Button>`, the `<FloatingButton>` will need to expose the same color prop and pass it through to the `<Button>`. So might as well set up that target interface, even if we take the halfway approach of only supporting the color prop in `<FloatingButton>` and not `<Button>`

src/FloatingButton/FloatingButton.less
```diff
@@ -0,0 +1,51 @@
...
+}
+
+.FloatingButton--left {
+  .items--start();
+}
+
+.FloatingButton--center {
+  .items--center();
+  left: 50%;
+  transform: translate(-50%, 0);
```
jish: Can we be consistent in our use of positioning? If we're using flexbox positioning for our other positions, can we use flexbox here as well?

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+          cssClass.propStyle(positionX),
+          cssClass.propStyle(positionY),
+          className,
+        )}
+        style={{
+          ...this.horizontalStyle(),
+          ...this.verticalStyle(),
+        }}
+      >
+        <div className={classnames(cssClass.SUB_ELEMENT, className)}>
```
jish: Is it intentional that the `className` prop passed into this component is applied to more than one element?

src/FloatingButton/FloatingButton.tsx
arsalansufi: Ah super interesting. Looks like percentage values and specific pixel values are applied differently: https://stackoverflow.com/questions/29966499/border-radius-in-percentage-and-pixels-px-or-em

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,199 @@
...
+          <Button
+            className={cssClass.BUTTON}
+            onClick={this.mainButtonHandler}
+            value={
+              active
+                ? closeLabel || (
+                    <div>
+                      <span className="fa fa-times" /> Close
```
jish: I believe icons are added using the `<FontAwesome>` component

https://github.com/Clever/components/blob/507af1cff8d1b6c13a6943e0e72ace40f9f74a0b/src/InfoPanel/InfoPanel.jsx#L60

src/FloatingButton/FloatingButton.tsx
dhspaeth12: That did the trick!

src/FloatingButton/FloatingButton.tsx
dhspaeth12: It's still useful to be able to move the button away from the default offset from the edge of the screen (for example if they want to place it low enough to be below a top bar)

src/FloatingButton/FloatingButton.tsx
dhspaeth12: this is a little hacky. It's my way of persisting display while the transition runs. I don't believe it's possible to do this in pure css because the transition property doesn't act on display

src/FloatingButton/FloatingButton.tsx
dhspaeth12: I don't want to call it onOpen because when there's no button menu to open, this fires on every click (since we never enter the active state). Happy to remove this logic so it fires on every click though.

src/FloatingButton/FloatingButton.less
dhspaeth12: .items--center(); is consistent with items--end and items--start (align-items: [flex-start,center,flex-end] respectively).

I don't believe the left/transform are possible to achieve with flex (and I'm not using flex for the positioning elsewhere)

src/FloatingButton/FloatingButton.tsx
dhspaeth12: I removed the active check

src/FloatingButton/FloatingButton.tsx
jish: What effect are you trying to accomplish? It looks like you're already animating the buttons away with CSS transitions without using this hack? 🤔 Can you use `pointer-events: none;`?

dhspaeth12: I had to make the container pointer-events: none and subElements auto (if being displayed), but this worked

src/FloatingButton/FloatingButton.less
arsalansufi: Ah I see. I was also confused by these but now understand. We're using `items--start/center/end` for the buttons with respect to each other. But need explicit left/right/top/button values for the fixed positioning of the button group on the page. And for the left and right cases, we pass in styles via the `style` prop in the TSX code

src/FloatingButton/FloatingButton.less
```diff
@@ -0,0 +1,81 @@
...
+  transition: opacity @timingPromptly @motionSmooth, transform @timingPromptly @motionSmooth;
+}
+
+button.FloatingButton--button {
+  .boxShadow--heavy();
+  border-radius: 500rem;
+}
+
+.FloatingButton--inactive--bottom {
+  transform: translateY(2.5rem);
```
arsalansufi: To the best of my understanding, this style is only necessary for animation. It took me a while to process that. (I had to run the code locally to fully make that connection :smile:) I might leave a comment noting that

src/FloatingButton/FloatingButton.tsx
```diff
@@ -0,0 +1,169 @@
...
+  horizontalStyle() {
+    const { offsetX, positionX } = this.props;
+    if (positionX === "center") return {};
+    return positionX === "left"
+      ? {
+          left: `${offsetX}px`,
+        }
+      : {
+          // default to right if a bad positionX input provided
+          right: `${offsetX}px`,
```
arsalansufi: I know we generally use `rem` units throughout Clever for zoom-in zoom-out scaling. Rather than forcing the user to specify a pixel measure, maybe we should just let the user specify a string: `"30px"`, `"1.5rem"`. We can set up a sane `rem` default so that users don't have to worry about this for the common case

dhspaeth12: before I commit this, I get that rem is generally better for sizing but I'm not sure it's most appropriate in this case since developers probably wouldn't want offset from edge to be tied to font size by default

arsalansufi: Ah yeah, I see that. I was thinking of the case that a button is positioned at the top of the window and needs to be positioned just below a header, which will likely have its height defined in `rem` units. But the more common case is probably going to be bottom positioning. And for that case, a pixel default is okay

dhspaeth12: I realized that about rem units, deleted my comment, and used rem defaults 🤷‍♂

src/index.js
```diff
@@ -109,4 +109,5 @@ import WizardLayout from "./WizardLayout";
 export { WizardLayout };
 
 import FloatingButton from "./FloatingButton";
-export { FloatingButton };
+import { ColorGroup, PositionX, PositionY } from "./FloatingButton";
+export { ColorGroup, FloatingButton, PositionX, PositionY };
```
arsalansufi: Given that `ColorGroup`, `PositionX`, and `PositionY` are pretty generic names that other components might some day export, let's actually attach them as a static fields to the component itself rather than exporting them separately:

https://github.com/Clever/components/blob/40c581b19d1d8a15f3032123e7699bf6427ff10d/src/TopBar/index.tsx#L22-L25

```ts
export default class FloatingButton extends React.PureComponent {
  static ColorGroup = ColorGroup;
  static PositionX = PositionX;
  static PositionY = PositionY;
  ...
}
```

Client-code can then do something like this:

```jsx
import { FloatingButton } from "clever-components";

<FloatingButton
  colorGroup={FloatingButton.ColorGroup.GREEN}
/>
```

src/FloatingButton/FloatingButton.tsx
```diff
@@ -20,34 +20,58 @@ const cssClass = {
...
+  GREEN: "green",
+};
+
 const propTypes = {
   animate: PropTypes.bool,
 
   additionalButtons: PropTypes.array,
 
-  bgColor: PropTypes.oneOf(["green"]),
+  bgColor: PropTypes.oneOf(Object.values(ColorGroup)),
```
arsalansufi: Let's rename this prop `colorGroup` to match up with the enum name

src/index.js
dhspaeth12: I actually expect ColorGroup to be more generic, but on board for the others

dhspaeth12: for now I'll make it a class field though

PR #384, opened 2019-05-24
WizardLayout: add fullscreen props and styles

aastein:
**Jira:**

**Overview:**
Wizard layout style was breaking in IE 11 for bespoke fullscreen css. This PR adds a fullscreen prop the 
WizardLayout so the styles are always correct.

**Screenshots/GIFs:**
Tested fullscreen view locally in sd2 by copying over dist.
image

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/components/WizardLayoutView.jsx
+++ b/docs/components/WizardLayoutView.jsx
@@ -144,6 +144,7 @@ export default class WizardLayoutView extends React.PureComponent {
               stepper={stepper}
               subtitle="Ensure a smooth upcoming school year by following a few easy steps below."
               title="Back to school guide"
+              fullscreen={false}
             />
           </ExampleCode>
           {this._renderConfig()}
--- a/docs/components/WizardLayoutView.jsx
+++ b/docs/components/WizardLayoutView.jsx
@@ -144,7 +144,6 @@ export default class WizardLayoutView extends React.PureComponent {
               stepper={stepper}
               subtitle="Ensure a smooth upcoming school year by following a few easy steps below."
               title="Back to school guide"
-              fullscreen={false}
             />
           </ExampleCode>
           {this._renderConfig()}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "1.13.4",
+  "version": "1.14.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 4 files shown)


docs/components/WizardLayoutView.jsx
```diff
@@ -144,6 +144,7 @@ export default class WizardLayoutView extends React.PureComponent {
               stepper={stepper}
               subtitle="Ensure a smooth upcoming school year by following a few easy steps below."
               title="Back to school guide"
+              fullscreen={false}
```
cozmo: isn't this the same as just not including it?

aastein: true facts

aastein:
> Is there any reason why someone would not want to use the fullscreen prop?

for any use where it's not intended to be displayed fullscreen like the demo in the dewy docs page for the component. 

> Like doesn't this just allow us to ensure that wizard doesn't break in IE by setting it to true?

It more importantly corrects the wizard css to display correctly when it is being used as an entire page on any browser. Right now each use of this wizard in fullscreen has to override this components styles so it displays correctly:
https://github.com/Clever/sd2/blob/2689487dfc6841488323c3e873694d5095e21fb8/ui/pages/ContentMappingWizard/ContentMappingWizard.less#L13
and https://github.com/Clever/sd2/blob/1d07c3fe30bfb0bc9134b11055ff13fb48935a40/ui/pages/eoyGuide/wizard.less#L20

emilyhou:
> > Is there any reason why someone would not want to use the fullscreen prop?
> 
> for any use where it's not intended to be displayed fullscreen like the demo in the dewy docs page for the component.
> 
> > Like doesn't this just allow us to ensure that wizard doesn't break in IE by setting it to true?
> 
> It more importantly corrects the wizard css to display correctly when it is being used as an entire page on any browser. Right now each use of this wizard in fullscreen has to override this components styles so it displays correctly:
> https://github.com/Clever/sd2/blob/2689487dfc6841488323c3e873694d5095e21fb8/ui/pages/ContentMappingWizard/ContentMappingWizard.less#L13
> and https://github.com/Clever/sd2/blob/1d07c3fe30bfb0bc9134b11055ff13fb48935a40/ui/pages/eoyGuide/wizard.less#L20

Got it, thank you that makes sense.

PR #392, opened 2019-06-17
Add TypeScript Type declarations

jish:
_I would recommend reviewing this pull request commit by commit. There are some formatting changes mixed in which may make it more difficult to review._

**Jira:**

[[FEE-140] [Build process] Add type declarations to clever-components](https://clever.atlassian.net/browse/FEE-140)

**Overview:**

Some of our components are written in TypeScript. If a component is already using TypeScript we want consumers of Clever Components to have access to type declaration files. This will help ensure proper usage and help to avoid bugs caused by improper type usage.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


jish:
> "Am I wrong or don’t we need to change something in package.json to reference the exported types?"

It looks like we do need to include type declarations in the NPM package

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html

> "Also note that if your main declaration file is named index.d.ts and lives at the root of the package (next to index.js) you do not need to mark the "types" property, though it is advisable to do so."

I'll rename `index.js` to `index.ts`

jish:
> "Could we combine this PR with the types we have for components in SD2? Then every component would have types"

I don't think we should merge any types together. The TypeScript documentation encourages developers to keep types separate and combine them or depend on them as necessary.

https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#packaging-dependent-declarations

prime-time:
I was working on [this branch](https://github.com/Clever/components/compare/convert-to-ts?expand=1) which converts some files to typescript and adds more types. It still needs some a bit of work and validation, but I'm hoping to get it in a reviewable state soon. It looks like there will be some merge conflicts, so we should probably try and coordinate before merging.

jish:
> "> "How does this work with non typescript components? My gut would have said that this would make it so you can’t use any components that aren’t TS since they will have no exported types and so typescript won’t know about them..?"

I bundled this branch into Launchpad and it recognizes `.js` components as type `any` -- which I think is fine for now.

However, it did not find any components listed in sub-directories because the `index.ts` file was not reading the `index.js` files in sub-directories. After renaming them to `index.ts` TypeScript is aware that _something_ is there, even if it doesn't know what types they are.

So, we need to rename some `index.js` files to `index.ts`.

Thanks for the comments @cozmo 🙇

jish:
@lbanner3 @cozmo @prime-time I think this is ready to merge now. Would you like to give it another look?

prime-time:
> > "Could we combine this PR with the types we have for components in SD2? Then every component would have types"
> 
> I don't think we should merge any types together. The TypeScript documentation encourages developers to keep types separate and combine them or depend on them as necessary.
> 
> https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#packaging-dependent-declarations

I think we do want to port over the type definitions that Jefff made so that consumers can have their props typechecked. Instead of having them live in a single `index.d.ts` file, the prop types would live alongside each component and the compiler would generate the type declaration for this package. That's what I've been working on in the branch I linked to above, but it's not quite done yet. We could try to combine the changes together, but it might be tricky.

jish:
Spoke with @prime-time in person. TypeScript is using the `typeRoots` configuration option to look for types https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#types-typeroots-and-types

In a repository that already has its own custom types, TypeScript will continue to use those. In a repository that does not have any type declarations, TypeScript pulls the types from the Clever Components package.

PR #393, opened 2019-06-18
collapse FloatingButton after clicking any of its 'additional buttons'

ghirsch1:
**Jira:**
https://clever.atlassian.net/browse/DISC-556

**Overview:**
FloatingButton should collapse back to just the main button after any of the sub-buttons are clicked.

**Screenshots/GIFs:**
Screen Recording 2019-06-18 at 01 51 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/FloatingButton/FloatingButton.tsx
+++ b/src/FloatingButton/FloatingButton.tsx
@@ -127,6 +127,11 @@ export default class FloatingButton extends React.PureComponent {
     }
   };
 
+  additionalButtonHandler(button) {
+    this.setState({ active: false });
+    button.onClick();
+  }
+
   render() {
     const {
       additionalButtons,
@@ -201,7 +206,7 @@ export default class FloatingButton extends React.PureComponent {
                   cssClass.BUTTON,
                   colorGroup && cssClass.propStyle(colorGroup),
                 )}
-                onClick={button.onClick}
+                onClick={() => this.additionalButtonHandler(button)}
                 value={button.label}
                 size={size || Button.Size.M}
               />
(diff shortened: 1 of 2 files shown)


src/FloatingButton/FloatingButton.tsx
```diff
@@ -127,6 +127,14 @@ export default class FloatingButton extends React.PureComponent {
     }
   };
 
+  additionalButtonHandler(button) {
+    return () => {
+      const { active } = this.state;
+      this.setState({ active: !active });
```
dhspaeth12: I would explicitly set to ```false``` instead of toggling, since that's a more accurate reflection of what the additional button handler will do

ghirsch1: makes sense

src/FloatingButton/FloatingButton.tsx
```diff
@@ -201,7 +208,7 @@ export default class FloatingButton extends React.PureComponent {
                   cssClass.BUTTON,
                   colorGroup && cssClass.propStyle(colorGroup),
                 )}
-                onClick={button.onClick}
+                onClick={this.additionalButtonHandler(button)}
```
dhspaeth12: you'll want to pre-bind this with an arrow function.
```
onclick={() => additionalButtonHandler(button)}
```
For more info see https://www.freecodecamp.org/news/this-is-why-we-need-to-bind-event-handlers-in-class-components-in-react-f7ea1a6f93eb/

PR #394, opened 2019-06-25
Typescriptify

prime-time:
**Overview:**
This change converts old components written in JavaScript to TypeScript, and adds type declarations for component props. Many of the prop types were copied over from the types @jefff created.

I made this a major version bump because the stricter types will potentially surface type errors in consumers of this package. I avoided making API changes, though I did remove a few unused props. 

I also removed some static properties from components, mainly the `cssClass` objects. I don't think we refer to these outside of this library, so it seems unnecessary to include them.  

Overview of changes:
- Remove the react types shim in favor of the real types defined in [definitely typed](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react). This makes components defined as es6 classes generic over props and state.
I decided not to use the react generic type for functional components because there are unfortunately drawbacks to `React.FunctionComponent`, and only small advantages over the implicitly inferred `() => JSX.ReactElement` type. The major drawback is that `React.FC` doesn't enforce required props. I typed a small handful of components as `React.FunctionComponent` to fix some errors that occur when importing them into another project. 
- Convert all the source and test files to TypeScript. 
- Convert [legacy string refs](https://reactjs.org/docs/refs-and-the-dom.html#legacy-api-string-refs) to use the new API.
- Simplify the build step by removing the babel compilation step, and only compile with TypeScript.
- Use `const` assertions to more strongly type object literals that are used like enums. The values of these objects end up typed as string literals instead of `string`, and the union type of the values is useful for restricting props to be within a defined set of strings. Some of these are commonly used so I exported them for convenience. 
Before the Button component would accept any arbitrary string for the `size` prop, but now it will error if it's not in `"small" | "regular" | "large"`
Going forward, I recommend that we declare just a union type of string literals instead of using objects or enums. 

`const` assertion of an object:
Screen Shot 2019-06-27 at 3 59 56 PM
Screen Shot 2019-06-27 at 3 59 25 PM

**Screenshots/GIFs:**

**Testing:**
To test this branch, I ran the docs server locally and rendered all of the components. I also imported this version into several apps.

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


src/Count/Count.tsx
```diff
@@ -17,10 +46,12 @@ import Number from "../Number";
...
  */
-export default class Count extends PureComponent {
-  static format({ number, plural, short, singular, zeroOverride }) {
+export default class Count extends React.PureComponent<Props> {
+  static propTypes = propTypes;
+
+  static format({ number, plural, short, singular, zeroOverride }: FormatOptions) {
     const displayNumber = Number.format(number, short);
-    const rawNumber = parseInt(number || 0, 10);
+    const rawNumber = parseInt(`${number}` || "0", 10);
```
prime-time: `parseInt` takes a string as the first argument

src/DateInput/DateInput.tsx
```diff
@@ -52,7 +52,10 @@ const popperPlacementPositions = {
   TOP_START: "top-start",
 } as const;
 
-const dateType = PropTypes.oneOfType([PropTypes.instanceOf(Date), PropTypes.instanceOf(moment)]);
+const dateType = PropTypes.oneOfType([
+  PropTypes.instanceOf(Date),
+  PropTypes.instanceOf(moment as any),
```
arsalansufi: Curious why the `as any` is necessary here. Is there a whitelist with respect to what can be passed to `PropTypes.instanceOf`?

src/DropdownButton/Option.tsx
```diff
@@ -21,7 +21,7 @@ const propTypes = {
 
 export default class Option extends React.PureComponent<Props> {
   static propTypes = propTypes;
-  render() {
+  render(): never {
```
arsalansufi: Whoa! TIL

src/Table/tablePropTypes.ts
```diff
@@ -2,7 +2,10 @@ import * as PropTypes from "prop-types";
 
 import sortDir from "./sortDirection";
 
-export const sortDirection = PropTypes.oneOf([sortDir.ASCENDING, sortDir.DESCENDING]);
+export const sortDirection = PropTypes.oneOf<"asc" | "desc">([
```
arsalansufi: Hmm so if we ever change the `sortDirection` values,

https://github.com/Clever/components/blob/5967a9f8d5f1cfa8d7c7efa1ee739f90ecabd0ae/src/Table/sortDirection.js#L1-L4

we'll need to update them here as well.

Not a huge deal! But is there an easy way to avoid this coupling?

src/AlertBox/AlertBox.tsx
```diff
@@ -24,10 +33,25 @@ const cssClass = {
   CLOSE: "AlertBox--close",
 };
 
+const TYPES = {
```
arsalansufi: Any reason for not making this an enum? And then using that enum in the `Props` definition instead of or-ing string literals that have to be manually kept in sync with the values in this object?

```ts
enum Type {
  WARNING = "warning",
  SUCCESS = "success",
  ERROR = "error",
  INFO = "info",
}

export interface Props {
  ...
  type?: Type;
}
```

src/AlertBox/AlertBox.tsx
```diff
@@ -24,10 +33,25 @@ const cssClass = {
...
+  ERROR: "error",
+  INFO: "info",
+};
+
 /*
  * AlertBox is a closable, highlighted box
  */
-export default class AlertBox extends React.PureComponent {
+export default class AlertBox extends React.PureComponent<Props> {
+  static propTypes = {
```
arsalansufi: Now that we have real TypeScript types, is there any reason to keep old React prop types around?

src/TextArea/TextArea.tsx
```diff
@@ -31,6 +31,9 @@ export class TextArea extends React.Component {
     this.onBlur = this.onBlur.bind(this);
   }
 
+  // eslint-disable-next-line react/sort-comp
```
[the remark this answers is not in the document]
prime-time: eslint is reporting the following error:
`textAreaEl should be placed after onBlureslint(react/sort-comp)`
Which doesn't make any sense to me, since it's saying it should be placed between two instance properties. The error I got in other files told me to place the instance property after the last static property. Seems like a bug.

src/DateInput/DateInput.tsx
```diff
@@ -192,6 +192,7 @@ export default class DateInput extends React.Component<Props, State> {
             selected={this.props.value}
             popperClassName="DatePicker--Popper"
             popperPlacement={this.props.popperPlacement}
+            onChange={this.props.onChange}
```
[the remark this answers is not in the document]
prime-time: That seems like a good idea. My goal here was to avoid any potential breaking behavior, but I think my change is technically breaking since `onChange` would be called where it isn't currently been called.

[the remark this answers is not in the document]
prime-time: Something I thought of while typing `DateInput`: should we get rid of the option rendering of two different sub components? It makes it harder to type things because the subcomponents share some props, but not all and they have some props with the same name but different type. I think it's also kind of weird because these are two hefty components, and it's not like conditionally rendering a `div` or a `span`. 

When I checked the usage of `DateInput` in our codebase, I found only one place where `useTime` was used, and it's in the portal. Maybe we can simplify this component and add a shared component in the portal.

src/DateInput/DateInput.tsx
prime-time: Yeah it accepts types that can be instantiated with `new`

src/Table/tablePropTypes.ts
prime-time: This does actually type check the values of `sortDirection`, so if the values are modified then the compiler will throw an error here. It doesn't however catch if new values are added. I'm not too worried in this case since the the values of `sortDirection` are very unlikely to change. There may be a better way to do this though so I'll come back to it.

src/AlertBox/AlertBox.tsx
prime-time: Runtime type checking still provides some value, though the value is small since we often ignore the console warnings 🙈. I don't know if they're worth it for type checking. I'm mainly hesitant to remove them because I tried to avoid too many code changes in this pull request (yeah I know I still ended up with a lot of diffs). There are a bunch of places where we filter props based on `propTypes` like https://github.com/Clever/components/blob/031822e606f15f4d9e07efc890c8e70657e3d702/src/TopBar/TopBarButton.tsx#L31, so removing them wouldn't be trivial. I wouldn't be opposed to removing them in the future.

src/AlertBox/AlertBox.tsx
prime-time: What do you think about this:
```ts
type AlertBoxType = typeof alertBoxTypes;

const alertBoxTypes = ["warning", "success", "error", "info"] as const;
```

That makes `AlertBoxType` of type `"warning" | "success" | "error" | "info"`, and it can be used in the `Props` declaration. 

I think making it an enum would require `Type` to be exported, and all consumers would need to be updated to use the enum. I favor unions of literals over enums. enums are more verbose, and they tend to have keys in all caps which is way too aggressive for me. 
`AlertBox.Type.WARNING` 😨
`"warning"`  😎 
Another drawback to enums is that they emit extra javascript, whereas a union type doesn't emit any javascript.

`Type` is currently only used here
https://github.com/Clever/components/blob/5967a9f8d5f1cfa8d7c7efa1ee739f90ecabd0ae/src/AlertBox/AlertBox.jsx#L82, and it's being converted to an array so there's no reason it needs to be a plain object.

src/DropdownButton/Option.tsx
prime-time: As a maybe interesting side note, the compiler correctly infers `never` as the return type of a function that always throws
Screen Shot 2019-07-01 at 12 36 47 AM

The compiler usually does a good job of inferring the return type, but it seems that it infers `void` as the return type for `React.Component`s that unconditionally throw. `JSX.ReactElement` doesn't like being `void`.

src/TextArea/TextArea.tsx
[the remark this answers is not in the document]
arsalansufi: Got it!

src/DateInput/DateInput.tsx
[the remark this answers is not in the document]
arsalansufi: I definitely agree with removing the conditional rendering of two different underlying components. But I also think that we should maintain a `useTime` option in clever-components. It feels like a generally useful option. I wonder if `<ReactDateTime>` supports rendering without a time picker 🤔 That said, these changes are probably out of scope for this PR.

-----

Going back to the original question of `onSelect` vs. `onChange`, it seems like supplying both an `onSelect` and an `onChange` will result in redundant updates. Since `onChange` should take care of all needed updates, I imagine the `onSelect` --> `onChange` transition should be safe

src/Table/tablePropTypes.ts
arsalansufi: Ah good point about the type-checker catching incompatibilities. In that case, this feels lower priority / non-blocking

src/AlertBox/AlertBox.tsx
arsalansufi: Yeah my gut feeling is that we don't need React `propTypes` moving forward. The overhead of keeping two type definitions in sync doesn't feel worth the marginal benefit of runtime checking that we typically ignore :smile:

But great point that removing `propTypes` is a non-trivial change. We can leaving existing components as is.

Let's just update the new component script to not include `propTypes`

dhspaeth12: fwiw I discovered through local testing that changing
```
<SubComponent
  {...additionalProps} {/* props - propTypes */}
  ...
/>
```
to
```
<SubComponent
  {...this.props}
  ...
/>
```
is a breaking change

src/AlertBox/AlertBox.tsx
arsalansufi: Ah great point about having to update consumers if we switch to an enum. Given that we wanna make it as easy as possible to upgrade to the typed version of clever-components, I agree that we should do something more backwards compatible.

-----

In terms of your suggestion, I generally like it! I don't think it's super intuitive that `typeof` `["warning", "success", "error", "info"] as const` will generate a `"warning" | "success" | "error" | "info"` type. But I probably just need to brush up on my TypeScript.

-----

For new components, I think I still lean toward enums for clarity. They just feel like the right tool for the job. But also a valid point about the extra generated JS. This is a question that we can obviously hash out at a later time -- maybe next Frontend Guild meeting!

arsalansufi:
(Still working through https://github.com/Clever/components/pull/394/commits/91f8a6cfb86f56efd92d5da2159550c6a435eb68. Have a few things on my calendar. But will definitely finish reviewing today!)

src/Count/Count.tsx
```diff
@@ -4,6 +4,35 @@ import * as PropTypes from "prop-types";
...
+const propTypes = {
+  children: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  className: PropTypes.string,
+  plural: PropTypes.string,
+  singular: PropTypes.string.isRequired,
+  short: PropTypes.bool,
+  zeroOverride: PropTypes.string,
+};
+
+export const cssClass = {
```
arsalansufi: I see that previously `cssClass` was loaded onto `Count` and as a result accessible via `Count.cssClass`. Though I highly doubt that any consumer is accessing `Count.cssClass`, we may wanna keep it as a static member to minimize breaking changes. I see that we've left the `static cssClass` on other components

src/AlertBox/AlertBox.tsx
arsalansufi: Chatted with @prime-time offline about the merits of enums vs. unions of string literals:

```
enum Size {
  SMALL = "small",
  MEDIUM = "medium",
  LARGE = "large",
}

type Size = "small" | "medium" | "large";
```

I think I've been convinced of the merits of string literal unions! They give us the same type-checking benefits as enums, don't need to be imported by consumers, and are generally lighter weight.

Long-term, we can head in this direction. @prime-time It's also worth bringing this up at the next Frontend Guild meeting!

src/Stepper/Step.tsx
```diff
@@ -7,6 +7,25 @@ import CheckMark from "./CheckMark";
 import { FlexBox } from "../flex";
 import Exclamation from "./Exclamation";
 
+enum StepState {
```
arsalansufi: Given our earlier discussion on enums vs. string-literal unions, any reason you decided to go with an enum here?

src/Select/Select.tsx
```diff
@@ -3,12 +3,92 @@ import * as PropTypes from "prop-types";
...
+  required: PropTypes.bool,
+  value: PropTypes.oneOfType([
+    PropTypes.string,
+    selectValuePropType,
+    PropTypes.arrayOf(PropTypes.string),
+    PropTypes.arrayOf(selectValuePropType),
+  ]),
+  className: PropTypes.string,
+  error: PropTypes.string,
+  // Object.values isn't properly polyfilled in jsx files
```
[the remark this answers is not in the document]
prime-time: Good point, I did not look into this at all. I'm pretty sure you're right, so I'll just delete it.

src/Grid/Col.tsx
```diff
@@ -7,14 +7,26 @@ import Size from "./Size";
 
 import "../less/grid.less";
 
+export interface Props {
+  children?: React.ReactNode;
+  className?: string;
+  span?: number | { dflt?: number; s?: number; m?: number; l?: number };
+  wrapperComponent?: any;
+  // TODO: It's theoretically possible to make this component generic and type these extra props, but it's a pain
```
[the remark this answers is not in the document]
prime-time: This comment was actually copied from Jefff's code and I remember him complaining out this pattern.

src/Stepper/Step.tsx
prime-time: No good reason. I copied this one from sd2 and didn't feel like changing it. I think it's worth changing though so thanks for bringing it up.

src/AlertBox/AlertBox.tsx
prime-time: I agree, but I'm also biased 😄

src/AlertBox/AlertBox.tsx
prime-time: ugh I'm afraid to change any of that props filtering code

src/ConfirmationButton/ConfirmationButton.tsx
```diff
@@ -3,15 +3,49 @@ import * as PropTypes from "prop-types";
...
 
 import { Button } from "../Button/Button";
-import { ModalButton } from "../ModalButton/ModalButton";
+import { ModalButton, Props as ModalButtonProps } from "../ModalButton/ModalButton";
 import { propsFor, prefixKeys, unprefixKeys } from "../utils";
 
 import "./ConfirmationButton.less";
 
+export interface Props extends ModalButtonProps {
+  // The variadic type proposal would cover this, but for now, we copy-pasta
```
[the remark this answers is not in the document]
prime-time: oh yeah this was another comment copied from Jefff

src/Count/Count.tsx
prime-time: I searched sd2 and launchpad and didn't find any references, so even though it's a breaking change it shouldn't affect any consumers. Even though it's not being referenced anywhere, it feels like a bad pattern. I actually wanted to remove `cssClass` as a static property from all components, but I only ended up doing it for the ones that were converted from JavaScript. If we're going to make breaking changes, nows a good time.

As a side note, the reason I made this an export is so that the unit tests can still reference them. The unit tests use css selectors to find rendered elements. It would not be exported at the package level though.

If I can verify that these are not referenced by any consumers, would it be ok to remove these `cssClasses`, or should that be left for a future change?

src/DateInput/DateInput.tsx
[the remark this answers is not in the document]
prime-time: Ok, that's fine with me. Another option to consider is to expose a new `onSelect` prop, but we can always add that later if we need it.

src/Count/Count.tsx
arsalansufi: Ah yeah! If you can verify that it's not accessed, def feel free to remove it!

src/Stepper/Step.tsx
prime-time: the enum has been eliminated  🔪

src/Count/Count.tsx
prime-time: I checked again and didn't find any external references to these, so I went ahead and removed them from all.

src/DateInput/DateInput.tsx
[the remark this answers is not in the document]
prime-time: `onSelect` has been eliminated 🔪

prime-time:
Thanks for all the comments, @arsalansufi  🏅. I incorporated your suggestions, and rebased again to fix the merge conflicts.

PR #398, opened 2019-06-26
Add onclick handler for topbar logo

cozmo:
Sometimes you want to know if the logo has been clicked.

### DIFF
--- a/src/TopBar/index.tsx
+++ b/src/TopBar/index.tsx
@@ -14,6 +14,7 @@ const propTypes = {
   className: PropTypes.string,
   logoHref: PropTypes.string.isRequired,
   title: PropTypes.node,
+  onLogoClick: PropTypes.func,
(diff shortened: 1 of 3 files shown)


src/TopBar/index.tsx
```diff
@@ -14,6 +14,7 @@ const propTypes = {
   className: PropTypes.string,
   logoHref: PropTypes.string.isRequired,
   title: PropTypes.node,
+  onLogoClick: PropTypes.func,
```
jish: How motivated do you feel about converting these PropTypes to TypeScript? 😉

cozmo: I'd love to but last time I tried the linter complained so I reverted it. Unclear if that's still the case? I wrote this with types originally and then deleted them when I was writing the TopBar component.

PR #402, opened 2019-07-02
Create a custom route for fullscreen WizardLayout

bibekg:
**Jira:**
https://clever.atlassian.net/browse/FEE-142

**Overview:**
The `WizardLayout` component is difficult to test with its `fullscreen` prop applied. When rendered in the `WizardLayoutView` in the docs, it tries to take the full screen while the Dewey `Layout` is also trying to use the full screen. There's some conversation around this (and this temp solution) in the comments in the JIRA ticket linked above.

This PR creates a new route `/#/fullscreen-wizard-layout` for the purpose of testing the `WizardLayout` in fullscreen mode. This route renders nothing other than the `WizardLayout` and can be accessed via a button below the config checkboxes (demoed in gif below)

Furthermore, the gif shows that the `WizardLayout` styles are broken and could definitely use a good demo page like this one for debugging it.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/components/WizardLayoutView.jsx
+++ b/docs/components/WizardLayoutView.jsx
@@ -1,10 +1,11 @@
 import * as React from "react";
+import { Link } from "react-router";
 import * as PropTypes from "prop-types";
 
 import Example, { CodeSample, ExampleCode } from "./Example";
 import PropDocumentation from "./PropDocumentation";
 import View from "./View";
-import { WizardLayout, FlexBox, ItemAlign, Stepper } from "src";
+import { Button, WizardLayout, FlexBox, ItemAlign, Stepper } from "src";
 import { WizardLayoutContent } from "./WizardLayoutContent";
 
 import "./WizardLayoutView.less";
@@ -167,6 +168,13 @@ export default class WizardLayoutView extends React.PureComponent {
             {WizardLayoutToRender}
           </ExampleCode>
           {this._renderConfig()}
+          <Button
+            type="secondary"
+            size="regular"
+            // Intentionally not using Button's href prop because this link is internal and should
+            // be handled by react-router's <Link />, not an <a />
+            value={<Link to="/fullscreen-wizard-layout">View fullscreen (separate page)</Link>}
+          />
         </Example>
 
         {this._renderProps()}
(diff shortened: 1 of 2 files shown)


docs/components/WizardLayoutView.jsx
```diff
@@ -31,9 +41,11 @@ export default class WizardLayoutView extends React.PureComponent {
       goals: false,
     },
     hideSaveAndExit: false,
+    fullscreen: false,
```
johnhuangclever: We can remove this `fullscreen` state now

docs/components/WizardLayoutView.jsx
```diff
@@ -155,7 +176,7 @@ export default class WizardLayoutView extends React.PureComponent {
   }
 
   _renderConfig() {
-    const { showHeaderImg, customHelpContent, hideSaveAndExit } = this.state;
+    const { showHeaderImg, customHelpContent, hideSaveAndExit, fullscreen } = this.state;
```
johnhuangclever: here too

docs/components/WizardLayoutView.jsx
```diff
@@ -155,7 +176,7 @@ export default class WizardLayoutView extends React.PureComponent {
   }
 
   _renderConfig() {
```
johnhuangclever: Can we add a button/link here that navigates to `/fullscreen-wizard-layout`? That would be nice

docs/components/WizardLayoutView.jsx
bibekg: Oops, thanks!

docs/components/WizardLayoutView.jsx
bibekg: Good call. Was thinking of linking to it somewhere but couldn't think of where the best spot would be. This seems solid.

bibekg:
Applied your advice @johnhuangclever 

Let me know what you think. Also unsure whether to keep the "separate page" text. Was thinking it might be nice as a warning that it'll navigate away unlike the checkboxes above it.

PR #404, opened 2019-07-04
Exploring @testing-library/react as replacement for enzyme

bibekg:
**Jira:**
https://clever.atlassian.net/browse/FEE-132

**Overview:**
Ideally when testing UI components, all that is tested is the component's interface with the user, since that's ultimately what matters. Enzyme's API tends to encourage tests that test the component implementation a bit more than we'd like. We want to look into alternatives that give us more resilient, refactor-friendly tests. One possible alternative is [@testing-library/react](https://testing-library.com/docs/react-testing-library/intro). 

This PR implements react-testing-library in place of Enzyme for a single component, `AlertBox`. The resulting tests are significantly more interface-oriented than the pre-existing implementation-heavy Enzyme API. 

Although this RTL API proved to be better in this particular case, it's tough to conclude from this small example that RTL >> Enzyme in all cases. Tests would need to be written for more complex components to reach a stronger conclusion.

**Screenshots/GIFs:**

**Testing:**
- [x] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -41,6 +41,7 @@
     "tether": "^1.4.0"
   },
   "devDependencies": {
+    "@testing-library/react": "^8.0.4",
     "@types/classnames": "^2.2.3",
     "@types/jasmine": "^2.6.0",
     "@types/jasmine-enzyme": "^3.6.0",
--- a/src/AlertBox/AlertBox.tsx
+++ b/src/AlertBox/AlertBox.tsx
@@ -66,7 +66,10 @@ export default class AlertBox extends React.PureComponent<Props> {
       return null;
     }
     return (
-      <div className={classnames(`AlertBox--${type}`, cssClass.CONTAINER, className)}>
+      <div
+        data-testid="AlertBoxType"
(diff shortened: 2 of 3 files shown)


test/AlertBox_test.jsx
```diff
@@ -1,38 +1,41 @@
...
+    const { getByTestId, container } = render(
       <AlertBox type="warning" title="title" isClosable>
         <p> child </p>
       </AlertBox>,
     );
-    assert.equal(alertBox.find(".AlertBox--warning").length, 1);
-    alertBox.find(".AlertBox--close").simulate("click");
-    assert.equal(alertBox.find(".AlertBox--warning").length, 0);
+    assert.ok(container.firstChild);
+    getByTestId("AlertBoxCloseButton").click();
```
mkmathur: I think you can find the close button with `getByLabelText`, since it has an ARIA label.

PR #405, opened 2019-07-09
[IL-1158] Fix extra spacing in textarea in readonly mode

tnsardesai:
**Jira:**
https://clever.atlassian.net/browse/IL-1158
**Overview:**
When using TextArea with only 1 row of text we have a spacing issue. It is apparent when used with other components like TextInput. Example below: Announcement Body is TextArea. 
Screen Shot 2019-03-28 at 11 14 02 AM

**Screenshots/GIFs:**
IL-1158

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/TextArea/TextArea.tsx
+++ b/src/TextArea/TextArea.tsx
@@ -173,8 +173,9 @@ export class TextArea extends React.Component<Props, State> {
       rows: this.props.rows,
     };
 
+    // Don't increment rows in readOnly so that there is no extra spacing.
     let rows = this.props.rows;
-    if (this.props.placeholder) {
+    if ((this.props.placeholder || this.props.autoResize) && !this.props.readOnly) {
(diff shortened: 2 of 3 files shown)


src/TextArea/TextArea.tsx
```diff
@@ -174,15 +174,19 @@ export class TextArea extends React.Component<Props, State> {
...
       // margin
       rows = this.props.rows + 1;
     }
 
     let textarea = <textarea {...textAreaProps} rows={rows} />;
     if (this.props.autoResize) {
-      rows = this.props.rows + 1;
+      if (this.props.readOnly) {
+        rows = this.props.rows;
+      } else {
```
jish: We might want to add a comment explaining this behavior like the placeholder condition above. It also might be more clear to have it as a separate conditional instead of mixing it with the `placeholder` and `autoResize` logic, but I'll leave that decision up to you. 🙂

tnsardesai: In my new commit I simplified the logic for setting rows in one conditional statement which makes it more readable.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.0.0",
+  "version": "2.1.0",
```
jish: Technically this is a bugfix not a new feature. So I think `2.0.1` would be more appropriate

tnsardesai: fixed

PR #411, opened 2019-07-10
FEE-63 & FEE-155 optional tag in textInput, consistent styles for required tag in textArea and textInput

tnsardesai:
**Jira:**
https://clever.atlassian.net/browse/FEE-63
https://clever.atlassian.net/browse/FEE-155

**Overview:**
- Added explicit optional label for textInput which already existed in textArea
- Updated textArea so that styles are same as textInput when required tag is passed
Screen Shot 2019-07-10 at 2 45 50 PM
- Filed a jira https://clever.atlassian.net/browse/FEE-156 to investigate consolidating styles and logic for textArea/Input so that we don't get this inconsistent behavior in future

**Screenshots/GIFs:**
- textInput
Screen Shot 2019-07-10 at 3 17 32 PM
- textArea

textArea-required

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.0.5",
+  "version": "2.1.5",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -21,6 +21,7 @@ export interface Props {
   onKeyPress?: React.KeyboardEventHandler<HTMLInputElement>;
   onFocus?: () => void;
   onBlur?: () => void;
+  optional?: boolean;
   placeholder?: string;
   placeholderCaps?: boolean;
   readOnly?: boolean;
@@ -50,6 +51,7 @@ const propTypes = {
   onKeyPress: PropTypes.func,
   onFocus: PropTypes.func,
   onBlur: PropTypes.func,
+  optional: PropTypes.bool,
   placeholder: PropTypes.string,
   placeholderCaps: PropTypes.bool,
   readOnly: PropTypes.bool,
@@ -68,8 +70,22 @@ export class TextInput extends React.Component<Props, State> {
   static propTypes = propTypes;
   static defaultProps = defaultProps;
 
+  static validateProps(props) {
+    if (props.required && props.optional) {
+      throw new Error("You cannot pass both `required` and `optional` on a TextArea.");
+    }
+
+    if (["readOnly", "disabled", "inFocus"].filter(x => props[x]).length > 1) {
+      throw new Error(
+        "The readOnly, disabled, and inFocus props on a TextArea are mutually exclusive.",
+      );
+    }
+
+    return props;
+  }
+
   constructor(props: Props) {
-    super(props);
+    super(TextInput.validateProps(props));
     this.state = { inFocus: false, hasBeenFocused: false, hidden: true };
     this.onFocus = this.onFocus.bind(this);
     this.onBlur = this.onBlur.bind(this);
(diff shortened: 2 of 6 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.0.5",
+  "version": "2.1.5",
```
tnsardesai: Not sure if I should be updating minor version here. Because this commit is adding optional tag to textInput (a new thing and not an update). I think this is right but please verify.

src/TextInput/TextInput.tsx
```diff
@@ -68,8 +70,22 @@ export class TextInput extends React.Component<Props, State> {
   static propTypes = propTypes;
   static defaultProps = defaultProps;
 
+  static validateProps(props) {
+    if (props.required && props.optional) {
+      throw new Error("You cannot pass both `required` and `optional` on a TextArea.");
+    }
+
+    if (["readOnly", "disabled", "inFocus"].filter(x => props[x]).length > 1) {
```
arsalansufi: `inFocus` is actually a `state` value and not a `prop`. You'll see this if you look at the `Props` and `State` type definitions for the component:

https://github.com/Clever/components/blob/59a72dd0f241f073bc7c403170e0b3bb651a3685/src/TextInput/TextInput.tsx#L13-L39

So no need to include `"inFocus"` in this array right here. (I see that you copied this logic from `<TextArea>`. We may wanna update that validation method as well.)

-----

In terms of backwards compatibility, I wouldn't be surprised if existing consumers accidentally supply both a `readOnly` and `disabled` prop to some `<TextInput>` components. And since this code will generate a run-time error (as opposed to a compile-time error), I could see one of our client-side JS bundles crashing unexpectedly because of this.

Rather than throwing an error, a safer alternative might be to emit a `console.warning` or `console.error`. We should update the `<TextArea>` `validateProps` method as well. The main goal here is to notify developers that they've made a mistake. But we don't want that notification method to have consequences of its own (i.e. unexpected JS crashes)

package.json
arsalansufi: A minor version bump is correct! Minor version bumps represent backwards-compatible feature additions. Patch version bumps (the last digit) represent backwards-compatible bug fixes. And major version bumps (the first digit) represent backwards-incompatible changes

jish: If we do a minor version bump here, let's make sure to zero out the patch version as well 🙂 

```suggestion
  "version": "2.1.0",
```

https://semver.org/

arsalansufi: ^ Ah yes great catch!

src/TextInput/TextInput.tsx
tnsardesai: Yea makes sense. Since there are only 2 props I just changed it to use '&&'.

Kept throw new error for 'required && optional' as that was there in textArea already and a new feature in textInput so no consumer is using it yet.

src/TextInput/TextInput.tsx
```diff
@@ -68,8 +70,22 @@ export class TextInput extends React.Component<Props, State> {
   static propTypes = propTypes;
   static defaultProps = defaultProps;
 
+  static validateProps(props) {
+    if (props.required && props.optional) {
+      throw new Error("You cannot pass both `required` and `optional` on a TextArea.");
```
arsalansufi: Let's also update this validation to `console.error`. Same with the analogous block in `TextArea`

tnsardesai: As I mentioned in a previous comment
>  Kept throw new error for 'required && optional' as that was there in textArea already and a new feature in textInput so no consumer is using it yet.

Any reason why console.warn should be preferred here

arsalansufi: Ah sorry about that! Totally missed that comment. Great point about there being no consumers of this API.

My main motivator here is consistency. I could imagine someone being confused by the fact that we throw an error in some circumstances and `console.warn` in others.

For validation in other components, we also already use `console.warn`, so that seems to be what we've converged on.

https://github.com/Clever/components/blob/3a9471a2c4ef72eebbadc17d160e24949a819fbf/src/Select/Select.tsx#L153-L165

Finally, given that nothing catastrophically breaks when you specify both `required` and `optional`, I don't think we need to throw a full error.

PR #413, opened 2019-07-11
Fix source paths in docs

adamvictorclever:
**Jira:** N/A

**Overview:** 
Maybe I'm the only one that uses this, but I really like the links in the documentation that go to the code for the component.
Screen Shot 2019-07-10 at 5 39 03 PM

Unfortunately, a few URLs are wrong and many components didn't have a URL set at all. This fixes any incorrect URLs and adds them for components missing them!

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [X] Updated docs
- After merging:

### DIFF
--- a/docs/components/AlertBoxView.jsx
+++ b/docs/components/AlertBoxView.jsx
@@ -14,7 +14,7 @@ export default class AlertBoxView extends PureComponent {
     const { type } = this.state;
 
     return (
-      <View title="AlertBox" sourcePath="src/AlertBox/AlertBox.jsx">
+      <View title="AlertBox" sourcePath="src/AlertBox/AlertBox.tsx">
         <p>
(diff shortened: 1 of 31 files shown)


docs/bundle.js
```diff
@@ -295,6 +295,9 @@
 	(0, _reactDom.render)(_react2.default.createElement(
```
brollin: Assuming . you ran a `make docs` command to generate this file?

adamvictorclever: Yep! Is running `make docs` necessary? It wasn't quite clear to me

brollin: I'm pretty sure! Although I don't know the docs very well in this repo tbh...

PR #416, opened 2019-07-15
Fix new component script

brollin:
**Jira:**
N/A

**Overview:**
Since migrating this repo to TS, it seems the `new_component` script wasn't updated. Made the necessary updates, as well as made new components exported explicitly, rather than by `default`.

**Screenshots/GIFs:**
N/A

**Testing:**
Tested that adding a new component with `./bin/new_component.sh MyNewComponent`

**Roll Out:**
- Before merging:
  - [ ] ~Updated docs~
  - [ ] ~Bumped version in `package.json`~
- After merging:


package.json
```diff
@@ -16,23 +16,24 @@
     "registry": "https://registry.npmjs.org"
   },
   "dependencies": {
-    "classnames": "^2.2.5",
     "@types/classnames": "^2.2.3",
+    "@types/lodash": "^4.14.74",
+    "@types/node": "^12.6.2",
```
prime-time: We should use version 10 to match the node version https://github.com/Clever/components/blob/b91d962eff2e6d9b4028e52bcce8e4dc157d3f2e/Makefile#L4

tsconfig.json
```diff
@@ -7,7 +7,7 @@
     "noUnusedLocals": true,
     "noImplicitAny": false,
     "target": "es5",
-    "types": ["jest"]
+    "types": ["jest", "node"]
```
prime-time: Does the script not work without compiling with the node types?

brollin: Got it, so it does not unfortunately. It complained about the type of `process` 🤔

brollin: More specifically:
```
create_new_component.ts (3,29): Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i @types/node` and then add `node` to the types field in your tsconfig. (2591)
```

prime-time: Ah I didn't realize the script relied on that.

prime-time: We spoke about this in person and @cozmo suggested adding a separate tsconfig for these scripts. That seems like a good approach to me.

brollin: Thanks to our discussion offline, fixed in 3c101ec

package.json
brollin: 3c101ec

tsconfig.json
prime-time: 👌

PR #418, opened 2019-07-16
[TextInput] fixes show/hide refocus to address toolbar on voiceOver in iOS

tnsardesai:
**Jira:**
https://clever.atlassian.net/browse/IL-1312

**Overview:**
Voiceover on ipad/iphone changes focus to address bar when clicking on show/hide button with the keyboard open. Accessibility in oauth See gif.
refocus

**Screenshots/GIFs:**
refocus-fixed

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10
  - [x] Safari on iPhone

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -232,8 +232,9 @@ export class TextInput extends React.Component<Props, State> {
             type="button"
             className="TextInput--link"
             onClick={this.toggleHidden}
-            onMouseDown={() => {
-              // fixes accessibility bug
+            onMouseDown={event => {
+              // This prevents focus from jumping to the address bar or
+              // the first element on voiceover-enabled iOS devices
               event.preventDefault();
               event.stopPropagation();
             }}
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -228,7 +228,16 @@ export class TextInput extends React.Component<Props, State> {
           value={this.props.value as any}
         />
         {this.props.enableShow && (
-          <button type="button" className="TextInput--link" onClick={this.toggleHidden}>
+          <button
+            type="button"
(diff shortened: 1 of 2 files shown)


src/TextInput/TextInput.tsx
```diff
@@ -228,7 +228,16 @@ export class TextInput extends React.Component<Props, State> {
           value={this.props.value as any}
         />
         {this.props.enableShow && (
-          <button type="button" className="TextInput--link" onClick={this.toggleHidden}>
+          <button
+            type="button"
+            className="TextInput--link"
+            onClick={this.toggleHidden}
+            onMouseDown={() => {
```
arsalansufi: `onMouseDown` should give you access to an event `e` param. It's probably preferable to use that. (Right now, we're making use of the global `event` variable on the `window` object.) The change would look like this:

```jsx
onMouseDown={e => {
  e.preventDefault();
  e.stopPropagation();
}}
```

src/TextInput/TextInput.tsx
```diff
@@ -228,7 +228,16 @@ export class TextInput extends React.Component<Props, State> {
...
+              // fixes accessibility bug
```
arsalansufi: I would add a little bit of detail here. Maybe something like:

```
// This prevents focus from jumping to the nav bar on voiceover-enabled iOS devices
```

src/TextInput/TextInput.tsx
tnsardesai: Yea. Copy paste error on my part -.-

tnsardesai:
Updated based on your comments and also tested the changes again on iphone safari and chrome

PR #426, opened 2019-07-23
Autosuggest component

brollin:
**Jira:**
[IL-1311](https://clever.atlassian.net/browse/IL-1311) (accessibility ticket)

**Overview:**
Create an `<Autosuggest>` component! This is a component that is similar to `<Select>` in some ways, but has a subset of its features. We need this component to be accessible, so it was necessary to start from scratch, from an [accessible Autosuggest](https://github.com/moroshko/react-autosuggest).

This component allows you to enter text and render a list of suggestions.

As much as possible, I stuck to the underlying component's API. Similar to how the base component is wrapped in a rendering sense, the API methods are either wrapped with with some styling changes, or left untouched.

**Screenshots/GIFs:**
Screen Recording 2019-07-22 at 05 53 PM

**To do:**
- [ ] Update props documentation
- [ ] Fix the issue is where the suggestions list does not scroll down when keyboard up/down is used

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:


brollin:
Decided to add this only in `oauth`, more context here: https://github.com/Clever/oauth/pull/1530

PR #430, opened 2019-08-01
TextInputs announce errors and set aria flags

jish:
**Jira:**

[[IL-1287] Make TextInput error messages screen readable](https://clever.atlassian.net/browse/IL-1287)

**Overview:**

Aria attributes can be set on the client side or on the server side.
If they are set by the client, we need to notify screen readers that
something has been changed and needs to be announced. So we need an
`aria-live` tag on first render. If the contents of this `aria-live`
element change, the screen reader will be notified.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/TextArea/TextArea.tsx
+++ b/src/TextArea/TextArea.tsx
@@ -159,6 +159,7 @@ export class TextArea extends React.Component<Props, State> {
   render() {
     let wrapperClass = "TextArea";
     const errorMessage = this.currentErrorMessage();
+    const valid = errorMessage === "";
 
     if (errorMessage) wrapperClass += " TextArea--hasError";
 
@@ -179,6 +180,7 @@ export class TextArea extends React.Component<Props, State> {
     const inputNote = this.renderNote(errorMessage);
 
     const textAreaProps = {
+      ["aria-invalid"]: !valid,
       className: "TextArea--input",
       disabled: this.props.disabled,
       maxLength: this.props.maxLength,
@@ -221,7 +223,7 @@ export class TextArea extends React.Component<Props, State> {
           <label className="TextArea--label" htmlFor={this.props.name}>
             {this.props.label}
           </label>
-          {inputNote}
+          <span aria-live="polite">{inputNote}</span>
         </div>
         {textarea}
       </div>
--- a/src/TextArea/TextArea.tsx
+++ b/src/TextArea/TextArea.tsx
@@ -117,7 +117,7 @@ export class TextArea extends React.Component<Props, State> {
     this.textAreaEl.current.focus();
   }
 
-  currentErrorMessage() {
+  currentErrorMessage(): string {
     const { error, value, required } = this.props;
     const { hasBeenFocused } = this.state;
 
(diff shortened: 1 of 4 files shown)


test/TextInput_test.tsx
```diff
@@ -63,9 +63,11 @@ describe("TextInput", () => {
...
 
     // check the span element with error message
-    assert.equal(textInput.find("span").length, 1);
-    assert.equal(textInput.find("span").text(), "<FontAwesome /> This is an error");
-    assert(textInput.find("span").hasClass("TextInput--error"));
+    assert.equal(textInput.find("span.TextInput--error").length, 1);
+    assert.equal(
+      textInput.find('span[aria-live="polite"]').text(),
+      "<FontAwesome /> This is an error",
+    );
```
lubalexuan: 👍

src/TextArea/TextArea.tsx
```diff
@@ -179,6 +180,7 @@ export class TextArea extends React.Component<Props, State> {
     const inputNote = this.renderNote(errorMessage);
 
     const textAreaProps = {
+      ["aria-invalid"]: !valid,
```
lubalexuan: Why do we need to set `aria-invalid`?

jish: This article is referenced in the JIRA story: https://hiddedevries.nl/en/blog/2017-04-04-how-to-make-inline-error-messages-accessible

src/TextArea/TextArea.tsx
```diff
@@ -159,6 +159,7 @@ export class TextArea extends React.Component<Props, State> {
   render() {
     let wrapperClass = "TextArea";
     const errorMessage = this.currentErrorMessage();
+    const valid = errorMessage === "";
```
afumagalli: can `errorMessage` be `null`? should we do `const valid = !errorMessage` instead?

src/TextArea/TextArea.tsx
```diff
@@ -221,7 +223,7 @@ export class TextArea extends React.Component<Props, State> {
           <label className="TextArea--label" htmlFor={this.props.name}>
             {this.props.label}
           </label>
-          {inputNote}
+          <span aria-live="polite">{inputNote}</span>
```
afumagalli: polite

src/TextInput/TextInput.tsx
```diff
@@ -163,6 +163,7 @@ export class TextInput extends React.Component<Props, State> {
   render() {
     let wrapperClass = "TextInput";
     const errorMessage = this.currentErrorMessage();
+    const valid = errorMessage === "";
```
afumagalli: ditto

jish:
> "Have you tested with a screen reader to confirm that this change resolves the issue?"

Tested with Voice Over in Safari. It announces the aria instructions after we focus into and out of the text input once because that is how our required field logic kicks in currently... 🤔 I think that is a discussion we should bring up with the product and design teams, as we might want to review the required field behavior to make our text inputs more accessible. 🙂

src/TextArea/TextArea.tsx
prime-time: 🎩 
🐒 
🚌

PR #432, opened 2019-08-09
[Table] Pass noDataContent which will display in case no data in table

tnsardesai:
**Jira:**
https://clever.atlassian.net/browse/FEE-160

**Overview:**
Today if the data array is empty we show a row spanning all the columns with the text "No Data". It would be nice to allow users of the component to pass some value to override that.
Need this for https://clever.invisionapp.com/share/GDT1UXD9W4R

**Screenshots/GIFs:**
Before and when noDataContent is not passed:
table-no-data

When noDataContent is set to `<div> This is a test </div>`
table-no-data-content

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -42,6 +42,7 @@ export interface Props {
   paginated?: boolean;
   rowIDFn: Function;
   rowClassNameFn?: Function;
+  noDataContent?: React.ReactNode;
 
   // These must be all set together. TODO: enforce that
   lazy?: boolean;
@@ -77,6 +78,7 @@ const propTypes = {
   paginated: PropTypes.bool,
   rowIDFn: PropTypes.func.isRequired,
   rowClassNameFn: PropTypes.func,
+  noDataContent: PropTypes.node,
 
   // these must all be set together
   lazy: PropTypes.bool,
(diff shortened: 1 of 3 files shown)


vynmeister:
What's the use case for this? Would we ever want to keep the styling for the original "NO DATA" text and just change the message? Or are you planning to pass in a component?

src/Table/Table.tsx
```diff
@@ -384,53 +387,60 @@ export class Table extends React.Component<Props, State> {
...
-            numColumns={columns.length}
-            numPages={numPages}
-            showLastPage={!lazy}
-            isLoading={pageLoading}
-            lengthUnknown={lazy && numRows == null && !allLoaded}
-          />
-        )}
-      </table>
+        </table>
+        {displayedData.length === 0 && noDataContent}
```
vynmeister: Would you want this to be inside the `table`?

tnsardesai:
Use case: https://clever.invisionapp.com/share/GDT1UXD9W4R
Screen Shot 2019-08-09 at 3 18 28 PM

One possible solution was having a boolean that either renders the "NO DATA" row or not. I talked with @arsalansufi and decided against that as returning a table with just header seems like a bad idea.

I am not sure if we have a use case where we want to replace just the text on "NO DATA". At the moment we don't need it so that is not part of this PR

src/Table/Table.tsx
tnsardesai: That was my first idea. But I got a warning that we should only render `tr` elements inside a table. I can try passing `noDataContent` inside a cell in a row spanning all columns. Thanks for the suggestion

PR #434, opened 2019-08-30
Table row hover

johnhuangclever:
table-row-hover

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/Table/Table.tsx
+++ b/src/Table/Table.tsx
@@ -36,6 +36,7 @@ export interface Props {
   initialSortState?: SortState;
   onPageChange?: Function;
   onRowClick?: Function;
+  onRowMouseOver?: Function;
   onSortChange?: Function;
   onViewChange?: Function;
   pageSize?: number;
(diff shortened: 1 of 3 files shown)


src/Table/Table.tsx
```diff
@@ -36,6 +36,7 @@ export interface Props {
   initialSortState?: SortState;
   onPageChange?: Function;
   onRowClick?: Function;
+  onRowMouseOver?: Function;
```
adamvictorclever: Can we use a more specific type than `Function`? I see we're using `Function` a lot in this file, but it hides a lot of type information we could potentially have since all it indicates is that `onRowMouseOver` is a function and nothing about its arguments or return type. Even just something like `(e: any, rowID: any, rowData: any): void` is a lot better since it tells us the number of arguments we expect and names that indicate what they're for, which is helpful both for typechecking and for a user trying to figure out what arguments it takes.

PR #442, opened 2019-10-04
Add on click copy to copyable input

rzhang-clever:
**Jira:**
Part of: https://clever.atlassian.net/browse/DISC-937

**Overview:**
Adds a `onClickCopy` prop to CopyableInput that is called when the user clicks the copy button.

**Screenshots/GIFs:**
Tested by passing `onClickCopy={() => console.log("HEY HEY HEY")}` to the CopyableInput component:
copyable-input

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/CopyableInput/CopyableInput.tsx
+++ b/src/CopyableInput/CopyableInput.tsx
@@ -12,7 +12,7 @@ import "../less/forms.less";
 export interface Props extends TextInputProps {
   className?: string;
   enableCopy?: boolean;
-  onClickCopy?: Function;
+  onClickCopy?: () => void;
 }
 
 interface State {
--- a/src/CopyableInput/CopyableInput.tsx
+++ b/src/CopyableInput/CopyableInput.tsx
@@ -25,6 +25,7 @@ const propTypes = {
   enableCopy: PropTypes.bool,
   required: PropTypes.bool,
   size: PropTypes.string,
+  onClickCopy: PropTypes.func,
 };
 
 const defaultProps = {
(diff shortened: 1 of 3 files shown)


src/CopyableInput/CopyableInput.tsx
```diff
@@ -25,6 +25,7 @@ const propTypes = {
   enableCopy: PropTypes.bool,
   required: PropTypes.bool,
   size: PropTypes.string,
+  onClickCopy: PropTypes.func,
```
adamvictorclever: We should add a definition to the Props interface as well for Typescript support!

bgveenstra: Ah, now I see the interface [above](https://github.com/Clever/components/pull/442/files#diff-88afd73593c0110faa0f9eb8e3632e64R12).

rzhang-clever: Ah good call, thanks 👍
12225f3

src/CopyableInput/CopyableInput.tsx
```diff
@@ -12,6 +12,7 @@ import "../less/forms.less";
 export interface Props extends TextInputProps {
   className?: string;
   enableCopy?: boolean;
+  onClickCopy?: Function;
```
adamvictorclever: Ah sorry, I should have mentioned this originally - it's better to type it explicitly as `() => void` instead of `Function`. When it's typed as `Function` we know it's a function, but nothing about what arguments it takes or what it returns.

rzhang-clever: good call, that seems more reasonable to me as well
6acc2d7

PR #448, opened 2019-10-09
add linkUnderlined to button

jonjonray:
**Jira:**

**Overview:**
Includes option for accessible link under `linkUnderlined`. Same properties as `linkPlain`.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/docs/components/ButtonView.jsx
+++ b/docs/components/ButtonView.jsx
@@ -38,7 +38,6 @@ export default function ButtonView() {
           <Button type="link" href="http://clever.com" value="Link" />
           <Button disabled type="link" href="http://clever.com" value="Disabled Link" />
           <Button type="linkPlain" href="http://google.com" value="Plain Link" />
-          <Button type="linkUnderlined" href="http://google.com" value="Underlined Link" />
         </ExampleCode>
         <p>
           Here is a <Button type="linkPlain" href="//google.com" value="plain link" /> with no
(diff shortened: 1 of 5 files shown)


docs/components/ButtonView.jsx
```diff
@@ -38,6 +38,7 @@ export default function ButtonView() {
           <Button type="link" href="http://clever.com" value="Link" />
           <Button disabled type="link" href="http://clever.com" value="Disabled Link" />
           <Button type="linkPlain" href="http://google.com" value="Plain Link" />
+          <Button type="linkUnderlined" href="http://google.com" value="Underlined Link" />
```
brollin: So I was originally thinking that we would just add a new prop to `<Button>` that would be a `boolean` called `underlined`, which would conditionally apply the underline styling if and only if the `type` was `linkPlain` and `link` (the types of Buttons that are rendered as anchor tags).

I think my preference for the above comes from how I see the `type`s that we have. We basically have a minimal set of types that cover our main use cases in terms of styling (the `link` vs `linkPlain` has to do with margins around the link). Something like underlining might apply to both `link` and `linkPlain` (the two types that are rendered as anchor tags), and otherwise is identical to those other types, so I'm thinking the boolean might fit our use cases a little better. But what do you think? It's not a super strong preference!

jonjonray: Yeah, that definitely works and sounds much simpler as well, so let's do that

PR #453, opened 2019-11-06
Alertbox processing type

prime-time:
**Overview:**
This pull request introduces a `processing` AlertBox type. Now AlertBox and ToastStack have the same set of display types.

I also swapped in fontawsome icons to replace the custom SVGs that AlertBox uses. The point of that is to have a more consistent look between AlertBox and ToastStack instead of having similar but subtly different icons between them.

**Screenshots/GIFs:**

New `processing` AlertBox type:

The other AlertBox types have slightly different icons since they now use fontawsome instead of custom SVGs.

[ToastStack](http://clever.github.io/components/#/components/toast-stack)

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:


prime-time:
I wonder if the hourglass should be animated ⏳

prime-time:
Actually, hourglasses don't usually spin

src/AlertBox/AlertBox.tsx
```diff
@@ -38,10 +33,18 @@ const propTypes = {
...
   isClosable: PropTypes.bool,
 };
 
+const iconMap = {
+  processing: "hourglass-half",
+  warning: "exclamation-triangle",
+  success: "thumbs-up",
+  error: "exclamation-circle",
+  info: "bell",
+} as const;
```
jish: I'm not familiar with the `as const` declaration. What does it do here?

Could we declare our own `IconMap` type instead?

```ts
type IconMap = {
  [x in AlertBoxType]: string;
};

const iconMap: IconMap = {
  processing: "hourglass-half",
  warning: "exclamation-triangle",
  success: "thumbs-up",
  error: "exclamation-circle",
  info: "bell"
};
```

adamvictorclever:
hey @prime-time, just a heads up that I merged a components change that bumped the version to 2.18.0, so you'll need to bump the version one more time.

src/AlertBox/AlertBox.tsx
prime-time: `as const` is a relatively new TypeScript feature called const assertion. It makes the types narrower, which in this case restricts the values in the object to be in a set of string literals instead of `string`. `FontAwesome` accepts `string` as its `name` prop, so this isn't providing any real benefit here. If the `name` prop only accepted certain string literals, then the const assertion would prevent a compiler error. 

I'll just remove it for simplicity since it's not needed.

PR #464, opened 2019-12-10
Modal FocusTrap - Preventing focus trap error from bubbling 

dxu429:
**Jira:**
[FLARE-846](https://clever.atlassian.net/browse/FLARE-846)
**Overview:**
`FocusTrap` throws an error when it can not find an element in their component tree to focus on. This can happen in a Modal when `display` is set to `none`, which caused FLARE-846. There is an open issue in the `focus-trap-react` library `https://github.com/davidtheclark/focus-trap-react/issues/20` however, the library also provides an option where a fallback focus element can be set https://github.com/davidtheclark/focus-trap#focustrap--createfocustrapelement-createoptions ~~which I've used here to fallback on the current focused element, or `document.body`~~

Update: I've used `componentDidCatch` to swallow this specific error

Hiding the modal on small screen sizes was removed here: https://github.com/Clever/launchpad/commit/ed98bb1dbf572ad6f6e7c035ecc569dbdaac86fb
and the bug was mitigated by turning off the feature flag. 

**Screenshots/GIFs:**
Uncaught error thrown by FocusTrap, causes blank screen in portal
Screen Recording 2019-12-10 at 12 27 PM

With a fallback focus element:
Screen Recording 2019-12-10 at 12 26 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.23.0",
+  "version": "2.24.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -74,7 +74,10 @@ export class Modal extends React.Component<Props, State> {
   }
 
   componentDidCatch(error) {
-    if (error && error.message.includes("You can't have a focus-trap without at least one focusable element")) {
+    if (
+      error &&
+      error.message.includes("You can't have a focus-trap without at least one focusable element")
+    ) {
       return;
     }
     throw error;
--- a/src/Modal/Modal.tsx
+++ b/src/Modal/Modal.tsx
@@ -73,6 +73,14 @@ export class Modal extends React.Component<Props, State> {
     this.setState({ windowHeight: window.innerHeight });
   }
 
+  fallbackFocus() {
+    const el = document.activeElement;
+    if (el instanceof HTMLElement) {
+      return el;
+    }
+    return document.body;
+  }
+
   render() {
     // Width should be responsive with window size
     const width = Math.min(window.innerWidth, this.props.width);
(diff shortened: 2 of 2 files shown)


dxu429:
@adamvictorclever looks like I was mistaken when I said the fallback focus outside of the trap works..  The focus trap is still active, even though it is unmounted and clicks/focus will be prevented since it is still active.

The other option that was discussed was to swallow this specific error that was being thrown so that it doesn't bubble up (and possibly crashing the client app using the component). I'll make those changes here so we can discuss, although  I'm leaning towards leaving it up to the user to handle the error instead

adamvictorclever:
Hmm I don't really have any better alternatives in mind, but I'm also not sure I'm a fan of this approach. It feels really fragile since we're checking for a string match on the error message and the creator recommended against doing this https://github.com/davidtheclark/focus-trap-react/issues/20#issuecomment-362862743 Definitely curious if others have thoughts on this though.

brollin:
Hmm. Agreed on the possible fragility here. I am wondering if we can check for focusable elements? Presumably the `<FocusTrap>` is already doing that. If we don't find anything focusable, can we not render the FocusTrap around the Modal contents?

adamvictorclever:
@dxu429 Just an idea that came to me, no idea if this would work: Could we do something like
```
if (this.props.focusLocked && this.state.useFocusTrap) {
    <FocusTrap ...>
```
```
fallbackFocus() {
    this.setState({useFocusTrap: false});
    // maybe return document.body if it would error otherwise?
  }
```

Essentially use the `fallbackFocus` as a way to tell that we're rendering it with no focusable elements and stop rendering the Focus Trap?

dxu429:
> @dxu429 Just an idea that came to me, no idea if this would work: Could we do something like
> 
> ```
> if (this.props.focusLocked && this.state.useFocusTrap) {
>     <FocusTrap ...>
> ```
> 
> ```
> fallbackFocus() {
>     this.setState({useFocusTrap: false});
>     // maybe return document.body if it would error otherwise?
>   }
> ```
> 
> Essentially use the `fallbackFocus` as a way to tell that we're rendering it with no focusable elements and stop rendering the Focus Trap?

I initially avoided this bc it didn't seem like this function should have side effects. `fallbackFocus` is called during `componentDidMount` so technically `setState` call would be okay here, but it seems a bit hacky to me.. Setting the error state to not render a thing or show fallback UI are what error boundaries are for: https://reactjs.org/docs/react-component.html#static-getderivedstatefromerror

I wasn't able to get `static getDerivedStateFromError` to be called though

PR #466, opened 2020-01-13
IL-2104 - CopyableInput: accessibility & positioning

clcaelynn-clever:
**Jira:**

[IL-2104](https://clever.atlassian.net/browse/IL-2104)

**Overview:**

CopyableInput: Add `aria-live polite` tags for accessibility, and fix awkward positioning of _REQUIRED_ text vis-a-vis _Show|Copy_ toggles so that it parallels how TextInput handles it.

**Screenshots/GIFs:**

TextInput, for reference:
Screen Shot 2020-01-10 at 12 16 06 PM

OLD CopyableInput:
Screen Shot 2020-01-10 at 12 16 13 PM

NEW CopyableInput:
Screen Shot 2020-01-13 at 11 35 22 AM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/CopyableInput/CopyableInput.tsx
+++ b/src/CopyableInput/CopyableInput.tsx
@@ -93,13 +93,18 @@ export class CopyableInput extends React.Component<Props, State> {
         />
         <div className="CopyableInput--links">
           {this.props.enableShow && (
-            <button type="button" className="CopyableInput--link" onClick={this.toggleHidden}>
+            <button
+              type="button"
+              className="CopyableInput--link"
+              onClick={this.toggleHidden}
+              aria-live="polite"
+            >
               {this.state.hidden ? "Show" : "Hide"}
             </button>
           )}
           {this.props.enableCopy && (
             <CopyToClipboard text={this.props.value || ""} onCopy={this.copyPassword}>
-              <button type="button" className="CopyableInput--link">
+              <button type="button" className="CopyableInput--link" aria-live="polite">
                 {this.state.copied ? "Copied!" : "Copy"}
               </button>
             </CopyToClipboard>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.24.0",
+  "version": "2.25.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/CopyableInput/CopyableInput.less
+++ b/src/CopyableInput/CopyableInput.less
@@ -9,6 +9,7 @@
   .flexbox();
   background: @neutral_white;
   width: 100%;
+  position: relative;
 
   .TextInput {
     border: none;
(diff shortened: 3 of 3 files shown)


src/CopyableInput/CopyableInput.tsx
```diff
@@ -93,13 +93,18 @@ export class CopyableInput extends React.Component<Props, State> {
...
         <div className="CopyableInput--links">
           {this.props.enableShow && (
-            <button type="button" className="CopyableInput--link" onClick={this.toggleHidden}>
+            <button
+              type="button"
+              className="CopyableInput--link"
+              onClick={this.toggleHidden}
+              aria-live="polite"
+            >
               {this.state.hidden ? "Show" : "Hide"}
```
adamvictorclever: It looks like in Lockbox's version we also add accessibility tags to the show/hide and copied/copy text eg https://github.com/Clever/launchpad/pull/2988/files/c97eae2fba8400565e539b9d51c29fbccf30e59d#diff-866ea003e256b80e70105ef86c9bcee1R42

clcaelynn-clever: When I tested earlier, those actually seemed detrimental, as I found that adding them in seemed to make VoiceOver unable to read the current text of the button when navigating to it (via CTRL+OPT+Rarrow), though of course I'm happy to add it in if I missed something.

As-is, I found that VoiceOver correctly noted the text of the button as it changed. What does adding the additional ones (per Lockbox's version) do here?

adamvictorclever: Gotcha, thanks for testing that! I think in that case we can go with what you have. I'm not aware of specific benefits of what was done in Lockbox, it's very possible we just thought both were necessary (we're definitely still figuring out accessibility best practices)

clcaelynn-clever: As discussed, I went ahead and catalogued the exact behavior of how VoiceOver reads it for me, where "with" is with the additional tags that the Lockbox version has, and "without" is the version I have in my commit for CopyableInput. The left (of semicolon) string is what VoiceOver read out upon focusing on the button, and the right (of semicolon) string is what VoiceOver read out upon clicking the button. Parentheses denote that it was only read out _before_ clicking the button.

With additional tags:
button ; (Copy,) button
button ; press (Copy) button

Without additional tags:
Show/Hide, button ; press Hide/Show button, Show/Hide
Copy/Copied! button ; press Copy/Copied! button

I am definitely no expert in accessibility and have never thoroughly researched how these tags work or how best to deploy them, so I've just gone off of reading documentation for the tags and trying out VoiceOver myself and seeing what gives the clearest, largest amount of info!

PR #467, opened 2020-02-07
allow the select menu to stay open after selecting an item

prime-time:
**Overview:**
This change introduces an optional prop for the `Select` component called `dontCloseMenuOnSelect` which will prevent the menu from closing each time an item is selected. This behavior is especially helpful when used in combination with `multi` as it's much quicker to select multiple values if the menu stays open.

`Select` uses react-select v1 under the hood which has a prop called `closeOnSelect` (`closeOnMenuSelect in v2/3) which is defaults to true. Instead of `dontCloseMenuOnSelect`, we could mirror react-select and use a prop which defaults to true and call it something like `closeOnMenuSelect`. I'd be interested in hearing any opinions on that.

`closeOnSelect` doesn't seem to work in react-select v1.3, so I locked it to the version in the package.json
https://github.com/JedWatson/react-select/issues/3443

**Screenshots/GIFs:**
### after

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/docs/components/SelectView.jsx
+++ b/docs/components/SelectView.jsx
@@ -16,6 +16,7 @@ export default class SelectView extends Component {
       clearable: false,
       disabled: false,
       multi: false,
+      closeMenuOnSelect: true,
       readOnly: false,
       searchable: false,
       creatable: false,
@@ -55,6 +56,7 @@ export default class SelectView extends Component {
                 lazy={this.state.lazy}
                 creatablePromptFn={label => `Add new option: ${label}`}
                 multi={this.state.multi}
+                closeMenuOnSelect={this.state.closeMenuOnSelect}
                 readOnly={this.state.readOnly}
                 required={this.state.required}
                 error={this.state.error}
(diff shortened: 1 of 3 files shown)


docs/bundle.js
```diff
@@ -69779,6 +69779,7 @@
 	    className: PropTypes.string,
 	    error: PropTypes.string,
 	    size: PropTypes.oneOf(Object.values(Forms_1.FormElementSize)),
+	    dontCloseMenuOnSelect: PropTypes.bool,
```
brollin: Hmm, yeah after doing a little thinking, I think I'd prefer `closeMenuOnSelect`. My reasoning is:
* It becomes more straightforward for Clever developers to mentally map this prop to the React Select prop
* It's generally easier to reason about things when they are positively named (see for example this blog post on [avoiding negative variable names](http://davej.io/2018/05/negative-naming.html))

But those aren't super strong opinions. What do you think?

prime-time: I think you meant to comment on the diff in `Select.tsx` 😛

brollin: Lol yeah.. I read the PR description, had a comment and quickly found a place to drop it 😆 this was a pretty random place

prime-time: I also prefer positively named booleans. I can see a compelling case for either, but you swayed me in the direction of `closeMenuOnSelect`.

prime-time:
@brollin I replaced `dontCloseMenuOnSelect` with `closeMenuOnSelect` based on your suggestion.
Here is it in action

PR #477, opened 2020-02-20
[wizardlayout] support any button type

prime-time:
**Overview:**
This change lets us use any button type for the `WizardLayout`'s next button. I saw a couple of cases in some of our design mocks where the `secondary` button type is used. I'm not sure if we'd ever actually use the other button types, but they'll be there if we want them.

**Screenshots/GIFs:**
Here's an example of a design mock that uses a `secondary` button type for the next button.

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/docs/components/WizardLayoutView.jsx
+++ b/docs/components/WizardLayoutView.jsx
@@ -293,6 +293,13 @@ export default class WizardLayoutView extends React.PureComponent {
             optional: true,
           },
           {
+            name: "nextStepButtonType",
+            type: "String",
+            description: "One of primary, secondary, destructive, link, linkPlain, plain",
+            defaultValue: "primary",
+            optional: true,
+          },
+          {
             name: "onNextStep",
             type: "Function",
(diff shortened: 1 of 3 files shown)


josh-clever:
Instead of adding another prop here, I would consider inversion of control https://kentcdodds.com/blog/inversion-of-control

prime-time:
> Instead of adding another prop here, I would consider inversion of control https://kentcdodds.com/blog/inversion-of-control

Thanks for sharing that article. I agree that adding an extra prop isn't the ideal solution. If we applied the principles from the article, we could have say a `primaryButton` prop that takes in a button element: `<Button type="primary" value="something" />`, or maybe we could let the client pass in the entire footer. 

I might not need this change, so I may close this pull request, but it's good to think about how we might make the API more flexible for the future.

PR #478, opened 2020-03-04
[DISC-1370] Enable tree shaking for components

dxu429:
**Jira:**
[DISC-1370](https://clever.atlassian.net/browse/DISC-1370)

**Overview:**
Currently it is difficult for bundlers like webpack to tree-shake CommonJS modules because of its dynamic importing structure. However, it is very possible to do so with ES2015+ modules. This PR adds in the build step to outputting ES6/ES2015 modules to `dist/es` and in `package.json` adds to the `module` field the entry point for the es modules [pkg.module field](https://github.com/dherman/defense-of-dot-js/blob/master/proposal.md#typical-usage), used for module resolution.

`pkg.sideEffects` in array form tells bundlers which modules have side-effects, i.e. 

> A "side effect" is defined as code that performs a special behavior when imported, other than exposing one or more exports. An example of this are polyfills, which affect the global scope and usually do not provide an export.

We need to mark css/less files as side effect files so that the bundler will not tree shake them because css/less module imports are in this form: 
`import "abc.less";`this to compilers seems like dead code so it will be eliminated during compilation unless we specify otherwise.

**Screenshots/GIFs:**
No changes other than outputting to `dist/es`
**Testing:**
- [x] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:


prime-time:
🎄

prime-time:
https://github.com/Clever/components/blob/bd43a8980ce754d7012c5de7c9ba3d5266b0e773/src/index.ts#L4

Isn't this polyfill injection a side effect?

dxu429:
> https://github.com/Clever/components/blob/bd43a8980ce754d7012c5de7c9ba3d5266b0e773/src/index.ts#L4
> 
> Isn't this polyfill injection a side effect?

Polyfill injection is a side effect, however, we are not bundling `core-js` with this library so it isn't necessary to list it as such.

PR #479, opened 2020-03-09
Make RadioGroup and Radio generic components

adamvictorclever:
**Overview:** RadioGroup and Radio components currently always type IDs as strings and values as `any`. However, often we want to constrain the IDs to a specific set of strings or type values as a specific type. This PR makes RadioGroup and Radio generic so that we can optionally specify the types to constrain IDs and values to! 

This is important because before we were giving up type safety in both cases. As a specific example of ID types being important, take the commit form in SD2 where the user selects a radio button for the reason they're not using the Portal. In state, `noReason` is typed using a union of specific strings. However, the RadioGroup only promises that the options IDs are strings. This means that in the `onChange` callback we have to tell the compiler we know it's actually a more specific type of string, as seen here https://github.com/Clever/sd2/blob/bd6ef983d87485e4a33c2e0a4cf6a300546deeaa/ui/components/PortalCommitBanner/PortalCommitModals/PortalCommitNoModal.tsx#L69. This is problematic because we're getting less verification from the compiler! We could accidentally add an option that has an ID not of the correct type and we'd be none the wiser. Now, if RadioGroup is typed appropriately it will know the type of `id` without requiring a type assertion and would give a compile time error if an option with an invalid id was added.

I made the generics default to string and any for backwards compatibility. I expected that to get these type checking benefits the user would specifically have to add generics, but in doing testing I found that TypeScript's type inference seems to be good enough to figure out the generics on its own (at least for IDs)!

We should eventually add generics for other components as well (Select comes to mind).

**Screenshots/GIFs:**
None, it's just types

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/RadioGroup/Radio.tsx
+++ b/src/RadioGroup/Radio.tsx
@@ -1,31 +1,19 @@
 import * as classnames from "classnames";
-import * as PropTypes from "prop-types";
 import * as React from "react";
 import * as ReactDOM from "react-dom";
 
 import "./Radio.less";
 
-export interface Props {
+export interface Props<IDType extends string = string, ValueType = any> {
   checked?: boolean;
   children: React.ReactNode;
   className?: string;
   disabled?: boolean;
-  id: string;
-  onSelect?: (id: string, value: any) => void;
+  id: IDType;
+  onSelect?: (id: IDType, value: ValueType) => void;
   tabIndex?: number;
-  value?: any;
+  value?: ValueType;
 }
-
-const propTypes = {
-  checked: PropTypes.bool,
-  children: PropTypes.node.isRequired,
-  className: PropTypes.string,
-  disabled: PropTypes.bool,
-  id: PropTypes.string.isRequired,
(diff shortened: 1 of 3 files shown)


src/RadioGroup/Radio.tsx
```diff
@@ -1,32 +1,20 @@
...
-  id: string;
-  onSelect?: (id: string, value: any) => void;
+  id: IDType;
+  onSelect?: (id: IDType, value: ValueType) => void;
   tabIndex?: number;
-  value?: any;
+  value?: ValueType;
 }
 
-const propTypes = {
```
adamvictorclever: For reasons I don't totally understand, having prop types was causing type errors. I think in these cases it's fine to just remove the propTypes since we don't really get any extra value out of them.

PR #480, opened 2020-03-17
Fambam 54 resource tile

adamvictorclever:
**Jira:** https://clever.atlassian.net/browse/FAMBAM-54

**Overview:** This PR adds the ResourceTile component to Dewey! It's essentially the ResourceLinkView from Launchpad with some changes to make it more generic. Recommended reviewing commit by commit, which should make the diffs against the Launchpad version more clear.

**Screenshots/GIFs:**
Screen Shot 2020-03-17 at 3 55 38 PM

**Testing:**
- [X] Unit tests
- Manual tests:
  - [X] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [X] Updated docs
    - [X] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
      },
      ```
    - [X] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:


src/utils/index.ts
```diff
@@ -83,3 +83,28 @@ export function classNameFor(prefixes: string | string[], classSegments?: string
...
+/**
+ * Detect browser capabilities and apply faux media queries
+ *
+ * This will allow us to write CSS rules like:
+ *
+ *     .mq-hover .conditional-styles {
+ *       // only applied when the browser supports hover
+ *     }
+ */
+export function polyfillMediaQueries() {
```
adamvictorclever: These functions were taken from Launchpad - the hover styles for the ResourceTile are dependent on the `.mq-hover` class being on the document body. Since `polyfillMediaQueries` must be run for this component to work, I think it makes sense to move the function into Dewey, although every repo would still be responsible for running this code on boot.

src/ResourceTile/ResourceTile.tsx
```diff
@@ -0,0 +1,244 @@
...
+};
+
+export interface Action {
+  content: React.ReactNode;
+  href?: string;
+  onClick?: () => void;
+}
+
+export interface Props {
+  actions: [] | [Action] | [Action, Action];
```
adamvictorclever: This __may__ be an example of trying to be too strict with types so I'm open to opinions on this! Previously, we represented this as `Action[]` but always only used the first two actions in the array. Instead, I'm expressing this implicit behavior in an explicit way by  changing the type to only allow tuples of Actions with up to 2 entries. This may cause it to be slightly harder to pass in actions to this component since you have to write it in a way that convinces the compiler you're only passing in at most two actions.

src/ResourceTile/ResourceTile.tsx
```diff
@@ -36,96 +30,82 @@ export const CssClasses = {
...
 
-type Props = BasicProps & PassThroughProps;
-
 interface State {
   iconOrientation: IconOrientation;
 }
 
 export class ResourceTile extends React.PureComponent<Props, State> {
-  static defaultProps = {
+  static defaultProps: Pick<Props, "actions" | "size" | "openInSameTab"> = {
```
adamvictorclever: TypeScript support for `defaultProps` is actually really cool! For example, we make `actions` required in the `Props` above so that when we use `actions` in this component we never have to worry about it being undefined (which is more relevant if we had strict mode on). But even though `actions` is required in the Props, TypeScript treats them as optional when actually using this component since it knows it's in `defaultProps`!

src/ResourceTile/ResourceTile.tsx
```diff
@@ -36,96 +30,82 @@ export const CssClasses = {
...
 
   renderIcon() {
-    const { icon, title } = this.props;
+    const { icon, title, loadingSpinner } = this.props;
     const { iconOrientation } = this.state;
     const loading = icon === "";
 
     if (loading) {
-      return <LoadingSpinner size={LoadingSpinner.Size.S} />;
+      return loadingSpinner;
```
adamvictorclever: I'm open to thoughts on if this makes sense - previously, if the icon is an empty string we displayed a `<LoadingSpinner>`. However, we actually don't have a `LoadingSpinner` component in Dewey, each repo has its own version. As a way around this, we require the user of this component to pass in a `loadingSpinner` prop to be used in this case. It does feel a little weird that we have this special case that comes up when icon is an empty string. Maybe the prop types should make icon and loadingSpinner mutually exclusive and we use whichever is given?

src/ResourceTile/ResourceTile.less
```diff
@@ -184,7 +184,3 @@ button.ResourceTile--actionButton {
 .ResourceTile--size--small .ResourceTile--actionButton {
   font-size: @fontSizeMedium;
 }
-
-.DraggableWrapper--dragging .ResourceTile--draggableIcon {
```
adamvictorclever: Since DraggableWrapper is a Launchpad specific component that isn't yet in Dewey, I think it makes sense for this style to live in Launchpad instead of here.

src/ResourceTile/ResourceTile.tsx
arsalansufi: This seems totally reasonable to me!

src/ResourceTile/ResourceTile.tsx
arsalansufi: Nice!

src/ResourceTile/ResourceTile.tsx
arsalansufi: Mmm interesting. Feels like the real generalization here is to allow consumers to pass in a `ReactNode` for the icon, where that node isn't necessarily a loading spinner. What if we make a new optional `iconNode` prop for this? Open to other names!

docs/components/ResourceTileView.jsx
```diff
@@ -43,6 +43,11 @@ export default class ResourceTileView extends React.PureComponent {
               import { ResourceTile } from "clever-components";
               // OR
               import ResourceTile from "clever-components/dist/ResourceTile"; // Avoids importing all of clever-components.
+
+              // You also need to polyfillMediaQueries for hovering to work
```
arsalansufi: nit:  I'd rewrite `for hovering to work` as `for hover styles to work`

docs/components/ResourceTileView.jsx
```diff
@@ -43,6 +43,11 @@ export default class ResourceTileView extends React.PureComponent {
...
+              import { polyfillMediaQueries } from "clever-components";
+              // Where you application boots
```
arsalansufi: Minor typo: you --> your

src/utils/index.ts
arsalansufi: Ah right, this code is also coming back to me. Having consumers use this function seems reasonable! That said, I wouldn't be surprised if someone forgot this and wondered why hover styles weren't working.

Maybe we should draw more attention to the comment in the docs?

```
// IMPORTANT! You need to call polyfillMediaQueries for hover styles to work.
```

src/ResourceTile/ResourceTile.less
```diff
@@ -0,0 +1,186 @@
...
+@iconSizeMedium: 6.5rem;
+@iconSizeLarge: 8rem;
+@resourceTilePaddingSmall: 0.75rem;
+@resourceTilePaddingMedium: 0.875rem;
+@resourceTilePaddingLarge: 1rem;
+@fontSizeSmall: 0.75rem;
+@fontSizeMedium: 0.875rem;
+@fontSizeLarge: 1rem;
+
+.ResourceTile {
```
arsalansufi: Ooh one more thought: For components that have names that could actually conflict with in-repo components, we sometimes prefix the component styles with `dewey--`. We do this in the Dewey `TopBar` for example.

https://github.com/Clever/components/blob/e68b5faf67fa7e756e95445a0493e291ffb26a98/src/TopBar/index.less#L4

Might make sense to do that for this component as well. I wouldn't be surprised if someone set up a component named `<ResourceTile>` and had trouble styling it because Dewey styles were getting applied to it, not realizing that that was happening.

teacamperneil:
THIS IS AMAZING, TY ADAM!!

PR #481, opened 2020-03-19
Add "plain" theme to TopBar

kofi-clever:
**Jira:**
https://clever.atlassian.net/browse/FAMBAM-67

**Overview:**
Re-adding very basic support for theming to TopBar.

"Plain" theme is more minimalist - just a bar with no background, no drop shadow (will add drop shadow in-app on the content pane when scrolling).

Pending first usage in the new Family Portal!

**Screenshots/GIFs:**
[Screenshot from Gyazo](https://gyazo.com/f291dd4f7803623c04f6a5f71584d7c1)

**Mocks:**
https://www.figma.com/file/2dAb0x3W1ffZx0lJc8Ncw3/Family-Portal-MVP?node-id=755%3A1584

**Testing:**
- [n/a] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
    - [n/a] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
      },
      ```
    - [n/a] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:

### DIFF
--- a/src/TopBar/index.tsx
+++ b/src/TopBar/index.tsx
@@ -23,7 +23,6 @@ export interface Props {
   theme?: TopBarTheme;
 }
 
-
 /**
  * Global page-level header component.
  */
--- a/src/TopBar/index.tsx
+++ b/src/TopBar/index.tsx
@@ -28,7 +28,7 @@ export enum TopBarTheme {
  */
 export class TopBar extends React.PureComponent<Props> {
   static defaultProps: Partial<Props> = {
-    theme: TopBarTheme.PLAIN,
+    theme: TopBarTheme.DEFAULT,
   };
 
   static Button = TopBarButton;
--- a/src/TopBar/index.tsx
+++ b/src/TopBar/index.tsx
@@ -15,16 +15,28 @@ export interface Props {
   title?: React.ReactNode;
   customLogo?: React.ReactNode;
   onLogoClick?: Function;
+  theme?: TopBarTheme;
+}
+
+export enum TopBarTheme {
+  DEFAULT = "default",
+  PLAIN = "plain",
 }
 
 /**
  * Global page-level header component.
  */
 export class TopBar extends React.PureComponent<Props> {
+  static defaultProps: Partial<Props> = {
+    theme: TopBarTheme.PLAIN,
+  };
+
   static Button = TopBarButton;
 
+  static Theme = TopBarTheme;
+
   render() {
-    const { children, className, title, customLogo } = this.props;
+    const { children, className, title, customLogo, theme } = this.props;
 
     // If the last element is a "rounded" TopBarButton we need to add some additional padding to the right side.
     // To determine this we need to inspect the children;
@@ -45,6 +57,7 @@ export class TopBar extends React.PureComponent<Props> {
           "dewey--TopBar",
           className,
           needsRightPadding && "dewey--TopBar--rightPadding",
+          `dewey--TopBar--theme--${theme}`,
         )}
       >
         <TopBarButton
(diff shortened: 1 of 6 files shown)


src/TopBar/index.tsx
```diff
@@ -15,16 +15,28 @@ export interface Props {
...
+  PLAIN = "plain",
 }
 
 /**
  * Global page-level header component.
  */
 export class TopBar extends React.PureComponent<Props> {
+  static defaultProps: Partial<Props> = {
+    theme: TopBarTheme.PLAIN,
+  };
```
arsalansufi: I wonder if the default should actually be `TopBarTheme.DEFAULT`. Otherwise, when existing repos like launchpad update Dewey, their top bars will unexpectedly change

src/TopBar/index.tsx
```diff
@@ -15,16 +15,28 @@ export interface Props {
...
+
   static Button = TopBarButton;
 
+  static Theme = TopBarTheme;
+
   render() {
-    const { children, className, title, customLogo } = this.props;
+    const { children, className, title, customLogo, theme } = this.props;
```
arsalansufi: Question specific to the FamPo usage: One other difference between the current TopBar and the FamPo TopBar is the existence of side gutters (that vary based on screen width).

fampo

Any thoughts on how we should tackle that? We could build hardcoded gutters into the component, though we'd have to keep those in sync with the gutters we set in the FamPo. Alternatively, we could expose some `className` style prop for this.

src/TopBar/index.tsx
kofi-clever: bah - thanks for catching that! just had that in for testing

src/TopBar/index.tsx
kofi-clever: yeah, was wondering about a good way to handle that and open to suggestions! settled on letting FamPo adding the padding - just throw the top bar in a wrapper div and pad that with the same padding as the content pane

sounding like we'll have better standardised spacing on the grid in dewey2, so held off on additions here for now

arsalansufi: Makes perfect sense!

src/TopBar/index.tsx
```diff
@@ -15,16 +15,28 @@ export interface Props {
...
+export enum TopBarTheme {
+  DEFAULT = "default",
+  PLAIN = "plain",
 }
 
 /**
  * Global page-level header component.
  */
 export class TopBar extends React.PureComponent<Props> {
+  static defaultProps: Partial<Props> = {
```
adamvictorclever: I think we may want to change this type to only include props we have defaults for, eg `static defaultProps: Pick<Props, "theme">`. An issue I was seeing when working on `ResourceTile` is that TypeScript will actually use the type of `defaultProps` to determine which props can be omitted, even if they're normally required. I believe I was seeing that when the type was `Partial<Props>` it considered all props to optional since `defaultProps` would have a key for every prop.

src/TopBar/index.tsx
```diff
@@ -15,16 +15,28 @@ export interface Props {
   title?: React.ReactNode;
   customLogo?: React.ReactNode;
   onLogoClick?: Function;
+  theme?: TopBarTheme;
+}
+
+export enum TopBarTheme {
```
adamvictorclever: Our general convention recently has been to use unions of strings instead of enums, eg `"default" | "plain"`, with the advantage being that it's slightly easier for a consumer to use while still giving the same amount of type safety.

src/TopBar/index.tsx
kofi-clever: ohh, #til - thanks for calling out!

src/TopBar/index.tsx
kofi-clever: thanks! 🙏🏿 forget that we're exporting types now!

PR #482, opened 2020-03-20
make URL optional in ResourceTile

lbanner3:
**Jira:** n/a

**Overview:** url should be optional because of onClick behavior

**Roll Out:**
- Before merging:
  - [X] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/ResourceTile/ResourceTile.tsx
+++ b/src/ResourceTile/ResourceTile.tsx
@@ -61,7 +61,7 @@ export interface Props {
   overlays?: React.ReactNode[];
   size: IconSize;
   title: string;
-  url: string;
+  url?: string;
 }
 
 interface State {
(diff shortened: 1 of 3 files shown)


src/ResourceTile/ResourceTile.tsx
```diff
@@ -61,7 +61,7 @@ export interface Props {
   overlays?: React.ReactNode[];
   size: IconSize;
   title: string;
-  url: string;
+  url?: string;
```
adamvictorclever: I don't think we have to, but just pointing out that if we wanted we could make the types extra fancy and require that at least one of `url` or `onClick` is provided (right now it's technically not a type error to omit both)

adamvictorclever:
OH also make sure to version bump!

src/ResourceTile/ResourceTile.tsx
lbanner3: yeah it works, I'll give the fancier option a try

PR #484, opened 2020-04-30
Fancy progress bar

tnsardesai:
**Jira:**
https://clever.atlassian.net/browse/SSOAP-2412

**Overview:**
First commit updates color="CHANGING" according to design spec. Now <25 is red, <50 is orange, <75 is blue and <100 is green.

Second commit adds ability to add mark to the Progress Bar. Only the next mark is going to be thick others are thin. Location of marks is passed as Array<numbers>. 

Both changes are backwards compatible.

**Screenshots/GIFs:**
1) changing color
changing
2) optional marks
marks

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/ProgressBar/ProgressBar.tsx
+++ b/src/ProgressBar/ProgressBar.tsx
@@ -22,7 +22,7 @@ export interface Props {
   showLabel?: string;
   style?: React.CSSProperties;
   width?: string | number;
-  markNumbers?: Array<number>;
+  markNumbers?: number[];
 }
 
 const Size = {
@@ -100,6 +100,9 @@ export default class ProgressBar extends React.PureComponent<Props> {
     const topLabel = this._maybeTopLabel();
     const bottomLabel = this._maybeBottomLabel();
 
+    const markNumbersInRange = (markNumbers || []).filter(mark => mark < 100 && mark > 0).sort();
+    const firstAfterFill = markNumbersInRange.find(mark => mark > fill);
+
     return (
       <div className={classnames(cssClass.CONTAINER, inactive && cssClass.INACTIVE, className)}>
         <FlexBox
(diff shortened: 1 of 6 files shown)


src/ProgressBar/ProgressBar.tsx
```diff
@@ -105,6 +119,23 @@ export default class ProgressBar extends React.PureComponent<Props> {
               }}
             >
               <div className={classnames(cssClass.BAR_BORDER, cssClass.borderSize(size))} />
+              {markNumbers &&
+                markNumbers.map((element, index) =>
+                  element < 100 ? (
```
josh-clever, line 124: This looks like an eslint `no-confusing-arrow` violation https://eslint.org/docs/rules/no-confusing-arrow

I tend to agree with the sentiment of that lint rule, because the ternary operator makes it confusing. I recommend pulling the filtering logic up one level since we're just ignoring all of the elements that are too large by rendering a `null`.

```jsx
markNumbersInRange = (markNumbers || []).filter(mark => mark < 100)

{markNumbersInRange.map(...)}
```

Feel free to use whatever variable names you like, I just used those names for illustration purposes.

tnsardesai: `element < 100` is really just a safe gaurd against bad user input but I see your point that using filter will be better. 

about the eslint violation: Current code was in violation with eslint for that rule but the fix was in violation with the preetier package so `make format` was failing. Thats what the 3rd commit does. more context at https://clever.slack.com/archives/C0ERS5ABB/p1588281462092900

PR #496, opened 2020-07-14
Fambam 335 port messaging components

adamvictorclever:
**Jira:** https://clever.atlassian.net/browse/FAMBAM-335

**Overview:** This PR moves our shared messaging components into Dewey! We're doing this now because these components will also be used in the FamPo for implementing the messaging UI. Following this PR, let's convert the uses of these components in Launchpad to using the Dewey version and remove the copies from Launchpad.

Recommended reviewing commit by commit and skipping the first two commits if you're familiar with the components!

**Screenshots/GIFs:**
`<MessagingBubble>`
Screen Shot 2020-07-14 at 2 35 00 PM
`<MessaingAvatar>`
Screen Shot 2020-07-14 at 2 35 09 PM
`<MessagingInput>`
Screen Shot 2020-07-14 at 2 35 16 PM
`<MessagingThreadHistory>`
Screen Shot 2020-07-14 at 2 35 35 PM
`<MessagingThreadListItem>`
Screen Shot 2020-07-14 at 2 37 50 PM

**Testing:**
- [ ] Unit tests
We're aware there may be some quirks with these components on other browsers/screen sizes - these will be addressed before messaging launch.
- Manual tests:
  - [X] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [X] Updated docs

Gonna skip adding to ComponentsList for now, can follow up with that in the future.
      ```
      {
      },
      ```
- After merging:


src/MessagingAvatar/MessagingAvatar.tsx
```diff
@@ -11,7 +11,7 @@ const cssClasses = {
 type Props = {
   className?: string;
   text: string;
-  color?: { color: string } | { seed: string };
+  color?: { color: string, seed?: undefined } | { seed: string, color?: undefined };
```
adamvictorclever: I didn't realize this at the time I suggested it, but it turns out the `A | B` approach isn't sufficient for making mutually exclusive props. Previously, something like `{ color: "fff", seed: "abc" }` would have still been allowed even though it's really supposed to be one or the other. The fix is to explicitly mark the unused props as "undefined" or "never". 

This seems reasonable right now that there's only two props, but if we add more I think it's a fair question whether it's worth the effort to maintain the mutual exclusiveness. Either way, I think sticking with it for now is fine and we can always make it more permissive later without it being a breaking change.

src/MessagingThreadListItem/MessagingThreadListItem.tsx
```diff
@@ -47,12 +47,12 @@ export const MessagingThreadListItem: React.FC<
...
     className,
     icon,
-    isRead,
+    isRead = true,
     timestamp,
     title,
     hasDraft,
     selected,
-    status,
+    status = "active",
```
adamvictorclever, line 55: The intent with this commit is to make `MessagingNewThread` unnecessary by making `MessagingThreadListItem` flexible enough for both cases. As part of that, I made some props optional and also defaulted them to values that make the display the most "normal". 

In general, the merging into a single component seems to work pretty well. There is one visual bug I noticed where when displaying with a draft icon and only a title and no `children`, the title is slightly misplaced and not aligned with the draft icon, but I didn't want to block this PR on fixing that.

Screen Shot 2020-07-14 at 3 51 38 PM

.eslintrc.yml
```diff
@@ -54,6 +54,9 @@ rules:
     - 2
     - "never"
 
+  react/prop-types:
```
adamvictorclever, line 57: No clue why this started popping up as I worked on this change, but given that we're primarily using TypeScript over prop types I think it's fine to just turn this off.

src/MessagingInput/MessagingInput.tsx
```diff
@@ -21,47 +21,63 @@ interface Props {
...
 }
 
-export const MessagingInput: React.FC<Props & { ref?: React.Ref<TextArea> }> = React.forwardRef(
-  (props, ref) => {
-    const { className, recipientName, value, onChange, onSubmit, onBlur } = props;
+export interface MessagingInputHandle {
+  focus(): void;
+}
+
+const MessagingInputRenderFunction: React.ForwardRefRenderFunction<MessagingInputHandle, Props> = (
```
adamvictorclever, line 28: Some context on the change here: before we were exposing a `ref` to `TextArea`, which I thought felt a bit to specific (it shouldn't matter which component this uses under the hood). Instead, I only expose the `focus` method on the `MessagingInputHandle` interface and use `useImperativeHandle` to implement that.

src/MessagingAvatar/utils.ts
```diff
@@ -0,0 +1,70 @@
+/** A utility function that can be used with MessagingAvatar to convert a name into initials. */
+export function convertNameToInitials(name: string): string {
```
adamvictorclever, line 2: Given how useful this is and that we're going to need it in FamPo, I figured it made sense to add it as a util we export from Dewey.

src/MessagingBubble/MessagingBubble.tsx
```diff
@@ -10,11 +10,11 @@ const cssClasses = {
 
 interface Props {
   className?: string;
-  content: React.ReactNode;
+  children: React.ReactNode;
```
adamvictorclever, line 13: This is a small breaking change - I felt like it was more idiomatic to just use `children` for the content to display instead of another prop.

src/MessagingThreadListItem/MessagingThreadListItem.tsx
clcaelynn-clever: Hm, looks like both the title _and_ draft indicator are misaligned in this scenario, [as compared to the mocks](https://www.figma.com/file/paAte3jkRiLpgN0i9TvPbN/Clever-Messaging?node-id=1276%3A37183). Note: Is the linked mock still the most up-to-date for this?

I believe this breaks the new-thread-with-draft use case, which is currently working on the Teacher and Student pages. I'm a little reluctant to do that, if at all avoidable. 😕

src/MessagingAvatar/utils.ts
clcaelynn-clever: I can't wait to see the revisions this one undergoes. Turns out that 'convert an unrestricted name input to initials' is quite thorny. 😄

src/MessagingThreadListItem/DraftPencilIcon.tsx
```diff
@@ -0,0 +1,19 @@
...
+export function DraftPencilIcon() {
+  return (
+    <svg
+      xmlns="http://www.w3.org/2000/svg"
+      width="12"
+      height="12"
+      fill="none"
+      viewBox="0 0 12 12"
+    >
+      <title>message draft icon</title>
```
clcaelynn-clever, line 6: I believe this title takes care of screen-reader accessibility for the icon the same way as alt text on an img, right?

docs/components/MessagingThreadHistoryView.jsx
```diff
@@ -0,0 +1,125 @@
...
+  CONFIG_OPTIONS: "MessagingThreadHistoryView--configOptions",
+  CONFIG: "MessagingThreadHistoryView--config",
+  CONFIG_TOGGLE: "MessagingThreadHistoryView--configToggle",
+  CONTAINER: "MessagingThreadHistoryView",
+  INTRO: "MessagingThreadHistoryView--intro",
+  PROPS: "MessagingThreadHistoryView--props",
+};
+
+// Start with an older time so we can better display the day separators.
+// We display any dates in the future as "today".
```
clcaelynn-clever, line 22: Meant to address this at some point: Do you think that's undesirable behavior?

Similar question with regard to years, as the current dividers don't really display them. e.g. 5/8/2019 and 5/8/2020 are displayed identically.

clcaelynn-clever:
Wow, amazing job corralling this into a cohesive PR! So many components. ✨

Per the comment, assuming I'm looking at the correct mocks, my only reservation is about the breaking changes resulting from merging the two Thread components. This affects both the Teacher & Student pages, in a use case that we seem likeliest to see right after release (thread that has no messaging history and has a draft). Is a quick fix there readily available, or is it thorny?

src/MessagingThreadListItem/DraftPencilIcon.tsx
adamvictorclever: Yep, that's my understanding!

docs/components/MessagingThreadHistoryView.jsx
adamvictorclever: Hmm yeah good questions. In theory, we should never end up with times in the future, right? In practice it's the classic  "wtf how do timezones work??". But I think for now assuming we won't have times in the future seems okay, we can always revisit this later if need be.

I also think including years doesn't seem super important. I think the plan is to clear out old messages after a school year, right? So we probably won't end up with messages that are the same date in different years.

src/MessagingThreadListItem/MessagingThreadListItem.tsx
adamvictorclever: @clcaelynn-clever Just pushed a commit that I think fixes this. It ended up kind of clunky, but a simpler way was escaping me (if you see one let me know 😄 )

Screen Shot 2020-07-14 at 5 33 50 PM

_Smiley Dude once more has reason to smile_

docs/components/MessagingThreadHistoryView.jsx
clcaelynn-clever: Sounds good to me! 👍

src/MessagingThreadListItem/MessagingThreadListItem.tsx
clcaelynn-clever: A simpler way isn't obvious to me, so I think we can definitely take function over form here! Can be refactored at any time, after all - launching on-time is a priority right now. 😅 

Thanks for resolving this!

PR #497, opened 2020-07-17
Allow for messaging-avatar to display images

nbhatia823:
**Overview:**
Allow for `MessagingAvatar` component to display profile pictures instead of just initials. For now, the image is an optional prop, as when we are rendering a teacher avatar, they may or may not have a picture saved to their profile. By making `imageSrc` an optional prop, we can render a single `MessagingAvatar` that has `imageSrc=teacherProfile.imageURL`, where `imageURL` will possibly be undefined, instead of one for if the profile image is provided, and another if it isn't. The negative is that if an `imageURL` is guaranteed to be there, we will still be required to pass the `text` and `color` props, but this use case is not currently one we have.

**Screenshots/GIFs:**
Screen Shot 2020-07-17 at 4 23 51 PM

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:


src/MessagingAvatar/MessagingAvatar.tsx
```diff
@@ -4,23 +4,32 @@ import * as classNames from "classnames";
...
+  IMAGE: "Avatar--Image",
   CIRCLE: "Avatar--Circle",
   TEXT: "Avatar--Text",
 };
 
 type Props = {
   className?: string;
   text: string;
-  color?:
+  color:
```
nbhatia823, line 15: discussed with @adamvictorclever that this was meant to be required

src/MessagingAvatar/MessagingAvatar.tsx
```diff
@@ -4,23 +4,32 @@ import * as classNames from "classnames";
...
       };
+  imageSrc?: string;
 };
 
-export const MessagingAvatar: React.FC<Props> = ({ className, text, color }: Props) => {
+export const MessagingAvatar: React.FC<Props> = ({ className, text, color, imageSrc }: Props) => {
+  // If an imageSrc is provided, it takes precedent over displaying text
+  if (imageSrc) {
+    return (
+      <img alt="user avatar" className={classNames(cssClasses.IMAGE, className)} src={imageSrc} />
```
clcaelynn-clever, line 29: Since we require `text` as a prop, can we put that in the alt text here? Something like `${text} user avatar`, or else every single MessagingAvatar on the page will have identical alt text despite representing different people. 🙂 

This, incidentally, feels like a good reason to keep `text` as a mandatory prop, or else I'm not sure how we could communicate to users using screen-readers whose avatar they're on. Small detail, but I think it matters.

src/MessagingAvatar/MessagingAvatar.tsx
clcaelynn-clever: Can you verify that this isn't a breaking change on the Student or Teacher pages? Not that you shouldn't make it, just seems good to check.

src/MessagingAvatar/MessagingAvatar.tsx
nbhatia823: Ooo good call!

docs/components/MessagingAvatarView.jsx
```diff
@@ -50,6 +50,12 @@ export default class MessagingAvatarView extends React.PureComponent {
               text={convertNameToInitials("Kristen Stark")}
               color={{ seed: "52c5e9d0e6ed0694212d69ed" }}
             />
+            <br />
+            <MessagingAvatar
+              text={convertNameToInitials("Dewey the Fox")}
```
clcaelynn-clever, line 55: Doesn't matter cause we have an image anyway, but I might suggest different initials. 😅

src/MessagingAvatar/MessagingAvatar.tsx
adamvictorclever: Worth noting that the text passed in will normally be the initials. Is that descriptive enough for the alt text?

Otherwise maybe a couple alternatives:
- have this component take the full name and internally apply the initials function. I intentionally was hesitant to do this to keep the component flexible with accepting any text, but I don't feel too strongly about this.
- add an additional prop specifically for alt text

nbhatia823: @clcaelynn-clever will be interesting to hear your opinion on this in regards to accessibility. While the full name would be more complete, just want to point out that if we were to provide the whole name, the non-image alt-text would be more descriptive than even the text avatar would be, as the text avatar at most has initials.

clcaelynn-clever: I had figured it was descriptive enough based on an assumption that users employing screen-readers would probably navigate name-to-name when looking for a desired thread (or use the search!), and that the name would always be available on any given thread, but why not aim higher!

@nbhatia823 Hopefully it's pretty simple to change the component to simply apply the already-in-Dewey-next-to-this-component initials function? IIRC I wrote it to ignore any input that's an unbroken string of <=3 letters (e.g. "A" or "JR" or "ABC"), so it won't even break anything on the Student or Teacher pages. 🙂

clcaelynn-clever: The alt text's purpose is to explain the image in a text-only fashion, so it's actually not really a problem for the image-having avatar to have full-name alt text and the initials-only avatar to only have the initials as alt text - the initials _are_ all that's in the image, so it's appropriate! 🙂 

That said, perhaps more descriptive alt text for the image-having avatar would be good, something like '$(text) user avatar image`? Trying to keep the name first, since that's the most important and least repetitive info (which should come first in things like titles, alt text, aria-labels, etc., to avoid wasting someone's time if they're using a screen-reader).

clcaelynn-clever: (@afumagalli you have more experience with accessibility work than I do, so please correct me if I'm mistaken!)

nbhatia823: @clcaelynn-clever yes that makes sense! Okay so it seems the conclusion is that we will now have the text prop be a name, and then use `convertNameToInitials` to get the text for the avatar?

@adamvictorclever while it is less flexible as you mentioned, to me it seems to make sense for the component to do the conversion itself, instead of each use case needing to import `convertNameToInitials`. In the rare case that one wants to input specific text, they could also space separate what they want to display 😄

clcaelynn-clever: I think it's reasonable that the `text` prop is abbreviated in the avatar, yeah! The to-initials handling is particularly tuned for the expectation that something name-like is passed in, but it works for any string. Additionally, considering the tiny size of the component (visually), it's hard to think that a consumer could expect more than 3 letters to actually show up.

This also makes the component more robust, as (a) it will protect itself from over-long text being passed in, and (b) it will have better accessibility for our users.

tl;dr Sounds awesome, go for it! 🎊

afumagalli: too long didn't read this whole thread 😛 

you probably want this to say `${displayName}'s user avatar` to distinguish between different avatars and make it clear that it matches up to the thread for `displayName`

adamvictorclever: @nbhatia823 SGTM! I could imagine one possible way to still have this flexibility would be to introduce an optional `textToInitials` prop that defines the function to be used for converting the text to initials, defaulting to our built in function. That said, I don't think we need to add that now for a use case that doesn't really exist, but it's nice to know we have a nice path forward if we decide we want that flexibility in the future!

PR #500, opened 2020-07-20
Support button children

prime-time:
**Overview:**
The content for a `Button` can now be passed in as its children. Content can still be passed in as `value` for backwards compatibility. 

Instead of this
```js
<Button
  type="primary"
  disabled
  value={
    <>
      <span className="fa fa-spin fa-spinner" /> Saving...
    </>
  }
/>;
```

we can now do this
```js
<Button type="primary" disabled>
  <span className="fa fa-spin fa-spinner" /> Saving...
</Button>;
```

I updated some of the examples in the `Button` documentation to use children instead of `value`, but not all. The examples I updated are the ones where the `Button`'s content is a react node since it makes the code a bit more compact. We could update all of the examples to use children if we want to encourage using children over `value`, but the code is equally compact when the content is just a string. Should we change all examples to use children instead of `value`? If we really wanted to be consistent everywhere and encourage using children, we could also make the same change in all components that contain `Button`s.

**Screenshots/GIFs:**
Buttons using `children`
Screen Shot 2020-09-23 at 1 10 49 PM

ModalButton still works
Screen Shot 2020-09-23 at 1 10 29 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/docs/components/ButtonView.jsx
+++ b/docs/components/ButtonView.jsx
@@ -123,9 +123,18 @@ export default function ButtonView() {
       <PropDocumentation
         availableProps={[
           {
+            name: "children",
+            type: "ReactNode",
+            optional: true,
+            description:
+              "The content that appears in the button, and takes precedence over 'value'",
+          },
+          {
             name: "value",
-            type: "string",
-            description: "The text that appears on the button",
+            type: "ReactNode",
+            optional: true,
+            description:
+              "(Deprecated, use children instead) The content that appears in the button if children are not specified",
           },
           {
             name: "type",
--- a/docs/components/ButtonView.jsx
+++ b/docs/components/ButtonView.jsx
@@ -59,31 +59,21 @@ export default function ButtonView() {
 
       <Example title="Button with HTML content">
         <Button type="primary" value={<span className="fa fa-search" />} />
-        <Button
-          type="destructive"
-          value={
-            <div>
-              <span className="fa fa-trash" /> Remove
-            </div>
-          }
-        />
-        <Button
-          disabled
-          value={
-            <div>
-              <span className="fa fa-spin fa-spinner" /> Saving...
-            </div>
-          }
-        />
-        <Button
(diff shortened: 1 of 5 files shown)


src/ModalButton/ModalButton.tsx
```diff
@@ -60,7 +61,7 @@ export class ModalButton extends React.Component<Props, State> {
   }
 
   render() {
-    const buttonProps = propsFor(Button, this.props);
+    const buttonProps = omitKeys(propsFor(Button, this.props), ...excludedButtonProps);
```
arsalansufi, line 64: Nice catch!

docs/components/ButtonView.jsx
```diff
@@ -111,21 +102,28 @@ export default function ButtonView() {
...
+              "The content that appears in the button, and takes precedence over 'value'",
+          },
           {
             name: "value",
-            type: "string",
-            description: "The text that appears on the button",
+            type: "ReactNode",
+            optional: true,
+            description: "The content that appears in the button if children is not passed in",
           },
```
arsalansufi, line 127: > Should we change all examples to use children instead of value? If we really wanted to be consistent everywhere and encourage using children, we could also make the same change in all components that contain Buttons.

Nah, I'd say don't worry about it! The reality is that, even if we were to update all examples in clever-components, people would still see examples of the old `value` prop in code, which they're likely to copy.

The one other thing that I might do to make it clear that `children` is the way of the future is to add a deprecation note to the description of the value prop, e.g.

```
"Deprecated. Use children instead. This content will appear in the button if children are not specified."
```

arsalansufi (approved):
Thanks for tackling this Renato! I remember you brought this up at a Frontend Guild meeting long long ago haha. LGTM!

PR #501, opened 2020-07-20
Messaging avatar touchups

nbhatia823:
**Jira:**

**Overview:**

The way that centering of the avatars was done - using `display: table` and text centering, would cause weird behavior when interacting with other elements (i.e. in screenshot within div that had a border. Instead, this PR changes the Avatar to just use flex with centering, which better fits the actual use of the avatar and achieves the same benefit the `display: table` did (I think it was previously chosen because the default `display: inline`[ suffered from this
](https://stackoverflow.com/questions/5804256/image-inside-div-has-extra-space-below-the-image).

Old - 
Screen Shot 2020-07-18 at 11 20 33 AM

New - 
Screen Shot 2020-07-20 at 1 42 13 PM

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingAvatar/MessagingAvatar.tsx
+++ b/src/MessagingAvatar/MessagingAvatar.tsx
@@ -39,9 +39,8 @@ export const MessagingAvatar: React.FC<Props> = ({ className, text, color, image
     <div
       className={classNames(cssClasses.CIRCLE, className)}
       // If defined, use color. Otherwise, use seed to determine color.
-      // If all else fails, we can use text as the seed to determine color.
       style={{
-        backgroundColor: color.color || `#${_determineAvatarColor(color.seed || text)}`,
+        backgroundColor: color.color || `#${_determineAvatarColor(color.seed)}`,
       }}
     >
       <div className={cssClasses.TEXT}>{convertNameToInitials(text)}</div>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.45.0",
+  "version": "2.45.1",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


src/MessagingAvatar/MessagingAvatar.tsx
```diff
@@ -39,9 +39,8 @@ export const MessagingAvatar: React.FC<Props> = ({ className, text, color, image
     <div
       className={classNames(cssClasses.CIRCLE, className)}
       // If defined, use color. Otherwise, use seed to determine color.
-      // If all else fails, we can use text as the seed to determine color.
       style={{
-        backgroundColor: color.color || `#${_determineAvatarColor(color.seed || text)}`,
```
nbhatia823, line 44: Would there be benefit to leave this for a situation where `color.seed = ""?`

adamvictorclever: I agree this seems unnecessary since we always require providing the color or seed

adamvictorclever: Oh just re-read your comment, good point that it could technically be an empty string. Technically, so could `text`, but that seems unlikely. I could go either way on this, providing an empty string for seed shouldn't really be allowed but it's not terrible to have a fallback if someone does.

nbhatia823: That's a good point, that `text` could be empty too. If `seed` is empty, it defaults to a light blue based on the `determineAvatarColor` function, so I think that's sufficient rather than relying on `text`. I'll go with this change I've made. Thanks!

PR #503, opened 2020-07-24
make rendering Modal through react portal optional

jkwik:
**Jira:**

**Overview:**
GetClever is a google chrome tab managed web app that replaces the document body with an iframe containing the rendered `family-media-center` application. When a families subscription is expired, we want to immediately render a non-closable Modal prompting them to subscribe in order to continue using the product.

Currently, if `createPortal` is supported, Modal renders itself through reacts `createPortal` into `document.body` by default. In GetClever's case, this results in two Modals being rendered, one on top of each other, one in the document body, and the other in the resulting iframe:

Screen Shot 2020-07-23 at 4 24 47 PM

This adds an optional parameter `createPortal` that defaults to true, and when false, returns the rendered Modal as is, bypassing `createPortal`.

**Testing:**
- [x] Unit tests -still passing
- Manual tests:
  - [x] Chrome
  - [x] Safari - tested not passing in `createPortal` as well as passing both false and true
  - [ ] IE10

**Testing with GetClever**:

- with `createPortal` not specified:
Screen Shot 2020-07-23 at 4 24 47 PM

- with `createPortal` set to false:
Screen Shot 2020-07-24 at 8 43 17 AM

- with `createPortal` set to true:
Screen Shot 2020-07-23 at 4 24 47 PM

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/docs/components/ModalView.jsx
+++ b/docs/components/ModalView.jsx
@@ -86,6 +86,13 @@ export default class ModalView extends Component {
               defaultValue: "true",
               optional: true,
             },
+            {
+              name: "createPortal",
+              type: "Boolean",
+              description: "Whether or not to render the modal into document.body through react portal (if supported)",
+              defaultValue: "true",
+              optional: true,
+            },
           ]}
           className={cssClass.PROPS}
           title="Modal"
--- a/docs/components/ModalView.jsx
+++ b/docs/components/ModalView.jsx
@@ -89,7 +89,8 @@ export default class ModalView extends Component {
             {
               name: "createPortal",
               type: "Boolean",
-              description: "Whether or not to render the modal into document.body through react portal (if supported)",
+              description:
+                "Whether or not to render the modal into document.body through react portal (if supported)",
               defaultValue: "true",
               optional: true,
             },
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.46.0",
+  "version": "2.47.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


jonjonray:
> Currently, if createPortal is supported, Modal renders itself through reacts createPortal into document.body by default. In GetClever's case, this results in two Modals being rendered, one on top of each other, one in the document body, and the other in the resulting iframe:

@jkwik could you point me to the component in GetClever that's causing the double render issue?

jkwik:
> > Currently, if createPortal is supported, Modal renders itself through reacts createPortal into document.body by default. In GetClever's case, this results in two Modals being rendered, one on top of each other, one in the document body, and the other in the resulting iframe:
> 
> @jkwik could you point me to the component in GetClever that's causing the double render issue?

@jonjonray yup! Here is where we manipulate the DOM to put all rendered contents into an iframe: https://github.com/Clever/sidekick/blob/vlc0515-release/src/lib/ui/LearningClubNavBar/index.ts#L23-L47. The reason why react renders twice is still a gray area to me, after a bit of research I think it's a result of React's strict mode.

jonjonray:
Got it! Yeah seems odd but since the default for this prop is true anyways, it should be backwards compatible

jkwik:
> Got it! Yeah seems odd but since the default for this prop is true anyways, it should be backwards compatible

Yeah I figured an appropriate change in the future (when we need the flexibility for Modal to do so) is to be able to pass a DOM element as a prop and have Modal render itself through portal into that DOM element if specified (and into document.body if not supplied). I tried messing around with this for a while but couldn't figure it out and this solution does what I was looking for 😄 .

PR #506, opened 2020-07-31
Small messaging changes

nbhatia823:
**Jira:**

**Overview:**

Messaging fixes:
1. Drop `recipientName` from `messagingInput` to appease pendo security risk
2. Have button height match input
3. Add padding to messaging bubbles on mobile

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.47.2",
+  "version": "2.47.3",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/MessagingInput/MessagingInput.less
+++ b/src/MessagingInput/MessagingInput.less
@@ -40,7 +40,7 @@
 
 button.MessagingInput--SendButton {
   // Sizing for this button is pretty particular
-  padding: 0.625rem @size_l 0.625rem;
+  padding: 0.6875rem 1.5rem 0.6875rem;
 }
 
 // For mobile/tablet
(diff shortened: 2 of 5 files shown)


src/MessagingInput/MessagingInput.less
```diff
@@ -40,7 +40,7 @@
 
 button.MessagingInput--SendButton {
   // Sizing for this button is pretty particular
-  padding: 0.625rem @size_l 0.625rem;
+  padding: 0.6875rem 1.5rem 0.6875rem;
```
clcaelynn-clever, line 43: `line-height: normal` is what we're missing here that would prevent different heights in different environments (as we've seen).

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.47.2",
+  "version": "2.47.3",
```
clcaelynn-clever, line 3: I think it should be 2.48.0, since we're actually breaking MessagingInput by removing a prop.

nbhatia823: Oh okay, in that case its a major change right, not minor?

clcaelynn-clever: The component isn't in use in production, so it isn't breaking anything, come to think of it. Let's just leave it as v2.47.3, then. 🤷‍♀️

PR #508, opened 2020-08-07
MessagingThreadHistory support pagination

nbhatia823:
**Jira:**

https://clever.atlassian.net/browse/SSOAP-2655?filter=25653

**Overview:**

chat_scroll

Changes to `MessagingThreadHistory` component:
1. Add ref to container. This meant converting the container to a div instead of a FlexBox, because I didnt want to add an extra wrapper div when the FlexBox is just a div with CSS. I've added the CSS directly from the related components
2. Add an index field to the `MessageData`. Array indices aren't a reliable way to provide keys 
3. Change effect of scrolling down 1) to not rely on messages length, as when we load older messages, we dont want the scroll to happen and 2) `useLayoutEffect` so the scroll happens before the DOM gets painted, preventing potential flickering
4. Add scroll on own message send, or if very close to the bottom of the container

NOTE - on mobile (or at least student/teacher portal), it will scroll on all new messages. This is because we use a completely different way of laying the page out, and scroll works very differently on fixed pages and elements interact with the dom differently. 

I think this is fine for now, because messagePagination wont work on mobile yet either for the same reason. So I think for now its better to scroll on all new messages on mobile than to not scroll if the user is at the bottom of the thread and they get a new message.

mobile_scroll

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingThreadHistory/MessagingThreadHistory.tsx
+++ b/src/MessagingThreadHistory/MessagingThreadHistory.tsx
@@ -1,10 +1,10 @@
 import * as React from "react";
-import { useEffect, useRef } from "react";
+import { useLayoutEffect, useRef } from "react";
 import * as classNames from "classnames";
 import * as moment from "moment";
-
-import { FlexBox } from "../index";
 import { MessageMetadata } from "./MessageMetadata";
+import { cssClass as FlexBoxCSS } from "../flex/FlexBox";
+import { cssClass as FlexItemCSS } from "../flex/FlexItem";
 
 import "./MessagingThreadHistory.less";
 
@@ -19,6 +19,7 @@ export interface MessageData {
   placement: "left" | "right" | "center";
   timestamp?: Date;
   content: React.ReactNode;
+  index?: number;
 }
 
 interface Props {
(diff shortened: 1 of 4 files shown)


src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -27,30 +28,69 @@ interface Props {
...
-  //  on messages themselves (new reference each render) nor on message count alone (different
-  //  threads can have the same message count). Instead, need the threadID AND message count.
-  useEffect(() => {
-    if (messages.length > 0) {
-      lastMessageRef.current.scrollIntoView();
-    }
-  }, [messages.length, threadID]);
+const SCROLL_BUFFER = 200;
 
+// Always returns true with current mobile strategy of scrolling the entire thread history
```
nbhatia823, line 33: This isn't ideal, but with how different our mobile and desktop layouts are, we are going to need to figure out how we want to approach things like scrolling in a uniform way.

src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -19,6 +19,7 @@ export interface MessageData {
   placement: "left" | "right" | "center";
   timestamp?: Date;
   content: React.ReactNode;
+  index?: number;
```
adamvictorclever: Just for my understanding, the intent is that we'd pass in the `index` on the Twilio message object? I have't read the documentation, but I see Twilio's message object does have an `index` field so I'd assume that meets our needs.

nbhatia823: Yes thats exactly right! Done [here](https://github.com/Clever/launchpad/pull/3503/commits/1b3133a7de713ae02883054edcf3f51878aa5352#diff-5f76153f13f1f361d2dff59c57a95a7aR32)

PR #510, opened 2020-08-11
DISC-1795 Allow passing of aria-label to Button component

rzhang-clever:
**Jira:**
https://clever.atlassian.net/browse/DISC-1795

**Overview:**
This PR allows consumers of the Button component to pass an aria-label prop that doesn't get overwritten by the existing aria-label logic. No changes to the Button interface, just fixes the logic

**Testing:**
Tested with locally-running launchpad

new aria-label passes properly
Screen Shot 2020-08-25 at 12 21 38 PM

existing aria-labels unaffected
Screen Shot 2020-08-25 at 12 21 58 PM

### DIFF
--- a/src/Button/Button.tsx
+++ b/src/Button/Button.tsx
@@ -19,6 +19,7 @@ export interface Props {
   submit?: boolean;
   style?: React.CSSProperties;
   underlined?: boolean;
+  ariaLabel?: string;
   [additionalProp: string]: any;
 }
 
@@ -53,6 +54,7 @@ const propTypes = {
   submit: PropTypes.bool,
   style: PropTypes.object,
   underlined: PropTypes.bool,
+  ariaLabel: PropTypes.string,
 };
 
 const defaultProps = {
@@ -84,6 +86,7 @@ export class Button extends React.PureComponent<Props> {
 
   render() {
     const {
+      ariaLabel,
       className,
       disabled,
       href,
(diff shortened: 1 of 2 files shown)


src/Button/Button.tsx
```diff
@@ -123,6 +123,7 @@ export class Button extends React.PureComponent<Props> {
       // use <button>s for all disabled links and things with no href prop (buttons)
       return (
         <button
+          aria-label={typeof value === "string" ? (value as string) : null}
           {...additionalProps}
```
dxu429, line 127: Should we put other props before this as well (other than `classes`)?

rzhang-clever: I thought about it but I figured it could mess with some things and just wanted to avoid that risk. For example the type prop on the dewey Button accepts linkPlain, primary, secondary, etc., but here in the component when returning an html button type represents the html button type field and not the dewey Button type field

dxu429: I'm wondering if we should just add a new optional prop `ariaLabel`. Something like:
```
const { ... ariaLabel ... } = this.props;
...
const aria = ariaLabel ?? typeof value === "string" ? (value as string) : null;
...
<Button
  {...additionalProps}
  ..
  aria-label={aria}
/>


```

This way we can keep the pattern where we spread `additionalProps` on top. What do you think?

rzhang-clever: I think I could go either way - Do you feel strongly about it?

dxu429: It seems a bit cleaner imo

rzhang-clever: Yeah sounds good 👍

rzhang-clever:
Didn't mean to rerequest just yet. Feel free to ignore this for now, and this ticket will be handed off to Mike

src/Button/Button.tsx
```diff
@@ -119,11 +122,14 @@ export class Button extends React.PureComponent<Props> {
       underlined ? "Button--linkUnderlined" : "",
     );
 
+    const aria = ariaLabel || (typeof value === "string" ? (value as string) : null);
```
wardmike, line 125: using `||` here instead of `??` because it seems `??` is not supported by all browsers

PR #516, opened 2020-08-26
Tooltip: making all tooltips focusable for accessibility

wardmike:
**Jira:** [DISC-1801](https://clever.atlassian.net/browse/DISC-1801)

**Overview:**

making all tooltips focusable by tab for accessibility

**Screenshots/GIFs:**
tested in dewey
tab-index-3

tested in launchpad
tab-index-1
tab-index-2

hover still works
hover

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/Tooltip/Tooltip.tsx
+++ b/src/Tooltip/Tooltip.tsx
@@ -11,7 +11,7 @@ import { Values } from "../utils/types";
 import "./Tooltip.less";
 
 export interface Props {
-  children: React.ReactNode;
+  children: React.ReactElement;
   className?: string;
   clickTrigger?: boolean;
   content: React.ReactNode;
@@ -59,7 +59,7 @@ const defaultProps = {
 
 export const cssClass = {
   CONTENT: "Tooltip--content",
-
+  FOCUSABLE_TRIGGER: "Tooltip--focusable-trigger",
   align: (textAlign) => `Tooltip--content--${textAlign}`,
 };
 
@@ -104,6 +104,8 @@ export default class Tooltip extends React.Component<Props> {
       </BootstrapTooltip>
     );
 
+    const child = React.Children.only(children);
+
     return (
       <OverlayTrigger
         delayShow={delayMs}
@@ -113,7 +115,12 @@ export default class Tooltip extends React.Component<Props> {
         rootClose
         trigger={hide ? [] : ["focus", clickTrigger ? "click" : "hover"]}
       >
-        {children}
+        {React.cloneElement(child, {
+          tabIndex: 0,
+          "aria-describedby": this.id,
+          className: cssClass.FOCUSABLE_TRIGGER,
+          ...child.props,
+        })}
       </OverlayTrigger>
     );
   }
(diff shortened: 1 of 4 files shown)


dxu429:
Focus might already be enabled via the OverlayTrigger. Would adding a tabIndex to the OverlayTrigger make it tab-able/focusable?

wardmike:
> Focus is already be enabled via the OverlayTrigger. Would adding a tabIndex to the OverlayTrigger make it tab-able/focusable?

@dxu429 i tried that but it didn't work 😕  all existing instances of the focusable OverlayTrigger used the span w/ tabIndex, and that's the only way i could figure out how to do it

dxu429:
Ahh got it, I'm guessing the OverlayTrigger doesn't produce a focusable DOM element then, is that right? And thats why we need to wrap the children with a focus-able element here?

dxu429:
I'm not sure if this makes the tooltip accessible though right? Does the actual popover need to be tab-able as well & have an aria-label?

wardmike:
> I'm not sure if this makes the tooltip accessible though right? Does the actual popover need to be tab-able as well & have an aria-label?

@dxu429 i think this is accessible? because using tab to focus on the question mark symbol adds the tooltip to the DOM, and the HTML in the tooltip is accessible
Screen Shot 2020-08-27 at 2 42 11 PM

the tooltip renders at the end of the DOM (which is in OverlayTrigger & not something we can control), but there is an `aria-label` to let the user know which tooltip describes it
Screen Shot 2020-08-27 at 2 50 11 PM

unfortunately, it's not possible to add an `aria-label` with the actual content of the tooltip because tooltips allow HTML. we'd have to change them to only allow strings if we wanted to put the content into an `aria-label`

dxu429:
re: `aria-label` We don't need to make any change in the component for this. The user should just add that to the children of Tooltip, doesn't need to be handled by the Tooltip component

dxu429:
From discussion on Slack:
An alternative approach that doesn't wrap `children` with another element would be to use `React.Children.<toArray | map>` with `React.cloneElement` and add `tabIndex` to each child, making each child element tab-able and focusable. 

```
{React.Children.map(children, child => {
      if(!React.isValidElement(child)) {
        return child;
      }
      return React.cloneElement(child, { tabIndex: 0 , ...child.props}); // not sure if ...child.props is necessary but in case users don't want the element to be focusable?
})}
```

src/Tooltip/Tooltip.tsx
```diff
@@ -113,7 +114,13 @@ export default class Tooltip extends React.Component<Props> {
         rootClose
         trigger={hide ? [] : ["focus", clickTrigger ? "click" : "hover"]}
       >
-        {children}
+        <span
+          className={cssClass.FOCUSABLE_TRIGGER}
+          onMouseDown={e => e.preventDefault()}
+          tabIndex={0}
```
dxu429, line 120: Can we also add `aria-describedby={this.id}` here?

wardmike: added! thanks for catching that!

wardmike:
@dxu429 implemented & tested your changes. can you re-review when you get a chance? 😁

dxu429 (approved):

PR #521, opened 2020-08-31
Reapply pagination changes

nbhatia823:
**Jira:**
https://clever.atlassian.net/browse/SSOAP-2655

**Overview:**
This reapplies this pr - https://github.com/Clever/components/pull/508
It allows for paging for MessagingThreadHistory

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingThreadHistory/MessagingThreadHistory.tsx
+++ b/src/MessagingThreadHistory/MessagingThreadHistory.tsx
@@ -1,10 +1,10 @@
 import * as React from "react";
-import { useEffect, useRef } from "react";
+import { useLayoutEffect, useRef } from "react";
 import * as classNames from "classnames";
 import * as moment from "moment";
-
-import { FlexBox } from "../index";
 import { MessageMetadata } from "./MessageMetadata";
+import { cssClass as FlexBoxCSS } from "../flex/FlexBox";
+import { cssClass as FlexItemCSS } from "../flex/FlexItem";
 
 import "./MessagingThreadHistory.less";
 
@@ -19,6 +19,7 @@ export interface MessageData {
   placement: "left" | "right" | "center";
   timestamp?: Date;
   content: React.ReactNode;
+  index: number;
 }
 
 interface Props {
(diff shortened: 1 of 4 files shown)


src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -27,30 +28,69 @@ interface Props {
...
-    </FlexBox>
+    ref &&
+    ref.current &&
+    ref.current.scrollTop + SCROLL_BUFFER >= ref.current.scrollHeight - ref.current.clientHeight
   );
-};
+}
+
+function isOwnMessage(message: MessageData) {
+  return message.placement === "right";
```
adamvictorclever, line 43: Not at all blocking, but thinking aloud here... 

This check feels kind of wrong to me, it feels like we're mixing levels of abstraction in this component. Up to this point, this component is kind of a dumb component that just renders the messages like it's told to. This component isn't aware who posted a message, it just knows that something told it to render a message to the `right` so it's going to do that. Instead, we have our higher level component `Chat` that's aware of who posted which message and can decide where to place it based on that.

In terms of what this could mean for us going forward, it's possible that `MessagingThreadHistory` is just at the wrong level of abstraction and it *should* be aware of this stuff and `Chat` maybe doesn't need to handle that. 

Alternatively, maybe this component was at the right level of abstraction and the logic added in this PR belongs in `Chat`. There's a very pragmatic case to be made that this is the wrong answer - `Chat` exists separately in `launchpad` and `family-portal` and we'd rather keep the logic shared. Maybe the answer to that is that some or all of `Chat` should be shared as well.

These are tough questions and I don't have an answer right now. I definitely think sticking with what we have is fine for now. That said, I know I've mentioned a few times that I'd like to be able to share some of the higher level and layouting components between `launchpad` and the `family-portal` more easily. If we do get to that, it would be nice to revisit this and see if we want to change our assumptions about how the components should work. I'd also love to hear any thoughts you have on these questions!

nbhatia823: Thank you for sharing these thoughts! I think they are very important and highlight an interesting problem.

I agree with you that it does feel like we are making this component more aware. I would suggest that the component gained its first taste of awareness when we added the initial "scroll to new message" behavior. It seems with the current approach the `chat` should be acting as a controller for the dumb UI components. Drawing clear lines for what these components should and shouldn't do seems like a great idea.

And I also agree that we should figure out the role of each of these components and confirm the purpose of having the levels of abstraction. It seems we should be sharing business logic across the `portals`, and at an absolute minimum between the `Student` and `Teacher` portals, so it might be worth discussing the place of the dumb components vs the higher level components, if there should indeed be a distinction. I do think there should be a distinction, but I think this could warrant another exploration of whether the components should live in their own messaging components repo (similar to the quick discussion of whether we should allow breaking changes to messaging components, slightly separating them from the rest of Dewey). 

I would be excited for us to all have a conversation about how we want to handle this going forward, because it would allow us to speed up iteration and break down some of the walls between our teams' work.

PR #523, opened 2020-09-10
Linkify messages

nbhatia823:
**Jira:**
https://clever.atlassian.net/browse/SSOAP-2687
Also addresses bug bash issue: 
https://clever.atlassian.net/browse/SSOAP-2664

> Long message (such as url) without breaks is cut off in message bubble. In the picture, the url continues on the second line beyond the cutoff point; the first line has a breakpoint because of the `-` (also true on teacher mobile, student desktop, and student mobile)

**Overview:**

Allows for links in messaging bubbles. We had used the `react-linkify` package in `RichText` so I updated it to the newest version which allows for more customization. We only want to link URLs in messages that have the protocol (i.e. https..) and we want to open in a new tab. 

This PR also fixes a bug that would cause a very long string to go off the screen instead of wrapping.

**Screenshots/GIFs:**

Screen Shot 2020-09-10 at 1 23 05 PM
Screen Shot 2020-09-10 at 1 23 02 PM

New wrapping behavior - 

Screen Shot 2020-09-10 at 1 22 53 PM

Old wrapping behavior -

Screen Shot 2020-09-10 at 1 22 31 PM

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


nbhatia823:
Not sure why the linter is complaining since none of the errors are from the files I touched but I'll fix a few errors from other files (based on CircleCI).

src/MessagingBubble/MessagingBubble.less
```diff
@@ -22,6 +23,14 @@
   white-space: pre-wrap;
 }
 
+.MessagingBubble--Link,
+.MessagingBubble--Link:focus,
+.MessagingBubble--Link:hover {
```
jonah-clever, line 28: not sure if the `:focus` and `:hover` are necessary if the CSS is the same as for `.MessagingBubble--Link`? But maybe so

jonah-clever (approved):
One comment, looks good tho

jonah-clever:
oh and the failing tests

src/MessagingBubble/MessagingBubble.less
nbhatia823: Without these it will still show the default link text-decoration on hover/focus (light blue, no underline) - I read that if you were to specify the element as well like `.MessagingBubble--Link **a**` then I wouldn't need to include the `:focus` and `:hover`, but I thought this was a bit clearer vs. specifying the `a` tag.

nbhatia823:
@adamvictorclever thanks for the explanation! I didn't notice at a quick glance that there was a new linting error for `MessagingBubble`. So if I don't want to fix errors in other files right now I'm okay to leave the error limit where it's at? 😝

PR #526, opened 2020-09-15
Add span wrapper to RichText to match old behavior

nbhatia823:
**Overview:**

I updated the `RichText` component in https://github.com/Clever/components/pulls?q=is%3Apr+is%3Aclosed because I updated the underlying library `Linkify` from v 0.2.x to 1.0.0. It seems that in the older version, Linkify would wrap any output in an extra `span` element -

OLD Linkify v0.2.x
Screen Shot 2020-09-14 at 5 58 45 PM

NEW Linkify v1.0.0
Screen Shot 2020-09-14 at 5 58 34 PM

We rely on that `span` being there, because otherwise the `RichText` output will take the parent element's styling. In the example above it results in each child span element being treated as a child of the outer `Flexbox class=CollectionPageHeader--teacherAnnouncementBody`, instead of it being a single child span.

New Linkify without added span - 

Screen Shot 2020-09-14 at 5 58 16 PM

New Linkify with added span - 

Screen Shot 2020-09-14 at 6 27 59 PM

This PR adds in that `span` as the wrapper for any `RichText` output. 

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/RichText/RichText.tsx
+++ b/src/RichText/RichText.tsx
@@ -21,12 +21,14 @@ export default function RichText(props: Props) {
 
   return (
     <Linkify componentDecorator={componentDecorator}>
-      {lines.map((item, index) => (
-        <span key={index}>
-          {item}
-          {index < lines.length - 1 && <br />}
-        </span>
-      ))}
+      <span>
+        {lines.map((item, index) => (
+          <span key={index}>
+            {item}
+            {index < lines.length - 1 && <br />}
+          </span>
+        ))}
+      </span>
     </Linkify>
   );
 }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.54.0",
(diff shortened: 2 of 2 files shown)


src/RichText/RichText.tsx
```diff
@@ -21,12 +21,14 @@ export default function RichText(props: Props) {
 
   return (
     <Linkify componentDecorator={componentDecorator}>
-      {lines.map((item, index) => (
-        <span key={index}>
-          {item}
-          {index < lines.length - 1 && <br />}
-        </span>
-      ))}
+      <span>
```
nbhatia823, line 24: Only change is adding this `span`

lubalexuan (approved):
🌯

PR #529, opened 2020-09-16
Searching for the magic scroll

nbhatia823:
**Jira:**
https://clever.atlassian.net/browse/SSOAP-2770

**Overview:**
Ideally we want to scroll to a new message if a user is at the bottom of the thread when they receive a new message. The current way to do this doesn't work if a user gets sent a long message. With this fix, we save if the user was at the bottom of the thread before the new message comes in, instead of after the message comes. 

That is preferred, because if we calculate after the message comes in, the new message gets inserted at the bottom of the container and only if the total size is within `SCROLL_BUFFER` do we scroll to the bottom of the page, but we don't know how big that message will be. Therefore, its better to do the checks beforehand! 

I've also reorganized the code a bit and added some comments.

**Screenshots/GIFs:**

GIF - https://share.getcloudapp.com/Qwu0pYvJ
New behavior on left, old behavior on right

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
~  - [ ] Updated docs~
~  - [ ] Bumped version in `package.json`~

### DIFF
--- a/src/MessagingThreadHistory/MessagingThreadHistory.tsx
+++ b/src/MessagingThreadHistory/MessagingThreadHistory.tsx
@@ -31,7 +31,7 @@ interface Props {
 const SCROLL_BUFFER = 200;
 
 // Always returns true with current mobile strategy of scrolling the entire thread history
-function isScrolledToBottom(ref: React.MutableRefObject<HTMLDivElement>) {
+function calcIsScrolledToBottom(ref: React.MutableRefObject<HTMLDivElement>): boolean {
   return (
     ref &&
     ref.current &&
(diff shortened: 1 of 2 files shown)


src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -48,11 +48,21 @@ export const MessagingThreadHistory = React.forwardRef(
...
+
+    // Maintains a reference as to what the last (highest) messages index was on the last render
     const lastMessageIndex = useRef(
       messages.length > 0 ? messages[messages.length - 1].index : null,
     );
-    const messagesWithDividers = _interleaveMessagesWithDividers(messages, lastMessageRef);
+
+    // Determines whether the user was scrolled to the bottom before we potentially
+    //  insert new messages at _interleaveMessagesWithDividers()
+    const isScrolledToBottom = calcIsScrolledToBottom(containerRef);
```
nbhatia823, line 63: The main difference is I calculate this value before `useLayoutEffect`, as that gets fired after the whole render function is run and the new messages are appended to the bottom of the thread history.

TheRealBenjiLi (approved):
🌮

PR #534, opened 2020-09-18
Table2beta Select all button selects all rows

jonah-clever:
**Overview:**

Want "Select All" checkbox to select all rows, not only the current page. 

**However** this isn't really feasible with lazy data loading option. So, if data is passed in as the `data` prop, "Select All" checkbox will select all rows. If data is loaded with lazy loading, Select All will only select the current page

Essentially, lazy loading works by loading one page of data at a time, when that page is accessed. To select all rows, the component would have to query for all pages one by one, which defeats the purpose of lazy loading. So if the consumer wants this feature, they should load all data in one query outside of the component and pass it in as `data`.

**Screenshots/GIFs:**

Screen Shot 2020-09-18 at 2 21 13 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/Table2BetaView.jsx
+++ b/docs/components/Table2BetaView.jsx
@@ -412,9 +412,7 @@ export default class Table2BetaView extends React.PureComponent {
               name: "lazy",
               type: "boolean",
               description:
-                'Puts the table in "lazy" mode - data is fetched ' +
-                'from the getData function rather than needing the whole "data" ' +
-                "array up front.",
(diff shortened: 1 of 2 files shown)


TheRealBenjiLi (approved):
🌮

PR #539, opened 2020-09-23
Add id to input component of TextInput (a11y)

dengjulie:
**Jira:**
https://clever.atlassian.net/browse/FAMBAM-539

Notes from accessibility audit:
image

**Overview:**
This PR adds an id to to the input component of TextInput to match that of the "for" attribute of the label component. We're learning from the most recent accessibility audit to family-portal that these should match for blind people to determine which label goes with which form element. Rather than making the change just in fampo, this seems applicable to all TextInputs. 

This change is compatible even if prior uses of TextInput pass in an id prop as that will take precedence over the default `this.props.name` and in both cases, the same value will be applied to the two tag attributes as desired. 

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -205,7 +205,8 @@ export class TextInput extends React.Component<Props, State> {
 
     // (a11y) The 'for' attribute of the label tag and the 'id' attribute
     // of the input tag should match so screen readers can determine which
-    // label goes with which form element
+    // label goes with which form element. If an 'id' isn't specified, we
+    // use 'name' instead for this purpose.
     const id = additionalProps.id || this.props.name;
 
     return (
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -219,6 +219,7 @@ export class TextInput extends React.Component<Props, State> {
         </div>
         <input
           {...additionalProps}
+          id={this.props.name}
           autoComplete={autoComplete}
           className="TextInput--input"
           disabled={this.props.disabled}
--- a/src/TextInput/TextInput.tsx
+++ b/src/TextInput/TextInput.tsx
@@ -203,6 +203,11 @@ export class TextInput extends React.Component<Props, State> {
       propTypes,
     ) as (keyof typeof propTypes)[]);
 
+    // (a11y) The 'for' attribute of the label tag and the 'id' attribute
+    // of the input tag should match so screen readers can determine which
+    // label goes with which form element
+    const id = additionalProps.id || this.props.name;
+
     return (
       <div
         className={classnames(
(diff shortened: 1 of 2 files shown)


lbanner3:
My initial thoughts are that this should be fine because we very rarely lean on using the id property on TextInputs if at all to be honest.

src/TextInput/TextInput.tsx
```diff
@@ -219,6 +219,7 @@ export class TextInput extends React.Component<Props, State> {
         </div>
         <input
           {...additionalProps}
+          id={this.props.name}
```
arsalansufi, line 222: > My one concern is if the TextInput component is currently being used with an id prop passed in as an additional prop, as this would override it.

Great callout! I definitely think that, in an ideal world, this component should always be accessible. And one way to accomplish that is to always set `id` to `this.props.name`.

At the same time, I think that engineers are pretty used to being able to specify `id` for any component and having that stick. And in fact, we have some end-to-end tests that look for elements with specific IDs, so I do still think that we should use the `id` when specified.

Since we ultimately just need the label's `for` attribute to match the input's `id` attribute, another option might be to change what we set the label `htmlFor` prop to, e.g.

```
<label htmlFor={this.props.id || this.props.name}>
<input id={this.props.id || this.props.name}>
```

I think that this works for all cases, e.g. `id` and no `name`, `name` and no `id`, and both.

It's also worth adding a comment to the code noting the need for these to match just to avoid someone accidentally undoing this.

arsalansufi (commented):
A thought for you in response to your question!

src/TextInput/TextInput.tsx
dengjulie: Great suggestion, thanks Arsalan!

dengjulie:
Made an update based on Arsalan's suggestion and updated the PR description as well!

src/TextInput/TextInput.tsx
```diff
@@ -203,6 +203,11 @@ export class TextInput extends React.Component<Props, State> {
       propTypes,
     ) as (keyof typeof propTypes)[]);
 
+    // (a11y) The 'for' attribute of the label tag and the 'id' attribute
+    // of the input tag should match so screen readers can determine which
+    // label goes with which form element
```
arsalansufi, line 208: I might add something like the following to this comment, to make the intention of the line of code below clearer.

> If an 'id' isn't specified, we use 'name' instead for this purpose.

arsalansufi (approved):
One small suggestion for the comment, then LGTM!

PR #545, opened 2020-09-30
Fambam 541 MessagingInput accessibility and translation support

adamvictorclever:
**Jira:** https://clever.atlassian.net/browse/FAMBAM-541 (see ticket for design instructions/mock)

**Overview:** This PR makes a few changes to make the MessagingInput more accessible and also support translations in the FamPo.

The first commit fixes some accessibility support for `TextArea`. Previously, the underlying `textarea` would never have an `id`, which meant that labels could never properly be associated with them. Previously we used `htmlFor` with the `name` but, I don't believe that actually works. 

Instead, we make a similar change to the one Julie recently made and support using the `name` as the `id` by default. I also added support for explicitly setting an `id` since `name` and `id` do have slightly different requirements (`id` needs to be unique, `name` doesn't). I don't think there will be many cases where explicitly setting an `id` is necessary though.

We also make a change to only include the `TextArea's` internal `label` if there's any text to display. This avoids an issue where the accessibility checker complained about the underlying `textarea` in `MessagingInput` having two labels.

The second commit makes the accessibility changes for `MessagingInput` by adding an external `label` and removing the placeholder text.

The third commit makes it possible to override the text used for the label and send button. This is a temporary measure to be used in the FamPo so that we can pass translations through. In the future, we'd like to support translations inside `clever-components` so that passing them through isn't necessary.

**Screenshots/GIFs:**
Screen Shot 2020-09-30 at 1 01 20 PM

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/TextArea/TextArea.tsx
+++ b/src/TextArea/TextArea.tsx
@@ -16,6 +16,7 @@ export interface Props {
   maxLength?: number;
   minLength?: number;
   name: string;
+  id?: string;
   onBlur?: React.FocusEventHandler<HTMLTextAreaElement>;
   onChange?: React.ChangeEventHandler<HTMLTextAreaElement>;
   onFocus?: React.FocusEventHandler<HTMLTextAreaElement>;
@@ -207,10 +208,17 @@ export class TextArea extends React.Component<Props, State> {
     // note on the upper right corner
     const inputNote = this.renderNote(errorMessage);
 
+    // (a11y) The 'for' attribute of the label tag and the 'id' attribute
+    // of the textarea tag should match so screen readers can determine which
+    // label goes with which form element. If an 'id' isn't specified, we
+    // use 'name' instead for this purpose.
+    const id = this.props.id || this.props.name;
+
     const textAreaProps = {
       ["aria-invalid"]: !valid,
       className: "TextArea--input",
       disabled: this.props.disabled,
+      id,
       maxLength: this.props.maxLength,
       minLength: this.props.minLength,
       name: this.props.name,
@@ -251,9 +259,11 @@ export class TextArea extends React.Component<Props, State> {
         )}
       >
         <div className="TextArea--infoRow">
-          <label className="TextArea--label" htmlFor={this.props.name}>
-            {this.props.label}
-          </label>
+          {this.props.label && (
+            <label className="TextArea--label" htmlFor={id}>
+              {this.props.label}
+            </label>
+          )}
           <span aria-live="polite">{inputNote}</span>
         </div>
         {textarea}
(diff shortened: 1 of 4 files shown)


src/TextArea/TextArea.tsx
```diff
@@ -251,9 +259,11 @@ export class TextArea extends React.Component<Props, State> {
         )}
       >
         <div className="TextArea--infoRow">
-          <label className="TextArea--label" htmlFor={this.props.name}>
-            {this.props.label}
-          </label>
+          {this.props.label && (
```
arsalansufi, line 262: I wonder how many TextArea's exist today with neither a `label` prop specified nor a manually configured label (like in the MessagingInput). It's probably very few. But worth noting that those may end up completely label-less after this is rolled out and imported in repos.

One alternative would be to expose a `noLabel` prop, though I'm not a huge fan. I might just do a quick pass through launchpad's TextArea's to make sure all specify a `label` prop. I know FamPo is okay.

arsalansufi (commented):
This looks great! I'll let Julie give the final LGTM

src/TextArea/TextArea.tsx
arsalansufi: Okay LOL, there are quite a few in launchpad. Literally the first two that I looked at haha.

- https://github.com/Clever/launchpad/blob/03be51877180c72625a97be086e7ec39ab9fbd46/src/ui/components/ClassroomActivityExperiment/ClassroomActivityExperiment.tsx#L90-L100
- https://github.com/Clever/launchpad/blob/03be51877180c72625a97be086e7ec39ab9fbd46/src/ui/components/CollectionPageHeader/EditTeacherAnnouncementModal.tsx#L80-L89

adamvictorclever: Hmm, I think I'm not understanding how this change makes those cases any worse. It is true that there may now be some cases where the `<label>` is no longer included, but is there any benefit to rendering a label with no text? It's not clear to me what makes that preferable over no label.

arsalansufi: Ah of course 🤦‍♂️

You're absolutely right. Missed the piece that we're just rendering an empty label for those today.

Ignore me :hide:

arsalansufi (approved):
I know Julie hasn't had a chance to look at this yet, but LGTM!

src/TextArea/TextArea.tsx
adamvictorclever: Haha no worries, glad we're on the same page 😄

PR #548, opened 2020-10-01
Fix timestamp positioning when no content preview is provided

afumagalli:
**Overview:**

We are implementing progressive loading of the MessagesPage which will mean for all threads we will no longer have a message to render a preview in the thread list. However we will have a timestamp of the last message, and we want to show that without a preview. But the timestamp positioning was incorrect in such a case. This fixes it.

**Screenshots/GIFs:**

Before:
Screen Shot 2020-10-01 at 2 20 27 PM

After:
Screen Shot 2020-10-01 at 2 33 56 PM

With an indicator:
Screen Shot 2020-10-01 at 2 34 05 PM
Screen Shot 2020-10-01 at 2 35 09 PM

**Testing:**
- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs

- After merging:

### DIFF
--- a/docs/components/MessagingThreadListItemView.jsx
+++ b/docs/components/MessagingThreadListItemView.jsx
@@ -72,8 +72,11 @@ export default class MessagingThreadListItemView extends React.PureComponent {
             <MessagingThreadListItem
               icon={<Icon name="smiley-face" />}
               title="Smiley Dude"
+              status={status}
               hasDraft={hasDraft}
+              isRead={isRead}
               selected={selected}
+              timestamp={new Date()}
             />
           </ExampleCode>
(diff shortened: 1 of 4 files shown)


nbhatia823 (commented):
I'm wondering if we should move the indicator to the left of the timestamp instead of the right for the inline case? That way the times will be aligned in the `threadList` instead of shifted for any inline items with indicators.

afumagalli:
we just got designs for this and the indicator should be below timestamp so changes coming

afumagalli:
closing because its not needed anymore

PR #553, opened 2020-10-08
Table2Beta: Bug Fix

TheRealBenjiLi:
**Jira:**

**Overview:**
There was an infinite loop problem when setState is triggered in render. This PR fixes that for Table2Beta.

**Screenshots/GIFs:**

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] Firefox

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/Table2Beta/Table.tsx
+++ b/src/Table2Beta/Table.tsx
@@ -544,6 +544,7 @@ export class Table2Beta extends React.Component<Props, State> {
       <>
         {selectable && (
           <SelectedRowsHeader
+            className={className}
             selectedRows={selectedRows}
             contentType={selectedRowsHeaderContentType}
             actions={selectedRowsHeaderActions}
--- a/src/Table2Beta/Table.tsx
+++ b/src/Table2Beta/Table.tsx
@@ -208,6 +208,24 @@ export class Table2Beta extends React.Component<Props, State> {
     }
   }
 
+  componentDidUpdate() {
+    const { selectedRows } = this.state;
+    const { allRows } = this._getDisplayedData();
+
+    // This occurs when a new external filter is applied.
+    // We want selectedRows to change if the filter affects the selection
+    const newSelectedRows = new Set(selectedRows);
+    newSelectedRows.forEach((item) => {
+      if (!allRows.includes(item)) {
+        newSelectedRows.delete(item);
+      }
+    });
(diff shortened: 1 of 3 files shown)


src/Table2Beta/Table.tsx
```diff
@@ -508,8 +508,19 @@ export class Table2Beta extends React.Component<Props, State> {
...
+    // This occurs when a new external filter is applied.
+    // We want selectedRows to change if the filter affects the selection
+    const newSelectedRows = new Set(selectedRows);
+    newSelectedRows.forEach((item) => {
+      if (!allRows.includes(item)) {
+        newSelectedRows.delete(item);
+      }
+    });
+    if (newSelectedRows.size !== selectedRows.size) {
+      this.setState({ selectedRows: newSelectedRows });
```
meliaj, line 521: Is there a chance the size is the same but the sets are different?

TheRealBenjiLi: Nope! Line `514` declares newSelectedRows as a copy of selectedRows, so they have the same data. From there, newSelectedRows can only get smaller.

src/Table2Beta/Table.tsx
```diff
@@ -508,8 +508,19 @@ export class Table2Beta extends React.Component<Props, State> {
...
+    }
```
arsalansufi, line 522: A quick note! A `setState` operation like this probably belongs in `componentDidMount` + `componentDidUpdate`. Calling `setState` within `render` is not recommended as `render` is not supposed to have any side effects. The only case where calling `setState` within `render` makes sense is, for example, as part of an `onClick` handler:

```tsx
render() {
  return (
    <button onClick={() => this.setState({ clicked: true })>Click here</button>
  );
}
```

arsalansufi: One reason why `setState` is advised against in render is because it can cause infinite loops as you've seen :smile:

TheRealBenjiLi: Good feedback! I did not know `render` was supposed to be side effect free 👍

clcaelynn-clever (approved):
🍔 🚀

PR #556, opened 2020-10-09
Table2Beta: More Bug Fixes and Feedback

TheRealBenjiLi:
**Jira:**

https://youtu.be/2EwViQxSJJQ?t=19

**Overview:**
Moving the actions a little to the left, so they're lined up with the Action header. Also addressed a bug where, if you selected all rows in a filtered mode, and changed to a different filter, the SelectedRowsHeader would count it as selecting both all and no rows. 

**Screenshots/GIFs:**
Screen Shot 2020-10-09 at 3 17 25 PM

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] Firefox

**Roll Out:**
- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/Table2Beta/SelectedRowsHeader.tsx
+++ b/src/Table2Beta/SelectedRowsHeader.tsx
@@ -68,7 +68,7 @@ export default function SelectedRowsHeader({
               </div>
             </>
           )}
-          {allSelected && (
+          {rowsAreSelected && allSelected && (
             <div>{`All ${contentType.plural || "rows"} selected (${selectedRows.size})`}</div>
           )}
         </FlexItem>
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.62.2",
+  "version": "2.62.3",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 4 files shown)


src/Table2Beta/SelectedRowsHeader.tsx
```diff
@@ -68,7 +68,7 @@ export default function SelectedRowsHeader({
               </div>
             </>
           )}
-          {allSelected && (
+          {rowsAreSelected && allSelected && (
```
afumagalli, line 71: when would `rowsAreSelected = false` but `allSelected = true` - naming at least seems redundant, but I also am not familiar enough to know for sure

afumagalli (approved):

src/Table2Beta/SelectedRowsHeader.tsx
TheRealBenjiLi: Good question! This PR actually addresses a situation where `rowsAreSelected=false` but `allSelected = true`. To be fair, that scenario only occurred because I forgot to update `allSelected` 😛 
We use `allSelected` because SelectedRowsHeader does not know how many rows there are total. So we either 1) pass in the total number of rows available for selection, or 2) pass `allSelected`.

PR #550, opened 2020-10-02
Temporary prop overrides to allow translations to be passed in

prime-time:
**Jira:**
https://clever.atlassian.net/browse/FAMBAM-574

**Overview:**
This change introduces new props to allow us to override the "Show" and "Hide" button text. The change will be used for passing in translation strings, and it's only intended to be a temporary solution.

I'm open to thoughts on whether these props should be in the docs. I went ahead and added them and noted that they're intended to be temporary.

**Screenshots/GIFs:**
Screen Shot 2020-10-01 at 7 23 41 PM
Screen Shot 2020-10-01 at 7 23 50 PM

**Testing:**
- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE10

**Roll Out:**
- Before merging:
  - [x] Updated docs
  - [x] Bumped version in `package.json`
    - Breaking change?
      - If it is a beta component run `npm version minor`
      - If the component is not in beta run `npm version major`
    - New component or backward-compatible component feature change? Run `npm version minor`
    - Only changing documentation? All good. Skip this step.
  - After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [ ] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
        componentLink: "<COMPONENT LINK>",
        componentImg: "<COMPONENT LINK>.png",
        componentName: "<COMPONENT NAME>",
        componentImgAlt: "A <COMPONENT NAME> component",
      },
      ```
    - [ ] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:
  - [ ] Deployed updated docs (`make deploy-docs`)
  - [ ] Posted in #eng if I made a breaking change to a beta component

PR #560, opened 2020-10-20
Add Read Receipts to Messaging components

dengjulie:
**Jira:**
https://clever.atlassian.net/browse/FAMBAM-490

**Overview:**
This PR adds read receipts to the messaging components.

**Screenshots/GIFs:**
Design Mocks:
image

Mobile:
image

Web:
image

**Testing:**
- [ ] Unit tests
- Manual tests: (Tested by pointing local fampo at local clever-components server)
 - [x] Styling without read receipts remains the same
 - [x] Mobile
  - [x] Chrome
  - [x] Safari
  - [x] IE11

**Roll Out:**
- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/src/MessagingThreadHistory/MessageMetadata.tsx
+++ b/src/MessagingThreadHistory/MessageMetadata.tsx
@@ -12,21 +12,25 @@ interface Props {
   className?: string;
   placement: "left" | "right" | "center";
   timestamp?: Date;
+  showReadStatus?: boolean;
   children: React.ReactNode;
 }
 
 export const MessageMetadata: React.FC<
   Props & { ref?: React.Ref<HTMLDivElement> }
 > = React.forwardRef((props: Props, ref: React.Ref<HTMLDivElement>) => {
-  const { className, placement, timestamp, children } = props;
+  const { className, placement, timestamp, showReadStatus, children } = props;
   return (
-    <div ref={ref} className={classNames(cssClass(`Message--${placement}`), className)}>
-      {children}
-      {timestamp && (
-        <span className={cssClass(`Timestamp--${placement}`)}>
-          {_formatDateForTimestamp(timestamp)}
-        </span>
-      )}
+    <div className={cssClass("Message--container")}>
+      <div ref={ref} className={classNames(cssClass(`Message--${placement}`), className)}>
+        {children}
+        {timestamp && (
+          <span className={cssClass(`Timestamp--${placement}`)}>
+            {_formatDateForTimestamp(timestamp)}
+          </span>
+        )}
+      </div>
+      {showReadStatus && <div className={cssClass("ReadReceipt")}>Read</div>}
     </div>
   );
 });
--- a/src/MessagingThreadHistory/MessageMetadata.tsx
+++ b/src/MessagingThreadHistory/MessageMetadata.tsx
@@ -12,14 +12,14 @@ interface Props {
   className?: string;
   placement: "left" | "right" | "center";
   timestamp?: Date;
-  showReadStatus?: boolean;
+  readStatusText?: string;
   children: React.ReactNode;
 }
 
 export const MessageMetadata: React.FC<
   Props & { ref?: React.Ref<HTMLDivElement> }
 > = React.forwardRef((props: Props, ref: React.Ref<HTMLDivElement>) => {
-  const { className, placement, timestamp, showReadStatus, children } = props;
+  const { className, placement, timestamp, readStatusText, children } = props;
   return (
     <div ref={ref} className={cssClass("Message--container")}>
       <div className={classNames(cssClass(`Message--${placement}`), className)}>
@@ -30,7 +30,7 @@ export const MessageMetadata: React.FC<
           </span>
         )}
       </div>
-      {showReadStatus && <div className={cssClass("ReadReceipt")}>Read</div>}
+      {readStatusText && <div className={cssClass("ReadReceipt")}>{readStatusText}</div>}
     </div>
   );
 });
--- a/src/MessagingThreadHistory/MessageMetadata.tsx
+++ b/src/MessagingThreadHistory/MessageMetadata.tsx
@@ -21,8 +21,8 @@ export const MessageMetadata: React.FC<
 > = React.forwardRef((props: Props, ref: React.Ref<HTMLDivElement>) => {
   const { className, placement, timestamp, showReadStatus, children } = props;
   return (
-    <div className={cssClass("Message--container")}>
-      <div ref={ref} className={classNames(cssClass(`Message--${placement}`), className)}>
+    <div ref={ref} className={cssClass("Message--container")}>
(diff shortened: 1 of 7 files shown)


src/MessagingThreadHistory/MessageMetadata.tsx
```diff
@@ -12,21 +12,25 @@ interface Props {
...
+    <div className={cssClass("Message--container")}>
+      <div ref={ref} className={classNames(cssClass(`Message--${placement}`), className)}>
+        {children}
+        {timestamp && (
+          <span className={cssClass(`Timestamp--${placement}`)}>
+            {_formatDateForTimestamp(timestamp)}
+          </span>
+        )}
+      </div>
+      {showReadStatus && <div className={cssClass("ReadReceipt")}>Read</div>}
```
adamvictorclever, line 33: I think we're going to need some way for a consumer to specify the text to use here. Currently, we're able to translate the text in messaging components in the FamPo by passing the translation through as a prop. Ideally, we'd like to switch to translating in `clever-components` so that isn't necessary, but we're not at that point yet.

src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -20,12 +20,14 @@ export interface MessageData {
...
   content: React.ReactNode;
   index: number;
+  isRead?: boolean;
 }
 
 interface Props {
   className?: string;
   threadID: string;
   messages: MessageData[];
+  showReadStatus?: boolean;
```
adamvictorclever, line 30: Just curious how necessary this prop is. If the consumer doesn't want read status to be displayed, they could just omit the `isRead` prop on the `messages`.

src/MessagingThreadHistory/MessageMetadata.tsx
arsalansufi, line 33: Just noting real quick that we'll want to make the specific string potentially customizable so that we can pass in translations, e.g. `readStatusText={t("Read")}`

Here's a good example of what this might look like! https://github.com/Clever/components/pull/550

And we'll want to pass in translated strings in the FamPo PR

adamvictorclever (commented):
A couple small thoughts!

src/MessagingThreadHistory/MessagingThreadHistory.tsx
dengjulie: Ah, not super necessary. I think I initially wanted to make it difficult to have only a portion of the messages to show the read status but the consumer should probably make sure of that. Removing the field!

src/MessagingThreadHistory/MessageMetadata.tsx
dengjulie: Just changed it to a consumer passed text field! Thanks for the reminder!

dengjulie: Just changed it to a consumer passed text field! Thanks for the reminder!

src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -20,6 +20,7 @@ export interface MessageData {
   timestamp?: Date;
   content: React.ReactNode;
   index: number;
+  readStatusText?: string;
```
arsalansufi, line 23: I wonder if we should still have a `showReadStatus` boolean prop as you'd configured earlier. And the `readStatusText` prop should exist in addition to that.

When someone specifies `showReadStatus` without a `readStatusText`, we'll fallback to "Read".

While I like the idea of consolidating props, I view `readStatusText` as a temporary prop, to be dropped once we've updated Dewey to translate internally.

dengjulie: This might be awhile away, but I also think that having consumer passed text will help if ever in the future there is group messaging and we want teachers to see, "Seen by Kristen, James"

dengjulie: Also, just to double check do you mean the isRead value (within MessageData) or showReadStatus (within ThreadHistory)?

src/MessagingThreadHistory/MessageMetadata.tsx
afumagalli: not related to this PR, but chiming in that `switch to translating in clever-components` may not be the move since `clever-components` is a public repo and translation would seem to depend on Clever user data, no?

afumagalli:
a quick observation: I'm not a fan of showing `Read` on every message. other messagers (Messenger, iMessage, etc.) only show `Read` on the last read message. The rest is superfluous since we set threads as read to a certain point, we don't do this for individual messages.

dengjulie:
> I'm not a fan of showing Read on every message. other messagers (Messenger, iMessage, etc.) only show Read on the last read message. The rest is superfluous since we set threads as read to a certain point, we don't do this for individual messages.
More context: 
https://clever.atlassian.net/browse/FAMBAM-489
image

afumagalli:
hmmm... are we able to let the consumer make that choice then? because I don't know if I agree with that argument and believe we will want to change it later on

clcaelynn-clever:
Just a thought with regard to the group messaging future: Have we considered the Messenger style of using the Avatars themselves as the read indicator? Might scale better to multiple people, since even just first names tend to take up more space.

dengjulie:
> Just a thought with regard to the group messaging future: Have we considered the Messenger style of using the Avatars themselves as the read indicator? Might scale better to multiple people, since even just first names tend to take up more space.

That's a good note! For now, I think for now I'll leave it as a string prop and later when group messaging is implemented, we can switch the type to a react node, which would be a backward compatible change

dengjulie:
> hmmm... are we able to let the consumer make that choice then? because I don't know if I agree with that argument and believe we will want to change it later on

Offline discussion: This PR does allow for it to be a consumer choice so we can move forward with this PR and have that design discussion via slack

src/MessagingThreadHistory/MessagingThreadHistory.tsx
arsalansufi: > but I also think that having consumer passed text will help if ever in the future there is group messaging and we want teachers to see, "Seen by Kristen, James"

Great point! Given that `readStatusText` will provide permanent value beyond just the temporary value of translatability, I'm okay with us relying on it alone, without `showReadStatus`.

While this might make it marginally harder to guarantee consistency in default string choice (e.g. launchpad passes "Seen" and FamPo passes "Read"), I don't think that's a big deal.

> Also, just to double check do you mean the isRead value (within MessageData) or showReadStatus (within ThreadHistory)?

Ah! I was conflating the two together a bit. I was really just thinking conceptually about the prop(s) required to display a read receipt.

lbanner3 (approved):
🌮 👓 📖 🧾

afumagalli (approved):

PR #564, opened 2020-11-24
Fix build

prime-time:
**Overview:**
The build has been broken for the last few days, and this pull request is my attempt to fix it. The issue is that there are duplicate react type definitions being pulled in which are causing the TypeScript compilation to fail. This repo has v16 of the react typings as a devDependency, but v17 is also getting pulled in by type definitions for react-datepicker and react-linkify. I'm not entirely sure why those libraries are causing v17 of the react types to get pulled in because they specify `"*"` as the version they depend on. Moving @types/react-datepicker and @types/react-linkify from `dependencies` to `devDependencies` stops them from pulling in v17 of react types.

Only `@types` packages that appear in the exported declaration files should be in `dependencies`. @types/react-linkify doesn't appear in the type declarations exported by this repo, so it was fine to move from `dependencies` to `devDependencies`. @types/react-datepicker does appear in the generated `DatePicker.d.ts` file which means it can't be moved to `devDependencies`. In order to safely make @types/react-datepicker a devDependency, I changed the `DatePicker` types so that they no longer import types from react-datepicker. It actually returns the code to be almost the same as [it used to be](https://github.com/Clever/components/commit/40552ed27f0bbef3fda1e526bd830b731bbe2328) before we added types to the file. At some point we discussed the pros and cons of explicitly listing a subset of props we use when one of our components uses a component from another library, and we leaned towards listing a subset of props we plan to use. I went through all of our uses of `DatePicker` and the props listed in this change are the only ones we're using.

https://stackoverflow.com/questions/45176661/how-do-i-decide-whether-types-goes-into-dependencies-or-devdependencies

I tested this change by installing it in a couple of apps, and the only issue I ran into was some conflict with the lodash types. The lodash types appear in one of the exported declaration files, so we'd need to make some code change in order to make `@types/lodash` a devDependency. Bumping it to the latest version resolves the issue and is a smaller change, so I went with that.

**Screenshots/GIFs:**

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


clcaelynn-clever (approved):
👏 ✨ 🚀

PR #567, opened 2020-11-30
Ssoap 2846 alert heading

jonah-clever:
**Jira:** https://clever.atlassian.net/browse/SSOAP-2846

**Overview:**
H3 does not always match the top-level heading for AlertBox, which can cause accessibility issues. Adding optional `headingLevel` prop to set the `aria-level`, with default to 3 (so existing AlertBoxes see no change).

This will be used to fix one known instance of the accessibility issue in the Student Portal

Also removing redundant PropTypes

Bumped to 2.68 because I have another pending PR to merge that will bump to 2.67 which I will merge first, and then this one

**Testing:**

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
  
- After merging:

### DIFF
--- a/src/AlertBox/AlertBox.tsx
+++ b/src/AlertBox/AlertBox.tsx
@@ -18,6 +18,7 @@ export interface Props {
   type?: AlertBoxType;
   isClosable?: boolean;
   onClose?: () => void;
+  headingLevel?: 1 | 2 | 3 | 4 | 5 | 6;
 }
 
 const cssClass = {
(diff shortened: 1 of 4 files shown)


src/AlertBox/AlertBox.tsx
```diff
@@ -30,14 +30,6 @@ const cssClass = {
   CHILDREN: "AlertBox--children",
 };
 
-const propTypes = {
```
nbhatia823, line 33: 🔪

nbhatia823 (approved):
Nice!

PR #586, opened 2021-02-19
ID-48 TextInput2 birthday

johnhuangclever:
**Overview:**
Basic implementation of new Dewey 2.0 TextInput component. Spec here:
https://clever.atlassian.net/wiki/spaces/DES/pages/1435959474/Text+Inputs+Component+spec

**Screenshots/GIFs:**
TextInput2-birthday

**Testing:**

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


src/TextInput2/TextInput2.tsx
```diff
@@ -0,0 +1,101 @@
...
+          {!!requirement && (
+            <span className={cssClass.INFO_REQUIREMENT} aria-live="polite">
+              {requirement}
+            </span>
+          )}
+        </div>
+      )}
+      <div className={classnames(cssClass.INPUT_CONTAINER, isFocused && cssClass.INPUT_FOCUSED)}>
+        {icon && <i className={cssClass.LEADING_ICON}>{icon}</i>}
+        <input
```
jonjonray, line 79: minor nit: Since the label is an optional field, could we potentially add an extra prop for `aria-label` conditionally if there is no label? My accessibility knowledge is a bit fuzzy though so there might be a better way

jonjonray (approved):
LGTM. Left a minor accessibility nit, although not entirely sure if it has merit

src/TextInput2/TextInput2.tsx
johnhuangclever: Ahh, I was under the assumption that label is optional, but according to a11y I think we should keep the label required and have a `hideLabel` flag that makes it still available to screen readers. Interesting. Thanks!

PR #587, opened 2021-02-19
Fix new_component code to parse and edit multiline statements

johnhuangclever:
TL:DR; replaced regex code gen with AST code gen via recast

This [is an example of a line](https://github.com/Clever/components/pull/586/files#diff-5d864b760e32dc915a93785758ebd4d64be71b68ed646ff27a1ac966c36c9892R133) of codegen that this modified code is supposed to generate

the new_component script wasn't adding a sidebar link properly because the regex was failing on code statements that became multi-line statements via auto-formatting:
https://github.com/Clever/components/blob/e66f515115e30c864c960f45ff25bb5882e90920/docs/components/SideBar/SideBar.jsx#L110-L115

Instead of trying to figure out how to update the regex, I decided to play around with some tool I heard about but never had the chance to dive into and learn. AST code mods:
https://github.com/benjamn/recast
https://github.com/benjamn/ast-types

It turns out there are a few libraries that are popular, but recast seemed to be the one that fits the bill the most. I tried using babel, which is a slightly different library, but it mangled the formatting of the file. Recast is built to preserve as much formatting as possible, so I ended up with that. babel has nicer documentation though, which explains in more depth what this AST code mod stuff is all about:
https://github.com/jamiebuilds/babel-handbook/blob/master/translations/en/plugin-handbook.md
(a simple babel tutorial https://lihautan.com/step-by-step-guide-for-writing-a-babel-transformation/)

People tend to use this AST explorer to write some simple code that's similar to what they want to transform to be able to reason about the code structure. Check it out:
https://astexplorer.net/#/gist/3fc672d80a6e82fa02103ad43fcddfbe/204ff6161580db41d958fd94bf050e04fba5acc6

### DIFF
--- a/bin/componentUtils.ts
+++ b/bin/componentUtils.ts
@@ -1,6 +1,8 @@
 import * as _ from "lodash";
 import * as fs from "fs";
 import * as path from "path";
+import * as recast from "recast";
+import * as astTypes from "ast-types";
 
 const COMPONENT_NAME_PLACEHOLDER = "NewComponent";
 const COMPONENT_NAME_PLACEHOLDER_REGEX = /NewComponent/g;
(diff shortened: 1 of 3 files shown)


adamvictorclever (approved):
Nice!!! This is definitely an improvement over the regex version!

You might be interested in checking out [jscodeshift ](https://github.com/facebook/jscodeshift)which is a wrapper around `recast` with an [alternate API](https://github.com/facebook/jscodeshift#the-jscodeshift-api) that could be nicer in some cases. I'm not sure if this is one of those cases! I've used `jscodeshift` [a little in the past](https://github.com/Clever/css-classnames-codegen/pull/7) and found it fairly easy to work with.

johnhuangclever:
Yeah! I actually heard first about jscodeshift, but I think it's more CLI triggered rather than code-embedded. I think there's a world where if we wrote the `componentUtils` code from scratch we would set up a toolchain of jscodeshift modules, but I found it worked well to just embed the recast code into it. But I'm a noob at this so would love to hear your thoughts.

On that note I would love to zoom with you about this to hear about your experience with this in general! That https://github.com/Clever/css-classnames-codegen repo looks super interesting! I had no idea you played around with it at Clever, and I will definitely dive into the code to learn more about it

adamvictorclever:
Ohh good point about `jscodeshift` being CLI, I forgot about that. I'd have thought there'd be ways to run it programmatically, but Google results make me unsure.

Yeah, happy to chat on Zoom! css-classnames-codegen is the repo for [styles-ts](https://clever.atlassian.net/wiki/spaces/ENG/pages/1529938054/styles-ts). Most of it doesn't use jscodeshift, but (like you) I felt like trying out writing a codemod so I wrote one there 😄 Just threw some time on your calendar for a Zoom call Wednesday, but feel free to move it around to whatever works best 😃

PR #591, opened 2021-02-25
ID-55 multi select

johnhuangclever:
**Overview:**
Mostly feature complete multi select using downshift
Unfinished design specs here
https://clever.atlassian.net/wiki/spaces/DES/pages/1440023006/Drop-down+Inputs+Component+spec
https://www.figma.com/file/IBVNvo2dI221k4Dye4Hg0S/Dewey-2.0-WIP?node-id=21490%3A11194

**Screenshots/GIFs:**
ID-55-done-multiselect

**Testing:**

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,235 @@
...
+                {...getItemProps({ item, index: i })}
+              >
+                {item.value || item.stringValue}
+              </li>
+            ))
+          ) : (
+            <div
+              className={classNames(!creatable && cssClass.NO_ITEMS_FOUND)}
+              onClick={() => {
+                console.log(creatable);
```
prime-time, line 218: ✂️

src/MultiSelect/MultiSelect.less
```diff
@@ -0,0 +1,142 @@
...
+  }
+}
+
+.MultiSelect--item {
+  color: @neutral_dark_gray;
+  .borderRadius--l();
+  .padding--xs();
+
+  &--highlighted {
+    background-color: ~"@{primary_blue_tint_2}80";
```
prime-time, line 134: Interesting, I haven't seen this before

src/MultiSelect/MultiSelect.less
```diff
@@ -0,0 +1,142 @@
...
+  .borderRadius--l();
+  width: 0%;
+  .flexGrow();
+
+  font-family: "Proxima Nova";
+  .text--medium();
+  line-height: @size_l;
+
+  &::placeholder {
+    //&::-webkit-input-placeholder, &::-moz-placeholder, &:-ms-input-placeholder {
```
prime-time, line 89: Are these prefixes not needed?

src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,235 @@
...
+      <div
+        className={classNames(
+          cssClass.SELECT_CONTAINER,
+          isOpen && cssClass.SELECT_CONTAINER_FOCUSED,
+        )}
+        {...getComboboxProps()}
+      >
+        <div className={cssClass.SELECTED_ITEMS_CONTAINER}>
+          {selectedItems.length > 0 &&
+            selectedItems.map((item, i) => (
```
prime-time, line 161: I *think* it would be fine to render an empty array and skip the `selectedItems.length > 0` check.

src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,235 @@
...
+              >
+                {item.value || item.stringValue}
+                <span
+                  className={cssClass.SELECTED_ITEM_BUTTON}
+                  onClick={(e) => {
+                    e.stopPropagation();
+                    removeSelectedItem(item);
+                  }}
+                >
+                  &#10005;
```
prime-time, line 175: Is there an advantage to using the code rather than the symbol here? If we need to use the symbol, my preference would be to include the name as a comment or declare it as a constant.

src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,235 @@
...
+  NO_ITEMS_FOUND: "MultiSelect--itemsNoneFound",
+};
+
+function getSelectableItems(items: Item[], selectedItems: Item[], inputValue: string) {
+  return items.filter(
+    (item) =>
+      !selectedItems.some((si) => si.key === item.key) &&
+      (inputValue === "" ||
+        item.stringValue.toLocaleLowerCase().includes(inputValue.toLocaleLowerCase())),
+  );
```
prime-time, line 53: I think this could potentially be slow if the number of items is very large, and about half the items are selected. To avoid potential slowness in that case, you could create a set of `selectedItems.key`s at the top of this function.
Another small optimization is to store `inputValue.toLocaleLowerCase()` in a variable at the top of this function. 
I'm not sure how much this would come up in our use cases, so these optimizations may end up being unecessary.

src/MultiSelect/MultiSelect.less
johnhuangclever: I couldn't get them to work (seems like maybe related to this https://stackoverflow.com/a/41246875) so I commented them out. Looks it works when I separate the selectors out though

src/MultiSelect/MultiSelect.tsx
johnhuangclever: Nice!

src/MultiSelect/MultiSelect.tsx
johnhuangclever: Whoops, this is straight from the downshift example. I'm guessing they did it for readability, but I can add a comment as well

src/MultiSelect/MultiSelect.tsx
johnhuangclever: Nice! Sounds good.

I'm going to refactor the creatable code a little bit in a future PR, so I'll make those optimizations there

johnhuangclever:
any thoughts on the prop name `items` vs `options`? downshift API uses `items`, but I feel like `options` makes more sense here

prime-time:
The fact that those popular libraries use `options` makes a very strong case for it. Additionally, the HTML `<select>` element takes in a list of `<option>`s. I would say `options` is the way to go.

prime-time:
Another thing you could do is have this component take in the options as children, and then you wouldn't need to worry about the name 😄 .
It looks like material-ui's `Select` takes in its options as children.

johnhuangclever:
Yeah I think I'll leave that for the future 👍

prime-time:
It can be trickier to type children, so I agree that it's probably best to look into that in the future.

johnhuangclever:
ok! If the rest looks good, I'm gonna merge it

src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,245 @@
...
+import * as FontAwesome from "react-fontawesome";
+
+import Label from "../Label/Label";
+import { FormElementSize, formElementSizeClassName } from "../utils/Forms";
+import { Values } from "../utils/types";
+
+import "./MultiSelect.less";
+
+// value must be provided for searchability
+type Option = { key: string; value: string; renderedComponent?: React.ReactNode };
```
prime-time, line 14: It feels like we might be able to get away with two properties here instead of three. What if we dropped `key`, and used value for both purposes? `value` is what the HTML `<select>` uses as its identifier. 

How do you feel about shortening `renderedComponent` to `content`?

src/MultiSelect/MultiSelect.tsx
```diff
@@ -0,0 +1,245 @@
...
+        case useCombobox.stateChangeTypes.ItemClick:
+          return {
+            ...changes,
+            isOpen: true, // keep the menu open after selection.
+          };
+        default:
+          break;
+      }
+      return changes;
+    },
```
prime-time, line 121: Oooo fancy 🎩

prime-time (approved):
🎊 🍾 🏆🏁  🏎️ 💨

src/MultiSelect/MultiSelect.tsx
johnhuangclever: yeah sounds good

src/MultiSelect/MultiSelect.tsx
johnhuangclever: straight from the tutorial example 👌 https://www.downshift-js.com/use-multiple-selection

PR #593, opened 2021-02-26
make "add item X" work properly

johnhuangclever:
**Overview:**
Fix the `not found` and `create new item` case

Summary of changes:
* The `"Add X"` option now acts like a regular option item (i.e. reducers and highlighting)
* `"Add X"` will not be shown if the input value is an exact match of an existing item. Instead, it will show `no options`

**Screenshots/GIFs:**
ID-55-multiselect-create

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MultiSelect/MultiSelect.tsx
+++ b/src/MultiSelect/MultiSelect.tsx
@@ -10,6 +10,8 @@ import { Values } from "../utils/types";
 
 import "./MultiSelect.less";
 
+const ADD_NEW_ITEM_KEY = "MultiSelect--addNewItem";
+
 // value represents the searchable text of the option
 type Option = { value: string; content?: React.ReactNode };
 
@@ -47,11 +49,15 @@ export const cssClass = {
 function getSelectableOptions(options: Option[], selectedItems: Option[], inputValue: string) {
   const selectedValues = new Set<string>(selectedItems.map((si) => si.value));
   const inputLowerCase = inputValue.toLocaleLowerCase();
-  return options.filter(
-    (o) =>
-      !selectedValues.has(o.value) &&
-      (inputValue === "" || o.value.toLocaleLowerCase().includes(inputLowerCase)),
-  );
+  return [
+    // add a dummy "add item X" placeholder, we will optionally show it via the render code
+    { value: ADD_NEW_ITEM_KEY },
+    ...options.filter(
+      (o) =>
+        !selectedValues.has(o.value) &&
+        (inputValue === "" || o.value.toLocaleLowerCase().includes(inputLowerCase)),
+    ),
+  ];
 }
 
 /*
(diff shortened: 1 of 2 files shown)


src/MultiSelect/MultiSelect.tsx
```diff
@@ -220,33 +233,30 @@ const MultiSelect: React.FC<Props> = ({
...
-                  i === highlightedIndex && cssClass.MENU_OPTION_HIGHLIGHTED,
-                )}
-                key={`${o.value}${i}`}
-                {...getItemProps({ item: o, index: i })}
-              >
-                {o.content || o.value}
-              </li>
-            ))
+          // slice out the dummy "add item X" option
+          (selectableOptions.slice(1).length === 0 && (!creatable || hasExactMatch) ? (
```
prime-time, line 237:
```suggestion
          (selectableOptions.length === 1 && (!creatable || hasExactMatch) ? (
```
Since the sliced array isn't actually being used, this way would be a little simpler and saves the computer some work.

johnhuangclever: There's a change that the selectableOptions will *correctly* be 1 item, which is if the typeahead matches exactly one item

src/MultiSelect/MultiSelect.tsx
```diff
@@ -220,33 +233,30 @@ const MultiSelect: React.FC<Props> = ({
...
-                  addSelectedItem(newOption);
-                }
-              }}
-            >
-              {creatable ? `Create new item ${inputValue}` : "No options"}
-            </div>
+            selectableOptions.map((o, i) => {
+              const isAddNewItemOption = o.value === ADD_NEW_ITEM_KEY;
+              // hide the first dummy "add item X" option if not eligible
+              if (isAddNewItemOption && (inputValue.length === 0 || !creatable || hasExactMatch)) {
```
prime-time, line 243:
```suggestion
              if (isAddNewItemOption && (inputValue === "" || !creatable || hasExactMatch)) {
```
This way would be more idiomatic to check if `inputValue` is empty

src/MultiSelect/MultiSelect.tsx
prime-time: Ohhh I see, so the equivalent length check would actually be `selectableOptions.length <= 1`

johnhuangclever: oh maybe I misunderstood you. Actually `=== 1` might work here. I can take a look

src/MultiSelect/MultiSelect.tsx
johnhuangclever: Hmm yeah maybe I can do `!inputValue`

prime-time (approved):
Looks good to me ✖️ ✨

johnhuangclever:
Do you think it would be clearer if I just set up 2 separate variables?
e.g.
```
selectableOptions = get...()
selectableOptionsWithCreate = [{dummy}, ...selectableOptions]
```

PR #594, opened 2021-03-01
Modify MessagingThreadListItem to optionally display an alert icon

dengjulie:
**Jira:**
https://clever.atlassian.net/browse/PRTL-241

**Overview:**
Teachers Inviting Guardians aka teavites project:
We want to display an alert icon in the thread list for contacts whose invites have expired or been rejected, so that the teacher knows to take action.
image

**Screenshots/GIFs:**
Screen Shot 2021-02-28 at 5 25 56 PM

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingThreadListItem/MessagingThreadListItem.tsx
+++ b/src/MessagingThreadListItem/MessagingThreadListItem.tsx
@@ -86,6 +86,8 @@ export const MessagingThreadListItem: React.FC<
     </FlexBox>
   );
 
+  const isIndicatorAlignedWithSubContent = subContent && timestamp;
+
   return (
     <div ref={ref}>
       <FlexBox
@@ -114,16 +116,17 @@ export const MessagingThreadListItem: React.FC<
                 {timestamp && _formatDateForTimestamp(timestamp)}
               </div>
             )}
-            {!subContent && indicator}
           </FlexBox>
           {subContent && (
             <FlexBox alignItems={ItemAlign.CENTER}>
               <div className={subContentClass}>{subContent}</div>
-              {timestamp && indicator}
+              {isIndicatorAlignedWithSubContent && indicator}
             </FlexBox>
           )}
         </FlexItem>
-        <FlexItem alignItems={ItemAlign.CENTER}>{subContent && !timestamp && indicator}</FlexItem>
+        <FlexItem alignItems={ItemAlign.CENTER}>
+          {!isIndicatorAlignedWithSubContent && indicator}
+        </FlexItem>
       </FlexBox>
     </div>
   );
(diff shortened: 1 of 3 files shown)


src/MessagingThreadListItem/MessagingThreadListItem.tsx
```diff
@@ -110,10 +119,11 @@ export const MessagingThreadListItem: React.FC<
           {subContent && (
             <FlexBox alignItems={ItemAlign.CENTER}>
               <div className={subContentClass}>{subContent}</div>
-              {indicator}
+              {timestamp && indicator}
```
TheRealBenjiLi, line 122: Why do we only want to show the indicator when a timestamp is around? :O

src/MessagingThreadListItem/MessagingThreadListItem.tsx
```diff
@@ -110,10 +119,11 @@ export const MessagingThreadListItem: React.FC<
...
             </FlexBox>
           )}
         </FlexItem>
+        <FlexItem alignItems={ItemAlign.CENTER}>{subContent && !timestamp && indicator}</FlexItem>
```
TheRealBenjiLi, line 126: If I'm reading this right, this is rendering the indicator. But isn't that also being done on line 122?

src/MessagingThreadListItem/MessagingThreadListItem.tsx
dengjulie: So we have 3 options: either display timestamp + indicator, just timestamp, or just indicator. There is no instance of messaging using just an indicator yet, so this is a new use case for invites (mocks don't have timestamp, and we don't sort threads by order of time invites are sent out). With just timestamp, it looks fine to have it placed in the same horizontal axis of the title, but with just indicator, it looks a bit funky for it to be on the lower right of the thread list item. we want it vertically centered as another column of the overall parent flexbox on the right side of the thread item. Does that make sense?

src/MessagingThreadListItem/MessagingThreadListItem.tsx
dengjulie: Commented above, but maybe this is too confusing at first glance... I'll think of how to make it more clear.

dengjulie:
Pause on this, going to ask Jemma some design Qs

src/MessagingThreadListItem/MessagingThreadListItem.tsx
TheRealBenjiLi: Ah yep, that clears things up! :)

TheRealBenjiLi (approved):
🌮

PR #595, opened 2021-03-02
some jscodeshift transforms for new components

johnhuangclever:
to replace this block of recast code:
https://github.com/Clever/components/blob/14cb8a9ad537c3cd4bbd0bd6203522f46bfe38dd/bin/componentUtils.ts#L124-L195

Also add code generation for screenshot code for new components
https://github.com/Clever/components/pull/595/files#diff-b2496e80299b8c3150b1944450bd81c622e04e13d15c411d291db0927d75fd6bL26-L35

A fun exercise with hopefully some benefits of ASTs


bin/new_component.sh
```diff
@@ -5,4 +5,9 @@ cd "${BASH_SOURCE%/*}"
 
 NAME=$1
 ../node_modules/.bin/ts-node --project ../tsconfig.bin.json ./create_new_component.ts $NAME
+npx jscodeshift --dry-run --componentName="$NAME" -t ./addToSidebar.ts ../docs/components/SideBar/SideBar.jsx
 ../node_modules/.bin/prettier --write ../docs/components/SideBar/SideBar.jsx
+npx jscodeshift --dry-run --componentName="$NAME" -t ./addRoute.ts ../docs/docs.jsx
+../node_modules/.bin/prettier --write ../docs/docs.jsx
+npx jscodeshift --dry-run --componentName="$NAME" -t ./addScreenshot.ts ../docs/components/ComponentsView.jsx
```
adamvictorclever, line 12: Nice! Adding the screenshot config is useful!

package.json
```diff
@@ -85,6 +86,7 @@
     "ignore-styles": "^5.0.1",
     "jest-cli": "^22.4.3",
     "jest-enzyme": "^6.0.0",
+    "jscodeshift": "^0.11.0",
```
adamvictorclever, line 89: I'm curious what your experience with `jscodeshift` was and how you'd compare it to working with just recast?

bin/new_component.sh
```diff
@@ -5,4 +5,9 @@ cd "${BASH_SOURCE%/*}"
 
 NAME=$1
 ../node_modules/.bin/ts-node --project ../tsconfig.bin.json ./create_new_component.ts $NAME
+npx jscodeshift --dry-run --componentName="$NAME" -t ./addToSidebar.ts ../docs/components/SideBar/SideBar.jsx
```
adamvictorclever, line 8: Not even remotely blocking, but it could be cool to add some simple test cases for these codemods!

adamvictorclever (approved):
LGTM! Sorry for the delay on reviewing!

bin/new_component.sh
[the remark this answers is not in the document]
johnhuangclever: whoops, turns out that the `--dry-run` is not even the correct flag, so this was doing nothing. I will remove it.

Thanks!

package.json
johnhuangclever: Actually understanding it's API was really confusing for me but it's much more powerful if you can find the right functions/access patterns.

Recasts I think is alot clearer in what it's doing but hard to write.

bin/new_component.sh
johnhuangclever: Yeah! Would love to dive into how to do this properly. I notice you set those up in your repo and it's amazing

PR #597, opened 2021-03-03
ID-55 select2 required

johnhuangclever:
Probably helpful to review commit-by-commit

**Overview:**
Support a "required" and error state. Summary of changes:
* Move the `TextInput2Requirement` enum to more generalized form utils
* add requirement option and requirement label elements to Select2
* Support error states for required prop (error and initialError)
* open menu when clear button is clicked

**Screenshots/GIFs:**

https://user-images.githubusercontent.com/13126257/109743707-ff5bb880-7b85-11eb-98e0-5a4d6d0b8c44.mov

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/Select2/Select2.less
+++ b/src/Select2/Select2.less
@@ -122,8 +122,8 @@
 
 .Select2--errorIcon {
   color: @alert_red;
-  display: flex;
-  align-items: center;
+  .flexbox();
+  .items--center();
 }
 
 .Select2--clearButton {
--- a/src/Select2/Select2.less
+++ b/src/Select2/Select2.less
@@ -6,8 +6,19 @@
   position: relative;
 }
 
-.Select2--label {
+.Select2--infoRow {
+  .flexbox();
+  .justify--between();
   .padding--bottom--2xs();
+  color: @neutral_medium_gray;
+  .text--small();
+
+  &--labelHidden {
+    .justify--end();
+  }
+}
+
+.Select2--label {
   color: @neutral_medium_gray;
   .text--small();
 
@@ -28,6 +39,10 @@
     border: 0;
     /* stylelint-enable */
   }
+}
+
+.Select2--infoRequirement {
+  .text--italic();
 }
 
 .Select2--selectContainer {
--- a/src/Select2/Select2.less
+++ b/src/Select2/Select2.less
@@ -52,8 +52,7 @@
   .border--m(@neutral_silver);
   .borderRadius--l();
   background-color: @neutral_white;
-  .margin--top--3xs();
-  .padding--left--xs();
+  .padding--x--s();
   box-sizing: border-box;
   transition: box-shadow @timingQuickly ease-out;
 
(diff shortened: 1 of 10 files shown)


src/Select2/Select2.less
```diff
@@ -94,15 +115,23 @@
...
 
+.Select2--trailingElement {
+  .margin--left--xs();
+  min-height: @size_2xl;
+}
+
+.Select2--errorIcon {
+  color: @alert_red;
+  display: flex;
+  align-items: center;
```
prime-time, line 126: I notice that the flex mixins are used in a rule at the top of the file. Is there a particular reason for using native css flex here? I'm not suggesting one over the other, rather just wondering if it was intentional.

src/Select2/Select2.less
```diff
@@ -48,10 +62,22 @@
...
+  }
+
   /* stylelint-disable-next-line no-descending-specificity */
   &:hover {
     .border--m(@primary_blue);
   }
+
+  &--error:hover {
+    .border--m(@alert_red);
+  }
```
prime-time, line 76: I'm not 100% sure, but I think you could combine this with the rule a few lines above like this.
```suggestion
  &--error, &--error:hover {
    .border--m(@alert_red);
  }
```

src/Select2/Select2.tsx
```diff
@@ -111,20 +129,31 @@ const Select2: React.FC<Props> = ({
...
-      </label>
+      <div className={classNames(cssClass.INFO_ROW, hideLabel && cssClass.INFO_ROW_LABEL_HIDDEN)}>
+        <label
+          className={classNames(cssClass.LABEL, hideLabel && cssClass.LABEL_HIDDEN)}
+          {...getLabelProps()}
+        >
+          {label}
+        </label>
+        {requirement && (
+          <label className={cssClass.INFO_REQUIREMENT} aria-live="polite" {...getLabelProps()}>
```
prime-time, line 147: Woa interesting, I didn't know `aria-live="polite" ` was a thing.

johnhuangclever: Copied this from the `TextInput` component. Honestly I don't know how it fulfills a11y requirements https://github.com/Clever/components/blob/b54fd68cafa664519ca199853beb8554dc812966/src/TextInput/TextInput.tsx#L234 🤷‍♀️

src/Select2/Select2.less
johnhuangclever: whoops missed it, yeah. Thanks.

We need the flex mixins for cross-browser support. I think fampo set it up so we could use the native attributes and to autogenerate the platform-specific attributes

(we should use that at some point)

src/Select2/Select2.less
johnhuangclever: I tried using commas in alot of places and they didn't seem to work. Not sure if it's my css problem or a less quirk or something, but I couldn't get it to work

src/Select2/Select2.less
prime-time: Ah that's right, I forgot what the mixins did besides save us from typing a few characters 😄

src/Select2/Select2.less
prime-time: Sounds like I need to review selectors 🦧 . Carry on.

prime-time (approved):
Niiice, looks good to me. 

You probably also want to bump the version.

PR #602, opened 2021-03-08
Add guardian created account alert message to thread history

dengjulie:
**Jira:**
https://clever.atlassian.net/browse/PRTL-248

**Overview:**

**Screenshots/GIFs:**
inviteAcceptedDewey

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingThreadHistory/MessagingThreadHistory.less
+++ b/src/MessagingThreadHistory/MessagingThreadHistory.less
@@ -17,10 +17,32 @@
   margin-top: auto;
 }
 
+.ThreadHistory--AlertMessage--Container {
+  .border--m(@neutral_silver);
+  .borderRadius--l();
+  .margin--x--xs();
+  .margin--y--xs();
+  .padding--m();
+  align-items: center;
+  color: @neutral_dark_gray;
+}
+
+.ThreadHistory--AlertMessage {
+  .margin--left--m();
+}
+
 // For mobile/tablet
 @media only screen and (max-width: @breakpointM) {
   .ThreadHistory--Divider {
     margin-top: 0.75rem;
     margin-bottom: 0.75rem;
+  }
+  .ThreadHistory--AlertMessage--Container {
+    .padding--x--s();
+    .padding--y--xs();
+  }
+
+  .ThreadHistory--AlertMessage {
+    font-size: 0.875rem;
   }
 }
--- a/src/MessagingThreadHistory/MessagingThreadHistory.less
+++ b/src/MessagingThreadHistory/MessagingThreadHistory.less
@@ -17,32 +17,10 @@
   margin-top: auto;
 }
 
-.ThreadHistory--AlertMessage--Container {
-  .border--m(@neutral_silver);
-  .borderRadius--l();
-  .margin--x--xs();
-  .margin--y--xs();
-  .padding--m();
-  align-items: center;
-  color: @neutral_dark_gray;
-}
-
-.ThreadHistory--AlertMessage {
-  .margin--left--m();
-}
-
 // For mobile/tablet
 @media only screen and (max-width: @breakpointM) {
   .ThreadHistory--Divider {
     margin-top: 0.75rem;
     margin-bottom: 0.75rem;
-  }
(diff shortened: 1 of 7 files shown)


src/MessagingThreadHistory/MessagingThreadHistory.less
```diff
@@ -17,10 +17,32 @@
...
     margin-top: 0.75rem;
     margin-bottom: 0.75rem;
   }
+  .ThreadHistory--AlertMessage--Container {
+    .padding--x--s();
+    .padding--y--xs();
+  }
+
+  .ThreadHistory--AlertMessage {
+    font-size: 0.875rem;
```
prime-time, line 46: There's now a less mixin for this, thanks to @adamvictorclever 🎉 
https://github.com/Clever/components/blob/8ce483bdceecd1d3b12d73dd7643e2bda7d2888a/src/less/type-size.less#L23-L25
```suggestion
    .text--smallMedium();
```

src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -21,6 +25,7 @@ export interface MessageData {
   content: React.ReactNode;
   index: number;
   readStatusText?: string;
+  attributes: any;
```
prime-time, line 28:
```suggestion
  attributes: { [attr: string]: string};
```

I _think_ this would work, and would be a slightly better type.

prime-time (approved):
looks good to me

nbhatia823 (commented):
Lol I did not mean to submit review yet, my bad

src/MessagingThreadHistory/MessagingThreadHistory.tsx
```diff
@@ -144,6 +149,19 @@ function _interleaveMessagesWithDividers(
...
     );
+    if (message.attributes?.displayAlertMessageAfter === "guardian_created_account") {
+      messagesWithDividers.push(
+        <FlexBox
+          className={cssClasses.ALERT_MESSAGE_CONTAINER}
+          key={`alertMessage-${message.index}`}
+        >
+          <FontAwesome name="check" size="lg" />
+          <FlexItem className={cssClasses.ALERT_MESSAGE} grow>
+            Invite accepted! This guardian can now reply to your messages!
```
nbhatia823, line 160: It seems like this context belongs in the consumer (launchpad in this case) of `MessagingThreadHistory` because it is a specific implementation detail. I think the Dewey components are meant to be flexible and contextually-unaware as possible.

Instead I would think that `ThreadHistory` might have another `MessageMetadata` with `type="alert"`. That way we can reuse this alert container for different alerts as they grow at the consumer level. And we could also display different copy based on the user type but that would be up to the consumer.

Also, doesn't this mean that guardians will also see this message? Or is that what is supposed to happen

dengjulie: Good points! Refactored it so that more of the logic is left to the client, which should address both issues. Opted to just create a separate component from MessageMetadata since the component is quite different.

dengjulie:
tenor

nbhatia823 (approved):
I like the changes, although I feel there could still be benefit to decoupling the alert message from the previous message. For example, we may want to display an alert message as the first item in a certain thread, but that wouldnt be possible now. Regardless, that might be a future addition we make if this use case comes up.

dengjulie:
Oh I see, I misunderstood what you meant then. That makes sense. I'll give it a try and see if the changes are straightforward - we're fine with not using twilio messaging index as the index for thread history, yeah?

dengjulie:
Okay, I'll go ahead with this, partially because there's not a use case for another alert message yet and because I think the switch the internal messaging infrastructure could open up the possibility of storing alert messages as a message in the db instead of as an attribute of a message.

PR #603, opened 2021-03-10
Remove unused prop

dengjulie:
**Jira:**
n/a

**Overview:**
Removes an unused alignItems prop on non-FlexBox component. I was wondering why we were seeing this error in launchpad.
Don't check the git blame 🙈 
image

**Screenshots/GIFs:**
Manually tested, looks the same 

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/MessagingThreadListItem/MessagingThreadListItem.tsx
+++ b/src/MessagingThreadListItem/MessagingThreadListItem.tsx
@@ -137,9 +137,7 @@ export const MessagingThreadListItem: React.FC<
             </FlexBox>
           )}
         </FlexItem>
-        <FlexItem alignItems={ItemAlign.CENTER}>
-          {!isIndicatorAlignedWithSubContent && indicator}
-        </FlexItem>
+        <FlexItem>{!isIndicatorAlignedWithSubContent && indicator}</FlexItem>
       </FlexBox>
     </div>
   );
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.79.0",
+  "version": "2.79.1",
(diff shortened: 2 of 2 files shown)


src/MessagingThreadListItem/MessagingThreadListItem.tsx
```diff
@@ -137,9 +137,7 @@ export const MessagingThreadListItem: React.FC<
             </FlexBox>
           )}
         </FlexItem>
-        <FlexItem alignItems={ItemAlign.CENTER}>
-          {!isIndicatorAlignedWithSubContent && indicator}
-        </FlexItem>
+        <FlexItem>{!isIndicatorAlignedWithSubContent && indicator}</FlexItem>
```
adamvictorclever, line 140: Wait I'm trying to understand this, is `alignItems` a prop for our `FlexItem` component, but not actually supported by browsers?

dengjulie: There is no prop for alignItems on FlexItem! This was just a simple mistake of an additional prop that did nothing.

adamvictorclever: Ahh I see, I thought TypeScript would have caught it then but we didn't type it enough https://github.com/Clever/components/blob/master/src/flex/FlexItem.tsx#L17-L18

adamvictorclever (approved):

PR #606, opened 2021-03-19
M5G-154: unread orb color

jonah-clever:
**Jira:**

https://clever.atlassian.net/browse/M5G-154

**Overview:**

Updating `MessagingThreadListItem` to have an optional `unreadOrbColor` prop, which can be set to either "primary_blue" or "accent_teal". Teal will be used in student and guardian messaging, and teacher messaging will remain blue. Blue remains the default color

I added a SegmentedControl to the config options to select blue or teal, which is visible in the example when "is_read" is false

**Screenshots/GIFs:**

Screen Shot 2021-03-19 at 4 29 14 PM
Screen Shot 2021-03-19 at 4 30 43 PM

**Testing:**

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] Edge

**Roll Out:**

- Before merging:
  - [x] Updated docs
  - [x] Bumped version in `package.json` **(Minor)**

- After merging:

### DIFF
--- a/src/MessagingThreadListItem/MessagingThreadListItem.tsx
+++ b/src/MessagingThreadListItem/MessagingThreadListItem.tsx
@@ -26,9 +26,11 @@ const cssClasses = {
   UNREAD_TEXT: "ThreadListItem--UnreadText",
 };
 
-const unreadOrbColorClasses = {
-  primary_blue: "ThreadListItem--UnreadOrb--PrimaryBlue",
+type UnreadOrbColor = "accent_teal" | "primary_blue";
+
+const unreadOrbColorClasses: Record<UnreadOrbColor, String> = {
   accent_teal: "ThreadListItem--UnreadOrb--AccentTeal",
+  primary_blue: "ThreadListItem--UnreadOrb--PrimaryBlue",
 };
 
 export type Status = "active" | "off";
@@ -47,7 +49,7 @@ type Props = {
   title: string;
   hasAlert?: boolean;
   alertTooltip?: string;
-  unreadOrbColor?: "accent_teal" | "primary_blue";
+  unreadOrbColor?: UnreadOrbColor;
 };
 
 export const MessagingThreadListItem: React.FC<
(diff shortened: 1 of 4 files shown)


src/MessagingThreadListItem/MessagingThreadListItem.tsx
```diff
@@ -26,6 +26,11 @@ const cssClasses = {
   UNREAD_TEXT: "ThreadListItem--UnreadText",
 };
 
+const unreadOrbColorClasses = {
+  primary_blue: "ThreadListItem--UnreadOrb--PrimaryBlue",
+  accent_teal: "ThreadListItem--UnreadOrb--AccentTeal",
+};
```
jonah-clever, line 32: this feels a bit unorthodox, but worked so i thought what the heck. open to other ideas tho if there's a better way to handle this

nbhatia823: This seems pretty solid to me! I think this would have been the de-facto way to do it until recently when Typescript introduced this new feature called "Records." So, if you're interested i can describe it below but you could use your current approach as well - 

https://www.typescriptlang.org/docs/handbook/utility-types.html#example-3

A Record is a type that maps a set of strings to some other values, and enforces that each string in the set has to have a corresponding value.

So it would be something like:

```
type UnreadOrbColors = "accent_teal" | "primary_blue"

const unreadOrbColorClasses: Record<UnreadOrbColors, String> = {
   accent_teal: "ThreadListItem--UnreadOrb--AccentTeal",
  primary_blue: "ThreadListItem--UnreadOrb--PrimaryBlue",
}
```

and then in the props for the component, you would just reference the same type:

```
type Props = {
   ...
   unreadOrbColor?: UnreadOrbColor;
}

and then reference it the same way you already are with:

unreadOrbColor && unreadOrbColorClasses[unreadOrbColor],

```

The nice thing about this is that if someone added in another `unreadOrbColor`, they would get an error at `unreadOrbColorClasses` if they didnt also add the class since the Record enforces that each string in the type has a corresponding value in the record.

nbhatia823 (approved):
Nice! I relly like the segmented control in the View and the way you are handling the CSS classes seems solid 💯

src/MessagingThreadListItem/MessagingThreadListItem.tsx
jonah-clever: Nice, I like that, very TypeScripty™️. I'll do it

nbhatia823 (approved):
Looks good!

PR #617, opened 2021-04-07
TextTruncate can handle null and undefined text inputs

TheRealBenjiLi:
**Jira:**

[PRTL-371](https://clever.atlassian.net/browse/PRTL-371)

**Overview:**
Flare 1070 was caused by a null/undefined being passed into TextTruncate. While it would make sense to avoid that happening in the first place, this change will prevent it at the component level too!

**Screenshots/GIFs:**
This is when null is passed in: 
Screen Shot 2021-04-07 at 4 45 33 PM

Compared to normal: 
Screen Shot 2021-04-07 at 4 45 55 PM

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/TextTruncate/TextTruncate.tsx
+++ b/src/TextTruncate/TextTruncate.tsx
@@ -72,6 +72,13 @@ export default class TextTruncate extends React.PureComponent<Props> {
     } = this.props;
     const { truncated } = this.state;
 
+    // If text is null or undefined, return immediately
+    if (!text) {
+      return (
+        <></>
+      )
+    }
+
     if (text.length < maxCharsShown) {
       return (
         <div className={classnames(cssClass.CONTAINER, className)}>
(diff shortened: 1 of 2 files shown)


src/TextTruncate/TextTruncate.tsx
```diff
@@ -72,6 +72,13 @@ export default class TextTruncate extends React.PureComponent<Props> {
     } = this.props;
     const { truncated } = this.state;
 
+    // If text is null or undefined, return immediately
+    if (!text) {
+      return (
+        <></>
```
prime-time, line 78: I would actually return `null` here since fragments are typically used to wrap other elements.

prime-time (approved):
🔧

src/TextTruncate/TextTruncate.tsx
prime-time: I should add that this is actually the recommended way as opposed to my personal preference 😄 
https://reactjs.org/docs/conditional-rendering.html#preventing-component-from-rendering

TheRealBenjiLi: For sure! I didn't know null was an acceptable return here, since it's for rendering. Thanks! :)

PR #622, opened 2021-04-12
AnnouncementBubbleView fix + small styling tweak

nbhatia823:
https://clever.atlassian.net/browse/M5G-144

**Overview:**

This PR:
1. Fixes the demo View by changing the file suffix from .tsx -> .jsx :hide:. Everything worked with .tsx except for the ExampleCode block interestingly. I also improved the View by adding some extra headers/spacing.
2. Changes the `overflow-wrap` behavior to be the same on mobile and desktop. The old and new way are similar, but the new way works better for very long single words as it will break mid-word.
3. Extends the number of characters for the truncated view of the `QuotedAnnouncementBubble` to 300, matching the `MessagePreview`. It is still truncated by CSS, but I noticed that on wide viewports, the 100 characters wouldn't be enough to fill up the line.
4. Adds a missing CSS class to `AlertBox2View`

Old way with a long word - 
Screen Shot 2021-04-11 at 12 34 34 PM

New way with long word - 

Screen Shot 2021-04-12 at 9 43 54 AM

**Screenshots/GIFs:**

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


src/AnnouncementBubble/QuotedAnnouncementBubble.tsx
```diff
@@ -38,7 +38,7 @@ export const QuotedAnnouncementBubble: React.FC<Props> = ({
   // don't attempt to read the entire hidden message. This only works for plain-string children
   let content = children;
   if (!isExpanded && typeof content === "string") {
-    content = content.substring(0, 100);
+    content = content.substring(0, 300);
```
jonjonray, line 41: QQ about this: I don't think that we can actually fit 300 characters in the single line, non-expanded version of this for the quoted bubble (at least with the way it's currently implemented on the messaging page), so a screen reader would still pick up more characters than are actually visible to the user in that case.

If we need to support this increased character count for a different use-case, can we make this dynamic?

jonjonray (approved):
LGTM.

Left a small comment around the character count increase

src/AnnouncementBubble/QuotedAnnouncementBubble.tsx
nbhatia823: yeah i think you are right. i was not sure a dynamic way to do this. Chloe had raised this as an issue recently and I think we left figuring out a better way to handle this as a todo. i think i had initially done a quick calculation that 100 characters should be the max based on the 14px size and the max width of the bubble, but I saw an issue in Dewey when the Bubble wasn't limited by the chat width. Although, that shouldnt be a realistic scenario, I thought that for now to be safe I would increase the limit until we do some proper calculations to figure out what these limits should be. very open to feedback!

PR #625, opened 2021-04-13
Failure is a part of messaging

nbhatia823:
**Jira:**
https://clever.atlassian.net/browse/M5G-228

**Overview:**
This PR adds in errorMsg functionality to DMs by centralizing the logic for `MessagingBubble` and `AnnouncementBubble`. Instead of the errorMsg being a prop on those components, it is a prop on `MessageData/MessageMetadata`. This means we don't need to duplicate the logic, plus conceptually an error message is similar to a timestamp, so it makes sense for it to be a prop on the metadata instead of the bubbles themselves.

**Screenshots/GIFs:**

Desktop student - 
Screen Shot 2021-04-13 at 1 29 00 PM

Desktop teacher -

Screen Shot 2021-04-13 at 1 27 06 PM

Mobile teacher -
Screen Shot 2021-04-13 at 1 26 57 PM

Announcement error messages still show -
Screen Shot 2021-04-13 at 1 26 44 PM

Dewey View shows error messages randomly - 

Screen Shot 2021-04-13 at 1 11 49 PM

**Testing:**

Tested in Dewey local and Launchpad local

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
  - [x] Bumped version in `package.json`
    - Breaking change?
      - If it is a beta component run `npm version minor`
      - If the component is not in beta run `npm version major`
    - New component or backward-compatible component feature change? Run `npm version minor`
    - Only changing documentation? All good. Skip this step.
  - After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [ ] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
        componentLink: "<COMPONENT LINK>",
        componentImg: "<COMPONENT LINK>.png",
        componentName: "<COMPONENT NAME>",
        componentImgAlt: "A <COMPONENT NAME> component",
      },
      ```
    - [ ] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:
  - [ ] Deployed updated docs (`make deploy-docs`)
  - [ ] Posted in #eng if I made a breaking change to a beta component

PR #628, opened 2021-04-14
Remove extra CSS

nbhatia823:
**Overview:**

I added some CSS in this PR - https://github.com/Clever/components/pull/625/files?file-filters%5B%5D=.less&file-filters%5B%5D=.tsx. It works, but Jonathan and I made some changes at the same time, and message bubbles on narrow screens still can spill off the edge sometimes. I'm cutting down some of the CSS that isn't neccessary to try to get to the bottom of it. Pretty much, before this component would place error messages to the side that bubble is, but now it always floats to the right. This is the same behavior as the READ marker on the bubble. And, we only expect Error messages on "Own" messages, so that's why I felt we should keep it simpler until another use case comes up.

**Screenshots/GIFs:**

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [NA] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.93.0",
+  "version": "2.93.1",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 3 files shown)


lubalexuan (approved):
✂️

PR #644, opened 2021-04-26
Add ExpandableContainer component

dengjulie:
**Jira:**
https://clever.atlassian.net/browse/PRTL-380

**Overview:**
Add ExpandableContainer component, a component that expands to display its contents when toggled.

This component will be useful for messaging chat sidebar; mocks:

image

**Screenshots/GIFs:**
chrome-capture (3)

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:


dengjulie:
Should I mark this as a Beta component?

clcaelynn-clever:
> Should I mark this as a Beta component?

[Started a Slack thread in #frontend](https://clever.slack.com/archives/C0ERS5ABB/p1619542853106600) about this, out of curiosity. 😄

src/ExpandableContainer/ExpandableContainer.less
```diff
@@ -0,0 +1,20 @@
...
+  color: @neutral_dark_gray;
+}
+
+.ExpandableContainer--content {
+  .padding--s();
+}
+
+.ExpandableContainer--title {
+  .padding--s();
+  font-weight: 600;
```
clcaelynn-clever, line 14: No mixin for this?

src/ExpandableContainer/ExpandableContainer.tsx
```diff
@@ -0,0 +1,50 @@
...
+import FlexItem from "../flex/FlexItem";
+
+import "./ExpandableContainer.less";
+
+export interface Props {
+  children: React.ReactNode;
+  className?: string;
+  isExpanded?: boolean;
+  onClick?: () => void;
+  title?: React.ReactNode;
```
clcaelynn-clever, line 16: If the title is optional, it's possible to make an inaccessible version of this, I think, where there's no label, no aria-label, but there _is_ click functionality.

I suggest adding an `ariaLabel?: string` prop, which is stuck onto the FlexBox that has the onClick below; that way, consumers can make sure themselves that it's accessible and uniquely labeled. Making a note of this in the component's prop descriptions is also probably a best practice we should adopt. 😄

clcaelynn-clever (changes requested):
One accessibility change that would be great, but looks awesome! 🚀

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.100.0",
+  "version": "2.102.0",
```
prime-time, line 3: You skipped a version 🔢  😛

prime-time: Actually, it looks like `2.101` was just published, so this is the right version.

docs/components/ExpandableContainerView.jsx
```diff
@@ -0,0 +1,103 @@
...
+              // OR
+              import ExpandableContainer from "clever-components/dist/ExpandableContainer"; // Avoids importing all of clever-components.
+            `}
+          </CodeSample>
+        </header>
+
+        <Example title="Basic Usage:">
+          <ExampleCode>
+            <ExpandableContainer
+              className={"my--custom--class"}
```
prime-time, line 44:
```suggestion
              className="my--custom--class"
```
Just a tiny simplification

prime-time (approved):
This looks good to me ✨ 
This is more of a design question, but I think it would be cool if there was a smooth transition between the expanded/ non expanded states. Something to consider maybe as a potential future enhancement.

src/ExpandableContainer/ExpandableContainer.tsx
dengjulie: wow, good catch, you're an expert on accessibility! I'm more than happy to do that, but this also got me thinking, maybe title should be required, since I don't imagine this expandable container being too useful without the title. What do you think?

docs/components/ExpandableContainerView.jsx
dengjulie: 🔪

package.json
dengjulie: I skipped one because of a separate PR I was planning on merging first, but then John beat me to it so now I have to bump both PRs :shakesfist:

src/ExpandableContainer/ExpandableContainer.less
dengjulie: oh right there is text--semi-bold, but then I realized that in the mocks for the announcements sidebar, the text isn't bolded, so i'm going to remove this line and leave the font styling as a consumer decision

src/ExpandableContainer/ExpandableContainer.tsx
clcaelynn-clever: Yeah, I think that's actually the better route to go! I, too, can't see much of a use case for no title, and now the consumer can't accidentally create an inaccessible instance of the component. 🙂

clcaelynn-clever (approved):
🚀 🎉

PR #599, opened 2021-03-03
Improve build times

kofi-clever:
**Jira:**
https://clever.atlassian.net/browse/JULIE-1

**Overview:**
- @dengjulie promised me that if I make Dewey build faster, she'd come join DEIP with me.
- So, I spent a couple hours updating our dev server build process, evil-laughing to myself the whole time 😈
- From a brief run-in with the Dewey build last year, I'd noticed compiled CSS wasn't getting cached anymore, so I'd assumed that was the issue.
- On closer inspection, it turned out not very much of the webpack output was getting cached, so every code change required rebuilding almost the whole damn thing 😮
- So, I yanked the YOLO cord and just went full update on all our webpack libs and plugins.
- Also moved `webpack.config.js` to `webpack.config.ts`, because types.
- Turned on hot reloading as well, for a little extra kick.
- Looks like (on my machine) rebuild time after code change is down from ~30s to ~1s.
- I'll accept cupcakes as thank-yous.

**NOTE:** These changes only affect the dev build process (and build process for the docs on go/dewey) - all compilation for the NPM package remains unchanged.

**Screenshots/GIFs:**
https://gyazo.com/d1d84d24f323d02fae049814efd03702

**Testing:**
- Verified that a fresh build of the the index.html and bundle.js to the `/docs` output folder runs without error (outside of webpack-dev-server)
- [n/a] Unit tests
- Manual tests:
  - [n/a] Chrome
  - [n/a] Safari
  - [n/a] IE11

**Roll Out:**
:100:

- Before merging:
  - [n/a] Updated docs
  - [n/a] Bumped version in `package.json`
    - Breaking change?
      - If it is a beta component run `npm version minor`
      - If the component is not in beta run `npm version major`
    - New component or backward-compatible component feature change? Run `npm version minor`
    - Only changing documentation? All good. Skip this step.
  - After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [n/a] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
        componentLink: "<COMPONENT LINK>",
        componentImg: "<COMPONENT LINK>.png",
        componentName: "<COMPONENT NAME>",
        componentImgAlt: "A <COMPONENT NAME> component",
      },
      ```
    - [n/a] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:
  - [ ] Deployed updated docs (`make deploy-docs`)
[...]

PR #645, opened 2021-04-27
don't use async in View files

johnhuangclever:
**Overview:**
In moving to reduce build times, it looks like we removed this babel plugin that supported async/await (https://github.com/Clever/components/pull/599/files#diff-379120de71991cfea15f15a1225d1ba83a7a50e55a356c831f5f5e0be305ce0bL16). This is causing an error with async/await calls in the TableView.jsx and Table2BetaView.jsx

It looks like we're using typescript to compile/build the main components so this would only affect View components, so as a workaround right now, I've just convert the async/await calls to return a Promise. The reason why we haven't been able convert the View files to typescript/tsx is because we rely on some special babel transform to render our `<Example>` components (https://github.com/Clever/components/blob/d94a59a973bab8c717eebe9c0ef6461e2f366be7/transform-code-string.js). In theory we should be able to have ts-loader load this custom babel transform, and then move completely to typescript, but for now I've just remove the async/await code in the View.

In the long term, I would want to take a look at ts-loader using the babel plugin, which would also give us all the niceties of type safety and stuff in the View files.

**Screenshots/GIFs:**
image
image

Verified that lazy loading Tables work:
image

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/docs/components/Table2BetaView.jsx
+++ b/docs/components/Table2BetaView.jsx
@@ -612,13 +612,14 @@ export default class Table2BetaView extends React.PureComponent {
               this._lazyTable = t;
             }}
             lazy
-            getData={async ({ startingAfter, pageSize }) => {
+            getData={({ startingAfter, pageSize }) => {
               let start = 0;
               if (startingAfter != null) {
                 start = _.findIndex(tableData, (r) => r.id === startingAfter) + 1;
               }
-              await new Promise((resolve) => setTimeout(resolve, 1000));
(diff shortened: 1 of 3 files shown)


dxu429 (approved):
Lgtm

PR #647, opened 2021-04-28
AlertBox2 sub text

johnhuangclever:
**Overview:**
https://clever.atlassian.net/wiki/spaces/DES/pages/1440153759/Banners+Component+Spec+formerly+Global+Alerts
https://www.figma.com/file/JliXKTlDzqAWO9oypH1Tuz/Dewey-design-workshop-templates?node-id=87%3A7833

According to the spec, we want to keep alerts as much as a single line bolded alert message as possible. However, talking to Georgia, she mentioned that for IDM we may need some extra non-bolded text since our product is so complicated.

So I added some support for non-bolded text as <p> tags in the children. The catch here is that adding any extra buttons to the content throws the spacing out of whack because the Dewey 1.0 buttons have a larger height than the line height of text. So I added buttons as an extra prop so we can adjust the layout if they are provided.

**Screenshots/GIFs:**

https://user-images.githubusercontent.com/13126257/116480285-e5f98480-a835-11eb-8ae5-fde98879b5af.mov

**Testing:**

- [] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
      ```
      {
      },
      ```
- After merging:

### DIFF
--- a/src/AlertBox2/AlertBox2.less
+++ b/src/AlertBox2/AlertBox2.less
@@ -55,7 +55,6 @@
 
   & p {
     .text--regular();
-    .margin--top--2xs();
   }
 }
 
--- a/src/AlertBox2/AlertBox2.less
+++ b/src/AlertBox2/AlertBox2.less
@@ -5,7 +5,6 @@
   .padding--m();
   .flexbox();
   .justify--between();
-  .items--center();
 
   &--critical {
     background-color: fade(@alert_red, 15%);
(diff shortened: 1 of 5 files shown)


src/AlertBox2/AlertBox2.less
```diff
@@ -42,15 +41,27 @@
...
 
 .AlertBox2--content {
-  .flex--grow();
   .text--semi-bold();
   color: @neutral_dark_gray;
   .margin--left--m();
+
+  & p {
+    .text--regular();
+    .margin--top--2xs();
```
jonjonray, line 58: I think I understand the rationale as to why we want to update the text for `<p>` elements but wondering if there are any concerns around potentially adding top margin for every paragraph element vs only doing so for direct descendants (i.e. `& > p`)

jonjonray: I imagine this shouldn't be an issue if we conform to the rule of only having a single line for an alert but  I can see a world where a non-direct child element might not need this extra spacing (e.g. a modal that is being opened within the alert but is fixed to the page rather than the alert box) and could be confusing while developing

jonjonray (approved):
Left a comment around a <p> css selector but LGTM!

src/AlertBox2/AlertBox2.less
johnhuangclever: oh yeah, thanks. I can leave it up to the caller to specify spacing 👍

PR #634, opened 2021-04-23
add searchKey for multi-select options

johnhuangclever:
**Overview:**
Add optional `searchKey` for multi select options

The use case where we needed this was I have `value: "name.first"` which displays as `first name` and searching by `first name` doesn't show the result.

**Screenshots/GIFs:**
before:
image

after:
image

**Testing:**

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [x] Updated docs
  - [x] Bumped version in `package.json`
    - Breaking change?
      - If it is a beta component run `npm version minor`
      - If the component is not in beta run `npm version major`
    - New component or backward-compatible component feature change? Run `npm version minor`
    - Only changing documentation? All good. Skip this step.
  - After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [ ] Add an entry in `ComponentsView.componentsToDisplay` using this template:
      ```
      {
        componentLink: "<COMPONENT LINK>",
        componentImg: "<COMPONENT LINK>.png",
        componentName: "<COMPONENT NAME>",
        componentImgAlt: "A <COMPONENT NAME> component",
      },
      ```
    - [ ] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT LINK>.png`
- After merging:
  - [ ] Deployed updated docs (`make deploy-docs`)
  - [ ] Posted in #eng if I made a breaking change to a beta component

PR #650, opened 2021-05-06
Select2 string values

johnhuangclever:
## Overview
I have 2 main points here. Would love your thoughts/comments:

### Point 1
* For Select2 and Multiselect, the base `option.value` may not always represent the display value, so we need a reactContent type of prop for display purposes in the options
* The `option.value` may not always represent the searchable value, and if the custom react rendered value doesn't match the value, then it may be confusing for the user (addressed here https://github.com/Clever/components/pull/634)
  * For example, we have `name.first` as the value, but we need `First name` with an icon as the display and the user needs to be able to search for `First name`, and `name.` should not show any results
* for accessibility purposes, downshift also requires only a string representation of the option value to be put into the `<input>` tag
* ~My current hacky implementation is to determine an option's string value based on an implicit order, but it's kind of janky~
  * EDIT: updated with new code

### Point 2 
Taking a step back and looking what the ideal state would be, I think that leads me back to the more common triple prop set:
* label, or stringValue (required)
* key, or value (required)
* customLabel (optional)

That would allow us to have a base string key and value, with an optional react rendered prop where it is clear that it is for display purposes only. If that sounds good I will take a stab at that type of implementation.

Open to other thoughts if you have any and also other naming options here.

### Bonus change
I made it so that the create option for multi-select shows even if there are other results. The use case I need this for is something like `sis_id` exists, and I want to add `id` to the options. I wouldn't be able to in the previous mode

## Screenshots/GIFs:
image
image

image

## Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/docs/components/MultiSelectView.jsx
+++ b/docs/components/MultiSelectView.jsx
@@ -70,14 +70,14 @@ export default class MultiSelectView extends React.PureComponent {
               hideLabel={this.state.hideLabel}
               placeholder={this.state.placeholder}
               options={new Array(20).fill(0).map((_, i) => ({
-                value: `option_${i + 1}`,
-                content: (
+                label: `option ${i + 1}`,
+                customLabel: (
                   <FlexBox>
                     <FontAwesome style={{ marginRight: "8px" }} name="exclamation-triangle" />
                     <span>Option {i + 1}</span>
                   </FlexBox>
                 ),
-                searchKey: `option ${i + 1}`,
+                value: `option_${i + 1}_value`,
               }))}
               creatable={this.state.creatable}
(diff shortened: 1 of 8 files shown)


prime-time:
This seems reasonable to me 👍

johnhuangclever:
ok I implemented the proposed option prop changes. Gonna merge tomorrow
Appreciate any retroactive comments if you have any

PR #651, opened 2021-05-06
wrap the icon in a div to make fixedWidth work

johnhuangclever:
**Overview:**
I think this is because FontAwesome uses a `::before` psuedo-element to render the icon, so wrapping it in a div helps, but I'm not 100% sure

**Screenshots/GIFs:**
before:

https://user-images.githubusercontent.com/13126257/117246276-369e4e00-adf1-11eb-9b2c-c4f3843a9dec.mov

after:

https://user-images.githubusercontent.com/13126257/117246293-3b630200-adf1-11eb-8b76-de44bf389fac.mov

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AlertBox2/AlertBox2.tsx
+++ b/src/AlertBox2/AlertBox2.tsx
@@ -60,11 +60,13 @@ export const AlertBox2: React.FC<Props> = ({ children, className, type, buttons,
           !!buttons && cssClass.CONTENT_CONTAINER_WITH_BUTTONS,
         )}
       >
-        <FontAwesome
-          className={classnames(cssClass.ICON, `AlertBox2--icon--${type}`)}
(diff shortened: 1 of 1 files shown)


tuxagon:
If I'm understanding the change, the goal is to reduce the width taken up by the icon on the banner, as well as, put a `<div>` where the icon was so that it appears at the top of its container?

johnhuangclever:
The icons are separate widths, so we just want to make sure that the icons and text are in the same position for every icon type

tuxagon (approved):

PR #652, opened 2021-05-07
fix ordering of insertion of autogenerated code

johnhuangclever:
Follow up to https://github.com/Clever/components/pull/595

There's a bug with the component ordering, since there are `N` existing entries in a list with `N+1` places to insert them.

Your comment here was very wise (https://github.com/Clever/components/pull/595#discussion_r602608162) and I once again don't have time to write them but I'm very interested in doing so later I promise 🙏


johnhuangclever:
yolo

PR #654, opened 2021-05-12
Vc/fix menuitem default type

vchunduru93:
MenuItem was defaulting `anchor` if the item was not disabled, but we want it to default to `button`.


nbhatia823 (approved):
Awesome!

PR #655, opened 2021-05-13
M5G-292: MessagingAttachment component

jonah-clever:
**Jira:** 

https://clever.atlassian.net/browse/M5G-292

**Overview:**

This is the composable component `MessagingAttachment`. This component will later be used in `MessagingBubble` and `MessaingInput`, among other places. 

The component takes in props such as title, subtitle, icon, etc. which will be set by consumers and can have changing values based on state (uploading, uploaded, error, etc).

Notes:
* There is a hover state, which is visible for uploaded attachments (`isUpload && uploadComplete == true`) and for sent attachments (`isUpload == false`).
* There is a subcomponent <FileAttachmentIcon /> which takes in a `fileType` prop and renders the icon for the given file. Generally this will be used for  MessagingAttachment's `icon` prop. It also takes in `isUpload` and `uploadError` props which render a loading spinner and an error icon, respectively. Other `React.ReactNode`s can also be passed in for MessagingAttachment's `icon` prop

**Screenshots/GIFs:**

Screen Shot 2021-05-13 at 1 45 26 PM

Hover sent attachment
Screen Shot 2021-05-13 at 1 45 33 PM

Hover uploaded draft attachment
Screen Shot 2021-05-13 at 1 45 37 PM

Hover remove attachment button (does not cause attachment to enter hover state)
Screen Shot 2021-05-13 at 1 45 41 PM

**Testing:**

- [ ] Unit tests
  - I did not write unit tests for this component (yet?)
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11
  - [x] Edge

**Roll Out:**

- Before merging:
  - [x] Updated docs
    - New component: Run `npm version minor`
  - [x] After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [x] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT URL LINK>.png`
- After merging:


clcaelynn-clever:
This is looking great! What does the axe extension's accessibility audit functionality think of the component?

jonah-clever:
> This is looking great! What does the axe extension's accessibility audit functionality think of the component?

Ah yes, good call. I will do that after I eat (late lunch)

docs/components/ComponentsView.jsx
```diff
@@ -367,6 +369,12 @@ ComponentsView.componentsToDisplay = [
     componentName: "WizardLayout",
     componentImgAlt: "A WizardLayout component",
   },
+  {
+    componentLink: "messaging-attachment",
+    componentImg: "messaging-attachment.png",
+    componentName: "MessagingAttachment",
+    componentImgAlt: "A MessagingAttachment component",
+  },
```
clcaelynn-clever, line 377: Even though it's not strictly necessary since you added the sort up above, might still be nice to slot this in alphabetically? Just a nit, not blocking.

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+    return (
+      <>
+        <PropDocumentation
+          title="<MessagingAttachment /> Props"
+          availableProps={[
+            {
+              name: "attachmentID",
+              type: "string",
+              description: "Unique ID used to identify this attachment",
+              // optional: false,
```
clcaelynn-clever, line 167: Are the commented `optional: false` lines just there as a reminder?

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+            },
+          ]}
+          className={cssClass.PROPS}
+        />
+        <PropDocumentation
+          title="<FileAttachmentIcon /> Props"
+          availableProps={[
+            {
+              name: "fileType",
+              type: '"audio" | "catchall" | "doc" | "image" | "pdf" | "ppt" | "video" | "xls"',
```
clcaelynn-clever, line 237: Which icon is the `"catchall"` for? What exactly does it cover? 😮

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+  CONFIG: "MessagingAttachmentView--config",
+  CONFIG_LABEL_TEXT: "TextInput2View--configLabelText",
+  CONFIG_TOGGLE: "MessagingAttachmentView--configToggle",
+  CONTAINER: "MessagingAttachmentView",
+  INTRO: "MessagingAttachmentView--intro",
+  PROPS: "MessagingAttachmentView--props",
+  DEMO_ROW: "MessagingAttachmentView-demoRow",
+};
+
+// TODO: FINISH THIS DOCUMENT
```
clcaelynn-clever, line 23: What's left to be finished, out of curiosity?

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+            </FlexBox>
+          </ExampleCode>
+        </Example>
+
+        {this._renderProps()}
+      </View>
+    );
+  }
+
+  // TODO: Update prop documentation.
```
clcaelynn-clever, line 155: Is there still something to be added to the documentation, or is this todo now completed?

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -0,0 +1,188 @@
+@import (reference) "../less/utilities";
+
+// @import (reference) "~clever-components/dist/less/index";
```
clcaelynn-clever, line 3: Turned out to be unnecessary?

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -0,0 +1,188 @@
...
+
+@ICON_SMALL_SIZE: @size_2xl;
+@ICON_REGULAR_SIZE: @size_2xl;
+@HOVER_BACKGROUND_COLOR: #f9faff;
+@ERROR_RED: #df3537;
+@ERROR_RED_TEXT: #b30c10;
+@WHITE: #fff;
```
clcaelynn-clever, line 10: This one's `@neutral_white` from /src/less/colors.less.
https://github.com/Clever/components/blob/master/src/less/colors.less#L23

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -0,0 +1,188 @@
...
+  border-radius: 50%;
+  cursor: pointer;
+  background-color: @neutral_medium_gray;
+  color: @WHITE;
+  border: none;
+  padding: 0;
+  border: @size_3xs solid white;
+
+  // TODO: doesn't turn off after click...
+  // maybe this doesn't matter cause clicking will close the box?
```
clcaelynn-clever, line 83: Does this mean that the state persists after the click? Hm, since we expect the click to always _do something_, I think this might actually be fine, as you noted - the focus moves elsewhere anyway. 🤔

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -0,0 +1,188 @@
...
+    padding-left: 0.6875rem; // 11px
+    padding-right: 0.6875rem; // 11px
+
+    .MessagingAttachment--TextContainer {
+      max-width: 10.375rem; // 166px
+    }
+  }
+}
+
+// TODO: :not() does not work for IE
```
clcaelynn-clever, line 147: I assume this doesn't actually impede functionality in any way? Not ideal then, but we already have some serious styling issues on IE if I'm remembering right, so I think something this small may very well be fine.

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -0,0 +1,141 @@
...
+  onClickAttachment: (attachmentID: string) => void;
+  onRemoveAttachment?: (attachmentID: string) => void;
+  title: string;
+  subtitle?: string;
+  isUpload?: boolean;
+  uploadComplete?: boolean;
+  uploadError?: boolean;
+};
+
+// TODO: can this not be functional?
```
clcaelynn-clever, line 27: Not sure I understand this TODO? It's already a functional component, after all. 😮

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -0,0 +1,141 @@
...
+      alt={`MessagingAttachment ${fileType} icon`}
+    />
+  );
+}
+
+// TODO: how do I import this? It exists in LP
+// TODO: is this necessary? Helpful?
+// export function LoadingSpinnerIcon() {
+//   return <LoadingSpinner size={"s"} />;
+// }
```
clcaelynn-clever, line 141: Seems like it wasn't necessary in the end? Consumer is always free to use their preferred loading-spinner version as the icon.

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+              optional: true,
+            },
+            {
+              name: "uploadError",
+              type: "boolean",
+              description:
+                "Optional boolean for whether there was an error while uploading the Attachment",
+              defaultValue: "false",
+              optional: true,
+            },
```
clcaelynn-clever, line 228: This totally works, but what's the advantage over the `errorMsg?: string` approach mentioned in the spec, where we use something like `!!errorMsg` as the boolean check for if the component is in an error state, and `errorMsg` itself is the error message we display, rather than it being fed in through `title`?

Seems like having `errorMsg` means we don't need `uploadError`, basically.

docs/components/MessagingAttachmentView.jsx
```diff
@@ -0,0 +1,264 @@
...
+        subtitle: "Click to download",
+        isUpload: true,
+        uploadComplete: true,
+      },
+      {
+        attachmentID: "4",
+        icon: <FileAttachmentIcon fileType={"audio"} isUpload uploadError />,
+        onClickAttachment: () => console.log("clicked!"),
+        onRemoveAttachment: () => console.log("removed!"),
+        title: "File cannot exceed 10 MB",
```
clcaelynn-clever, line 86: Shouldn't this be `errorMsg` instead of `title`?

clcaelynn-clever (commented):
This looks awesome, great work! Left some comments/questions, but largely just small stuff! 🥳

docs/components/ComponentsView.jsx
jonah-clever: I updated it to sort alphabetically, so it doesn't matter where it is in the list https://github.com/Clever/components/pull/655/files/e0cf0ccd55f088b01f105721881f0b1bb29201b4#diff-bfd693f512631919c327aa9c7211594ea3ac21bd5c114f4bb8e729ff7b2310d5R14

docs/components/MessagingAttachmentView.jsx
jonah-clever: Screen Shot 2021-05-13 at 3 36 57 PM

This one, its just like a fallback

docs/components/MessagingAttachmentView.jsx
jonah-clever: nada

docs/components/MessagingAttachmentView.jsx
jonah-clever: done

src/MessagingAttachment/MessagingAttachment.less
jonah-clever: I think it's because it's focussed after clicking. I could remove the styling for :focus, but I think that's worse for accessibility

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: I guess I meant just the different way to write it, using `function` but you're right

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: I actually made it render the loading spinner if `isUpload == true`

docs/components/MessagingAttachmentView.jsx
jonah-clever: Genius

src/MessagingAttachment/MessagingAttachment.less
jonah-clever: It only effects the hover states. I think I can remake it without the not though

jonah-clever: It would just be some redundant css

jonah-clever:
> This is looking great! What does the axe extension's accessibility audit functionality think of the component?

Ok I went through and made a couple changes for axe

docs/components/ComponentsView.jsx
clcaelynn-clever: Yeah, I just noticed the list was already alphabetical (I think?) except for this addition, so I couldn't help pointing it out - the urge to alphabetize is so strong. 😅

docs/components/MessagingAttachmentView.jsx
```diff
@@ -83,7 +82,8 @@ export default class MessagingAttachmentView extends React.PureComponent {
         icon: <FileAttachmentIcon fileType={"audio"} isUpload uploadError />,
         onClickAttachment: () => console.log("clicked!"),
         onRemoveAttachment: () => console.log("removed!"),
-        title: "File cannot exceed 10 MB",
+        title: "My cool file.txt",
```
clcaelynn-clever, line 87: This kind of makes me think `title` should actually be optional, since it's unnecessary when there's an error? I think ideally we'd have an Error version of this component (a la AnnouncementBubble's structure), but that seems like overkill just for this.

We can add it in Milestone 2, when we'll be adding the full-bleed image preview version of this component.

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -43,6 +41,7 @@ export const MessagingAttachment: React.FC<Props> = ({
         <button
           className={cssClass("CloseButton")}
           onClick={(e) => handleRemoveClick(e, onRemoveAttachment, attachmentID)}
+          aria-label={`close attachment with title: "${title}" and ID: ${attachmentID}`}
```
clcaelynn-clever, line 42: Awesome to see, thank you! Might remove the `and ID: ${attachmentID}` portion though? We don't expose that in the visuals of the UI, so no need here - don't think having a whole UUID read by the screen-reader will actually help, just the title seems like plenty.

clcaelynn-clever (changes requested):
Awesome! Just a couple tiny changes, should be nice and quick. 🙌🏻

docs/components/MessagingAttachmentView.jsx
jonah-clever: I was thinking the same

docs/components/ComponentsView.jsx
```diff
@@ -213,6 +213,12 @@ ComponentsView.componentsToDisplay = [
     componentName: "Menu",
     componentImgAlt: "A Menu component",
   },
+  {
+    componentLink: "messaging-attachment",
+    componentImg: "messaging-attachment.png",
+    componentName: "MessagingAttachment",
+    componentImgAlt: "A MessagingAttachment component",
+  },
```
clcaelynn-clever, line 221: My heart is genuinely at ease. 😂

PR #660, opened 2021-05-18
Add ariaLabelledBy prop to RadioGroup

dengjulie:
**Jira:**
n/a

**Overview:**
Sometimes a label is not needed, because the label exists elsewhere, like in our new language selector page of family-portal 
image

This PR adds an `ariaLabelledBy` prop to RadioGroup that can be used instead. 

Also adds documentation:

**Screenshots/GIFs:**
image

image

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.111.1",
+  "version": "2.112.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/RadioGroup/RadioGroup.tsx
+++ b/src/RadioGroup/RadioGroup.tsx
@@ -16,6 +16,7 @@ interface Option<IDType extends string = string, ValueType = any> {
 }
 
 export interface Props<OptionIDType extends string, OptionValueType> {
+  ariaLabelledBy?: string;
   className?: string;
   disabled?: boolean;
   error?: React.ReactNode;
@@ -46,8 +47,24 @@ export default class RadioGroup<
   _optionRefsByID = {};
 
   render() {
-    const { className, disabled, error, label, onChange, options, selectedID } = this.props;
+    const {
+      ariaLabelledBy,
+      className,
+      disabled,
+      error,
+      label,
+      onChange,
+      options,
+      selectedID,
+    } = this.props;
 
+    if (ariaLabelledBy && label) {
+      console.warn(
+        "You should not pass both `ariaLabelledBy` and `label` props. The `ariaLabelledBy` prop is ignored.",
+      );
+    }
+
+    const labelID = label || !ariaLabelledBy ? this._labelID : ariaLabelledBy;
     const focusableOptionID = this._getFocusableOptionID(options);
 
     return (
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.111.1",
+  "version": "2.111.2",
```
adamvictorclever, line 3: Nit: It doesn't _really_ matter much, but technically according to [semver](https://semver.org/) this should be a minor version bump since it introduces new, backwards compatible, functionality.

src/RadioGroup/RadioGroup.tsx
```diff
@@ -57,11 +68,11 @@ export default class RadioGroup<
         onChange={this._handleNavChange}
       >
         <div
-          aria-labelledby={this._labelID}
+          aria-labelledby={labelID}
           className={classnames(cssClass.CONTAINER, className)}
           role="radiogroup"
         >
-          <div className={cssClass.LABEL} id={this._labelID}>
+          <div className={cssClass.LABEL} id={labelID}>
```
adamvictorclever, line 75: I think we don't want to display this `div` in the case that we're using `aria-labelledby`? Otherwise it seems like we're going to end up with two elements with the same `id` (since we're giving this `div` the `id` of `ariaLabelledBy`) which seems like an issue.

package.json
dengjulie: yeah I guess it should be considered new functionality, it was between a bug fix and new functionality to me

src/RadioGroup/RadioGroup.tsx
dengjulie: ah good catch! fixed! I also added some mechanisms to prevent label and ariaLabelledBy from being passed in together, and made sure to keep it backward compatible. added an example in the documentation too

src/RadioGroup/RadioGroup.tsx
```diff
@@ -46,8 +47,24 @@ export default class RadioGroup<
...
+      error,
+      label,
+      onChange,
+      options,
+      selectedID,
+    } = this.props;
+
+    if (ariaLabelledBy && label) {
+      console.warn(
+        "You should not pass both `ariaLabelledBy` and `label` props. The `ariaLabelledBy` prop is ignored.",
```
adamvictorclever, line 63: We could try to do this check at compile time with a pattern along the lines of https://gist.github.com/brian-mcallister-lab49/02c43693d5251daa6e0e7335bed950e2... but it's probably not necessary. I think at least a runtime warning is fine.

adamvictorclever (approved):
LGTM! 🎉

src/RadioGroup/RadioGroup.tsx
dengjulie: Oh that's a neat trick, but yeah I think this should be fine as it is since the outcome of having both props passed in is not bad and still accessible

package.json
dengjulie: Oh nice thanks for explaining this is good to know for the future

PR #662, opened 2021-05-19
M5G-295: messaging input attachment FileInput (initial version for testing, not finished)

jonah-clever:
**Jira:**

https://clever.atlassian.net/browse/M5G-295

**Overview:**

Adding a `FileInput` to `MessagingInput` for uploading attachments. This is just a preliminary version for Portals team to be able to start testing. Clicking the `FileInput` opens a file chooser menu, but the store function does not do anything as of now.

**Screenshots/GIFs:**

Screen Shot 2021-05-19 at 2 59 12 PM

**Testing:**

- [ ] Unit tests
  -  I did not add new unit tests, but existing tests pass
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11
  - [x] Edge

**Roll Out:**

- Before merging:
  - [x] Updated docs
- After merging:

### DIFF
--- a/docs/components/FileInputView.jsx
+++ b/docs/components/FileInputView.jsx
@@ -170,13 +170,6 @@ export default class FileInputView extends React.PureComponent {
               optional: true,
               defaultValue: <code>FormElementSize.FULL_WIDTH</code>,
             },
-            {
-              name: "isMessagingAttachmentInput",
-              type: "boolean",
-              description:
-                "Optional boolean which dictates if this FileInput is being used for uploading message attachments. If so, the component is rendered specifically for that purpose",
-              optional: true,
-            },
           ]}
           className={cssClass.PROPS}
           title="FileInput"
--- a/docs/components/FileInputView.jsx
+++ b/docs/components/FileInputView.jsx
@@ -170,6 +170,13 @@ export default class FileInputView extends React.PureComponent {
               optional: true,
               defaultValue: <code>FormElementSize.FULL_WIDTH</code>,
             },
+            {
+              name: "isMessagingAttachmentInput",
+              type: "boolean",
+              description:
+                "Optional boolean which dictates if this FileInput is being used for uploading message attachments. If so, the component is rendered specifically for that purpose",
+              optional: true,
+            },
           ]}
           className={cssClass.PROPS}
           title="FileInput"
--- a/docs/components/MessagingAttachmentView.jsx
+++ b/docs/components/MessagingAttachmentView.jsx
@@ -118,7 +118,30 @@ export default class MessagingAttachmentView extends React.PureComponent {
           <ExampleCode>
             <p>Sent Attachments:</p>
             <FlexBox className={cssClass.DEMO_ROW}>
-              {sentAttachments.map((attachment) => (
+              {[
+                {
+                  attachmentID: "1",
+                  icon: <FileAttachmentIcon fileType={"doc"} />,
+                  onClickAttachment: () => console.log("clicked!"),
+                  title:
+                    "MyCoolDoclajsdjasldjaslkdjasldkjasldjaslkdjasldjasldjalskjdalskjdaslkjasljd.doc",
+                  subtitle: "Click to download",
+                },
+                {
+                  attachmentID: "1",
+                  icon: <FileAttachmentIcon fileType={"image"} />,
+                  onClickAttachment: () => console.log("clicked!"),
+                  title: "Flyer.png",
+                  subtitle: "Click to view",
+                },
+                {
+                  attachmentID: "1",
+                  icon: <FileAttachmentIcon fileType={"audio"} />,
+                  onClickAttachment: () => console.log("clicked!"),
+                  title: "Morning message.m4a",
+                  subtitle: "Click to download",
+                },
+              ].map((attachment) => (
                 <MessagingAttachment
                   attachmentID={attachment.attachmentID}
                   icon={attachment.icon}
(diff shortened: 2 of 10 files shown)


zohnnyclever:
oh nice! @dengjulie do you know how we might use this change to do a test in our mobile app?

docs/components/MessagingAttachmentView.jsx
```diff
@@ -24,71 +24,6 @@ export default class MessagingAttachmentView extends React.PureComponent {
   static cssClass = cssClass;
```
jonah-clever, line 24: note: changes in this file are unrelated to the rest of the PR. Just some cleanup from recent changes to this component

docs/components/FileInputView.jsx
```diff
@@ -170,6 +170,13 @@ export default class FileInputView extends React.PureComponent {
               optional: true,
               defaultValue: <code>FormElementSize.FULL_WIDTH</code>,
             },
+            {
+              name: "isMessagingAttachmentInput",
+              type: "boolean",
+              description:
+                "Optional boolean which dictates if this FileInput is being used for uploading message attachments. If so, the component is rendered specifically for that purpose",
+              optional: true,
+            },
```
jonjonray, line 179: Wondering if we should avoid adding this to the docs for FileInput since we mostly use this prop internally from within MessagingInput

docs/components/TextAreaView.jsx
```diff
@@ -277,6 +277,20 @@ export default class TextAreaView extends React.Component {
...
+              optional: true,
+            },
+            {
+              name: "storeAttachment",
+              type: "(file, callbacks) => void",
+              description:
+                "Optional store function to be used with the FileInput for attachment uploads. This is only used if showUploadAttachmentButton == true, and should always be defined in that case.",
+              optional: true,
+            },
           ]}
```
jonjonray, line 294: same comment as above about potentially not documenting this new prop since we're using it internally

src/FileInput/FileInput.less
```diff
@@ -36,6 +36,26 @@
...
+  .FileInput--Icon {
+    margin: 0.25rem 0.5rem 0.25rem 0.75rem;
+    border-radius: 0.25rem;
+
+    &:hover {
+      fill: #f5f8fd;
+    }
+  }
+}
+
```
jonjonray, line 58: nit: might be worth using mixins for a few of these

height: @size_2xl
.padding--none()
.borderRadiusL()

src/FileInput/FileInput.tsx
```diff
@@ -163,9 +184,11 @@ export class FileInput extends React.Component<Props, State> {
   }
 
   render() {
-    const { iconOnly, label } = this.props;
+    const { iconOnly, label, isMessagingAttachmentInput } = this.props;
     const dropzoneStyle = iconOnly ? { display: "inline-block" } : {};
 
+    // console.log(this.state);
```
jonjonray, line 190: Nit: console.log

src/FileInput/FileInput.tsx
```diff
@@ -110,6 +130,7 @@ const propTypes = {
   store: PropTypes.func.isRequired,
   accept: PropTypes.string,
   size: PropTypes.string,
+  isMessagingAttachmentInput: PropTypes.bool,
```
jonjonray, line 133: Another thought here: wondering if there's an opportunity to potentially abstract this prop into something non-messaging related e.g. some kind of icon override since we're mainly checking this boolean so we can update the icon on line 203

src/MessagingInput/MessagingInput.tsx
```diff
@@ -31,6 +31,10 @@ interface Props {
   labelText?: string;
   disableSendButton?: boolean;
   showReturnKeyInstructions?: boolean;
+  // TODO: is this needed or is the existance of MessagingInput.store sufficient to decide this?
```
jonjonray, line 34: I have no strong preference, can see an argument to be made for both cases (more explicit vs less props)

jonjonray (approved):
Left some comments but LGTM overall

src/MessagingInput/MessagingInput.tsx
jonah-clever: I'm gonna go with more explicit for now, will remove comment

src/FileInput/FileInput.tsx
```diff
@@ -204,6 +204,10 @@ export class FileInput extends React.Component<Props, State> {
                 "FileInput",
                 formElementSizeClassName(this.props.size),
                 this.props.className,
+                !!this.state.filename &&
+                  !this.state.success &&
+                  !this.state.error &&
+                  "FileInput--AttachmentIsUploading",
```
jonjonray, line 210: It seems like the styles are only getting applied from the MessagingInput `.less` file but we would still end up adding an extra class for every usage of `FileInput`, which might not be ideal. 

Was thinking it might make sense to add this class to the top level component (MessagingInput) and apply it from there. Something like

```
.MessagingInput--AttachmentIsUpload .FileInput { ... }
```

jonjonray (approved):
LGTM! Left a comment about the upload class name, but it's somewhat minor

src/FileInput/FileInput.tsx
jonah-clever: I thought about this too. The issue is that the class is added here based on the state of the `FileInput`. I'm not sure exactly how to pass that info up to the `MessagingInput`, so the class could be set there. Open to ideas tho!

jonjonray: Ah yeah, that's a good point. The current approach makes sense then!

PR #666, opened 2021-05-27
M5G-365: attachment icon types

jonah-clever:
**Jira:**

https://clever.atlassian.net/browse/M5G-365

**Overview:**

Adding mapping from file extension to icon types for MessagingAttachment

**Screenshots/GIFs:**

Screen Shot 2021-05-27 at 2 59 14 PM

**Testing:**

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

**Roll Out:**

- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/MessagingAttachment/MessagingAttachment.tsx
+++ b/src/MessagingAttachment/MessagingAttachment.tsx
@@ -70,6 +70,59 @@ function handleRemoveClick(e, onRemoveAttachment, attachmentID) {
 
 // /// Sub-components, exported icon components to be used by consumers /// //
 
+export type AttachmentFileType =
+  | "pdf"
+  | "doc"
+  | "docx"
+  | "xls"
+  | "xlsx"
+  | "csv"
+  | "ppt"
+  | "pptx"
+  | "jpg"
+  | "jpeg"
+  | "gif"
+  | "png"
+  | "svg"
+  | "mp4"
+  | "mov"
+  | "wmv"
+  | "mp3"
+  | "wav"
+  | "m4a"
+  | "flv"
+  | "txt"
+  | null;
+
+function fileTypeToIcon(fileType: string): AttachmentIconType {
+  const mapFileTypeToIcon = {
+    pdf: "pdf",
+    doc: "doc",
+    docx: "doc",
+    xls: "xls",
+    xlsx: "xls",
+    csv: "catchall",
+    ppt: "ppt",
+    pptx: "ppt",
+    jpg: "image",
+    jpeg: "image",
+    gif: "image",
+    png: "image",
+    svg: "image",
+    mp4: "audio",
+    mov: "video",
+    wmv: "video",
+    mp3: "audio",
+    wav: "audio",
+    m4a: "audio",
+    flv: "video",
+    txt: "doc",
+    null: "catchall",
+  };
+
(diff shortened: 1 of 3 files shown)


src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -70,6 +70,63 @@ function handleRemoveClick(e, onRemoveAttachment, attachmentID) {
...
+    m4a: "audio",
+    flv: "video",
+    txt: "doc",
+    null: "catchall",
+  };
+
+  // TODO: is there a better/cleaner way to do this?
+  if (Object.keys(mapMimeToIcon).includes(mimetype)) {
+    return mapMimeToIcon[mimetype];
+  }
```
jonjonray, line 126: I'm wondering if it's possible for this to be false since we're expecting an `AttachmentMimetype` as the argument and all of those are mapped in `mapMimeToIcon` 

With that in mind, could we just do `return mapMimeToIcon[mimetype]` without the `if (Object.keys(mapMimeToIcon).includes(mimetype)) {...` and default return with the expectation that the consumer will pass it null for the catchall case?

jonjonray (approved):
LGTM

left a comment on the section you left a TODO on

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: So I made the props for this function `(mimetype: AttachmentMimetype | string)` so it would accept filetypes that are not included in the list `AttachmentMimetype`. The idea is that you can pass in any filetype and if it doesn't match one in the list, it defaults to "catchall".

jonjonray: ah missed that. So I still don't think it's currently possible for this function to receive an untyped string since the pop itself is of type `AttachmentMimetype` rather than string. If we want to support this flow, we may need to account for that in the `AttachmentIconProps` type definition

jonah-clever: ah good point! do you think it makes sense to make it work the way I'm suggesting? i.e. 

```
type AttachmentIconProps = {
  fileType: AttachmentFileType | String;
  ...
}
```

jonjonray: Could we just make it `fileType: string` since the consumer is going to be handling it's own list of filetypes?

jonah-clever: i guess so haha, ok I'll do that

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -70,6 +70,63 @@ function handleRemoveClick(e, onRemoveAttachment, attachmentID) {
...
+  return "catchall";
```
jonjonray: can probably simplify this to `return mapMimeToIcon[mimetype] || "catchall"`

jonjonray (approved):
LGTM

left one last comment on just simplifying the condition

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: genius 💡

PR #673, opened 2021-06-24
M5G-413 Attachments mobile refinements

cofauver:
## Overview:
 This PR updates the messaging attachment component to properly respond to the mobile setting. Here are the details as documented in [the figma](https://www.figma.com/file/HE5OffYDhp2nElIFahzYlt/Clever-Messages-v2-(Attachments)?node-id=430%3A22449): 

1. Ensure that attachments on mobile are a shorter height of 56px
2. Font on mobile attachments is adjusted to 14px
3. Adjust the padding within the attachment to match the (12px instead of 16px)
4. Ensure that the attachment width is responsive according to the sequence of mocks here. That is, the attachment and the message bubble should maintain the same width once the message bubble is small enough (because of the narrow screen size). **Note the gifs below aim to demonstrate this change, primarily**

While working on this PR, I also found an issue with the x icon on the uploaded attachment, which is fixed here.

The PR also updates the MessagingAttachment component documentation to properly label it as a **Beta** componenet.

#### Before:
image

ezgif com-gif-maker (2)

image

image

#### After
image

ezgif com-gif-maker (1)

image

image

## Testing:
Styling changes only, no unit tests.
- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

## Roll Out:

- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/MessagingAttachment/MessagingAttachment.tsx
+++ b/src/MessagingAttachment/MessagingAttachment.tsx
@@ -34,7 +34,9 @@ export const MessagingAttachment: React.FC<Props> = ({
   uploadComplete,
 }: Props) => {
   return (
-    <FlexBox className={cssClass("ParentContainer")}>
+    <FlexBox
+      className={cx(cssClass("ParentContainer"), isUpload && cssClass("ParentContainer--IsUpload"))}
+    >
       {onRemoveAttachment && (
         <button
           className={cssClass("CloseButton")}
--- a/docs/components/MessagingAttachmentView.jsx
+++ b/docs/components/MessagingAttachmentView.jsx
@@ -3,12 +3,13 @@ import * as React from "react";
 import Example, { CodeSample, ExampleCode } from "./Example";
 import PropDocumentation from "./PropDocumentation";
 import View from "./View";
-import { MessagingAttachment, FlexBox } from "src";
+import { Label, MessagingAttachment, FlexBox } from "src";
 import { FileAttachmentIcon } from "src/MessagingAttachment/MessagingAttachment";
 
 import "./MessagingAttachmentView.less";
 
 const cssClass = {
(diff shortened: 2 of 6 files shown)


src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -34,7 +34,9 @@ export const MessagingAttachment: React.FC<Props> = ({
   uploadComplete,
 }: Props) => {
   return (
-    <FlexBox className={cssClass("ParentContainer")}>
+    <FlexBox
+      className={cx(cssClass("ParentContainer"), isUpload && cssClass("ParentContainer--IsUpload"))}
```
jonjonray, line 38: Any reason why we're adding a new `ParentContainer--IsUpload`  class here? Seems like the only CSS selector for this class is also being applied to `ParentContainer`

jonjonray (approved):
Left a comment about an extra class name but LGTM!

src/MessagingAttachment/MessagingAttachment.tsx
cofauver: The only style being added here is added to `.MessagingAttachment--ParentContainer.MessagingAttachment--ParentContainer--IsUpload`, a more specific selector than `.MessagingAttachment--ParentContainer`. I add the `overflow: visible` property to this selector, which finds "uploaded" attachments (aka attachments in drafts in the input box), because we want to be able to see the x icon that flows out of the edge of the box. I need it to be more specific than `.MessagingAttachment--ParentContainer` because that has a style of `overflow:hidden` which helps most of our attachments not run off the edge of the page.

jonjonray: ah my bad, thought we were applying them separately, just realized it's chained

PR #676, opened 2021-07-01
M5G-430: AttachmentPreview

jonah-clever:
# Jira: 
[M5G-430](https://clever.atlassian.net/browse/M5G-430)

# Overview:

Implementing AttachmentPreview component, and adding an instance of this component to the MessagingAttachment component. Currently only image attachments will use the AttachmentPreview, but later other fileTypes will also show in an AttachmentPreview

---------------------------------------------------------------------------------------------------------------

_A note to weary travelers who chose to go any further: I beg of you, do not look upon the commit history. It is one of my life's greatest shames._

_Instead, please review the files as such._

_Firstly: The files related to AttachmentPreview_
* AttachmentPreviewView.jsx 
* AttachmentPreviewView.less
* ComponentsView.jsx
* SideBar.jsx
* docs.jsx
* AttachmentPreview.less
* AttachmentPreview.tsx
* src/AttachmentPreview/index.ts
* src/index.ts 

_Secondly: The files related to MessagingAttachment_
* MessagingAttachmentView.jsx 
* MessagingAttachment.tsx

_This path should ensure your safety, and my dignity. If you should choose to look upon the commit history, I can help you no further, and your fate is your own._ 

---------------------------------------------------------------------------------------------------------------

# Screenshots/GIFs:

Desktop
Screen Shot 2021-07-01 at 12 23 56 PM

Mobile
Screen Shot 2021-07-01 at 12 24 13 PM

For some reason, the height/width get funky at very small screen size (320x330px). Probably negligible for most users, but not great 🤔 
Screen Shot 2021-06-30 at 5 01 45 PM

# Testing:

- [ ] ~Unit tests~
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11
  - [x] Edge on BrowserStack, desktop and mobile

# Roll Out:

- Before merging:
  - [x] Updated docs
    - [x] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT URL LINK>.png`
- After merging:


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.124.0",
+  "version": "2.125.0",
```
jonah-clever, line 3: So... here's the thing. This PR does include breaking changes for MessagingAttachment (adding new required props). The very next thing I'm going to do is to go into LP and add those required props to every instance of MessagingAttachment, but they're still breaking. Because of this, the PR template suggests `npm version major`, but I didn't do that... Should I do that?

clcaelynn-clever: MessagingAttachment is a beta component (as I believe all Messaging components are), so we're allowed to make breaking changes with only a minor version bump! 🎉

docs/components/AttachmentPreviewView.jsx
```diff
@@ -0,0 +1,136 @@
...
+  render() {
+    const { showingPreview } = this.state;
+
+    const file = {
+      id: "12345",
+      title: "Raccooooooon.jpg",
+      fileType: "jpg",
+      url:
+        "https://external-content.duckduckgo.com/iu/?u=http%3A%2F%2Fppcorn.com%2Fus%2Fwp-content%2Fuploads%2Fsites%2F14%2F2016%2F01%2Fcute-raccoon-ppcorn.jpg&f=1&nofb=1",
+    };
```
clcaelynn-clever, line 32: This is amazing, haha. How confident are we that it's not going to break at some point though, as opposed to something sadly far more boring (e.g. from the Clever branding page)?

docs/components/AttachmentPreviewView.jsx
```diff
@@ -0,0 +1,136 @@
...
+            defaultValue: "close attachment preview",
+            optional: true,
+          },
+          {
+            name: "downloadButtonText",
+            type: "string",
+            description: "Optional text for the download button",
+            defaultValue: "Download",
+            optional: true,
+          },
```
clcaelynn-clever, line 115: Technically, if this is optional, then the button would actually require an `aria-label`. We can either make this text required (which seems alright?) or add an optional aria label prop as well.

I think the former might be easier, unless we have reason to expect a no-text design for this sometime soon!

docs/components/AttachmentPreviewView.jsx
```diff
@@ -0,0 +1,136 @@
...
+          {
+            name: "onClickDownload",
+            type: "(attachmentID: string) => void",
+            description: "Function to run on clicking the download button",
+          },
+          {
+            name: "onClose",
+            type: "() => void",
+            description: "function to run on clicking the close button",
```
clcaelynn-clever, line 129: literally the most nit of nits: `f` -> `F`? 😛

docs/components/AttachmentPreviewView.jsx
```diff
@@ -0,0 +1,136 @@
...
+            name: "className",
+            type: "string",
+            description: "Optional class name",
+            optional: true,
+          },
+          {
+            name: "closeButtonAriaLabel",
+            type: "string",
+            description: "Optional ARIA label for the close button",
+            defaultValue: "close attachment preview",
```
clcaelynn-clever, line 106: Nice! Since it's hard to conceive of more than one of these being open at a time, a default value is a great idea. 🙌🏻

src/AttachmentPreview/AttachmentPreview.less
```diff
@@ -0,0 +1,195 @@
+@import (reference) "../less/utilities";
+
+@HEADER_BAR_HEIGHT: @size_5xl;
+@DESKTOP_BUFFER_SIZE: @size_l;
+@MOBILE_BUFFER_SIZE: @size_xs;
+
+// TODO: can I add this to sizing.less?
```
clcaelynn-clever, line 7: I think we'd need to check with Design first, as I think our pantheon of constants is kind of sacred.

src/AttachmentPreview/AttachmentPreview.less
```diff
@@ -0,0 +1,195 @@
...
+  left: 0;
+  width: 100%;
+  height: @HEADER_BAR_HEIGHT;
+  .alignItems(center);
+  color: @neutral_white;
+}
+
+.AttachmentPreview--AttachmentName {
+  line-height: @size_l;
+  // TODO: why isn't this adding the ellipsis?
```
clcaelynn-clever, line 34: I think we resolved a similar issue elsewhere recently, but I can't quite remember what the details were!

Think I recall that @cofauver and @nbhatia823 were involved? 🤔

src/AttachmentPreview/AttachmentPreview.less
```diff
@@ -0,0 +1,195 @@
...
+  .text--truncate();
+}
+
+.AttachmentPreview--FileIcon {
+  max-height: @size_l;
+  max-width: @size_l;
+  padding: @size_ml @size_s @size_ml @size_l;
+  .padding--left--l();
+  filter: brightness(0) invert(1); /* makes the file icon white without needing a different <img> src attr */
```
clcaelynn-clever, line 43: Fancy! Works on all browsers we want to support?

src/AttachmentPreview/AttachmentPreview.less
```diff
@@ -0,0 +1,195 @@
...
+}
+
+.AttachmentPreview--PreviewWindow img {
+  .zIndex--8;
+  // render image as larger as natural size, or scale down to fit screen
+  object-fit: scale-down;
+  height: 100%;
+  width: 100%;
+
+  // TODO: fix this
```
clcaelynn-clever, line 72: I know this is annoying, but any chance you could make a JIRA ticket for this and then perhaps link the ticket # (e.g. `M5G-123`) inline with this TODO?

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,111 @@
...
+
+import "./AttachmentPreview.less";
+
+export interface Props {
+  attachmentID: string;
+  attachmentName: string;
+  attachmentURL: string;
+  className?: string;
+  closeButtonAriaLabel?: string;
+  downloadButtonText?: string;
```
clcaelynn-clever, line 17: As mentioned elsewhere, either this needs to be a required prop or we need an optional aria-label prop for when this one is absent!

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,111 @@
...
+  closeButtonAriaLabel,
+  downloadButtonText,
+  fileType,
+  onClickDownload,
+  onClose,
+}) => {
+  const ESC = 27;
+
+  function handleKeyUp(e) {
+    console.log("\ne:", e);
```
clcaelynn-clever, line 55: stray log

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,111 @@
...
+        <FlexBox grow className={cssClass.ATTACHMENT_NAME}>
+          {attachmentName}
+        </FlexBox>
+        <Button
+          type="linkPlain"
+          className={cssClass.DOWNLOAD_CONTAINER}
+          onClick={() => onClickDownload(attachmentID)}
+        >
+          <FontAwesome className={cssClass.DOWNLOAD_BUTTON} name="download" />{" "}
+          <span>{downloadButtonText || "Download"}</span>
```
clcaelynn-clever, line 82: Oh, I see this now. Could "Download" be the default value of the prop, instead of having it here?

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,111 @@
...
+      onClose();
+    }
+  }
+
+  React.useEffect(() => {
+    window.addEventListener("keyup", handleKeyUp);
+    return () => {
+      window.removeEventListener("keyup", handleKeyUp);
+    };
+  }, []);
```
clcaelynn-clever, line 66: For my understanding, can you explain what's going on here with the add and then the remove?

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,111 @@
...
+      </FlexBox>
+      <FlexBox className={cssClass.FOOTER_BAR}>
+        <Button
+          type="linkPlain"
+          grow
+          className={cssClass.DOWNLOAD_CONTAINER}
+          onClick={() => onClickDownload(attachmentID)}
+        >
+          <FontAwesome className={cssClass.DOWNLOAD_BUTTON} name="download" />{" "}
+          <span>{downloadButtonText || "Save"}</span>
```
clcaelynn-clever, line 106: Ah, now I see why there's not one default value!

I'd suggest splitting `downloadButtonText` into two props then, one for the 'desktop' size and one for the 'mobile' size, like `downloadButtonTextDesktop` and `downloadButtonTextMobile`. And then they can both have their appropriate default values. 🙂

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -12,27 +13,41 @@ function cssClass(element: string) {
 
 type Props = {
   attachmentID: string;
+  // TODO: should this be optional?
```
clcaelynn-clever, line 16: Actually, why do we need `attachmentURL` at all? The `onClickAttachment` is what we use to make opening it happen. 🤔

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -12,27 +13,41 @@ function cssClass(element: string) {
...
   onClickAttachment: (attachmentID: string) => void;
   onRemoveAttachment?: (attachmentID: string) => void;
-  title?: string;
   subtitle?: string;
-  isUpload?: boolean;
+  title?: string;
   uploadComplete?: boolean;
 };
 
+// TODO: should I use a discriminated union here? to keep the props neat?
```
clcaelynn-clever, line 31: Seems not strictly necessary IMO, considering how much work you've already done here. That said, it might actually help with keeping things neat going forward and prevent breakage!

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -12,27 +13,41 @@ function cssClass(element: string) {
 
 type Props = {
   attachmentID: string;
+  // TODO: should this be optional?
+  attachmentURL: string;
   errorMsg?: string;
-  icon: React.ReactNode;
+  // TODO: is this needed, or only mimeType?
```
clcaelynn-clever, line 19: Seems like only one is required, so whichever is more convenient? We should always have both available from the consumer side.

clcaelynn-clever (changes requested):
Looks great! A few small notes, but wow, great job putting all this together. 🥳  🙌🏻  🎉 🌮

docs/components/AttachmentPreviewView.jsx
jonah-clever: So it does have a default value of "Download". Would that make it so the aria-label is not needed?

src/AttachmentPreview/AttachmentPreview.tsx
jonah-clever: Yes, I can also add a comment explaining this. In React functional components, componentDidMount and componentWillUnmount aren't used. The way to achieve this behavior is with useEffect. If there is nothing passed into the second prop of useEffect ([]), then it will only get called when the component mounts. And the return value for the function gets called upon unmount. So essentially, this is the react hooks version of 

```
componentDidMount() {
    window.addEventListener("keyup", this.handleKeyUp);
  }

  componentWillUnmount() {
    window.removeEventListener("keyup", this.handleKeyUp);
  }
```

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: We need it to render the image, unless there's another way to do that

docs/components/AttachmentPreviewView.jsx
jonah-clever: What if I upload it to s3, in assets.clever.com perhaps?

src/AttachmentPreview/AttachmentPreview.less
jonah-clever: Yes, except for IE

jonah-clever: This is complements of @cofauver ; j

src/AttachmentPreview/AttachmentPreview.less
cofauver: I think I ended up needing to adjust a parent container (not the element with the misbehaving text, but it's parent) to change the `overflow` property? Unfortunately, this was one of so many changes that I don't remember the specifics.

nbhatia823: Also have a similar memory that the parent container needs to have a new property, but thinking it is that the `width: 100%` is what might do it.  So maybe try it in `AttachmentName` and/or `HeaderBar`. Hopefully either this or `overflow` as Cory mentioned will work!

docs/components/AttachmentPreviewView.jsx
clcaelynn-clever: I think that would be ideal, and in a way pretty straightforward! Not a blocker though, IMO.

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: Would it be unreasonable to use the mime package to get the mimetype from the filetype? if its reasonable, then I can remove the mimetype as a prop

docs/components/AttachmentPreviewView.jsx
clcaelynn-clever: I addressed this in further comments later on, as I imagine you saw or will see. Didn't realize that when I wrote this!

src/AttachmentPreview/AttachmentPreview.tsx
clcaelynn-clever: Ah nice, makes sense! Very smooth.

src/MessagingAttachment/MessagingAttachment.tsx
clcaelynn-clever: I think I'm forgetting the context for this, but why do we need `attachmentID` for that in the first place? No harm in needing it - we always have one generated consumer-side now - just wondering.

src/MessagingAttachment/MessagingAttachment.tsx
clcaelynn-clever: I think making the consumer pass it in is better, so that the component doesn't have to do any conversions itself. No harm in keeping it as a prop!

src/MessagingAttachment/MessagingAttachment.tsx
clcaelynn-clever: Don't let this question block you IMO!

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: Actually, its a little hacky, but I can use `fileTypeToIcon()` ([link](https://github.com/Clever/components/blob/master/src/MessagingAttachment/MessagingAttachment.tsx#L99)) to achieve this

src/AttachmentPreview/AttachmentPreview.less
jonah-clever: I'll just leave it here for now

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: Oh no, we need URL for rendering the image. The id is used throughout MessagingAttachment.tsx

docs/components/MessagingAttachmentView.jsx
```diff
@@ -170,6 +168,7 @@ export default class MessagingAttachmentView extends React.PureComponent {
   _renderProps() {
     return (
       <>
+        {/* TODO: add new props */}
```
nbhatia823, line 171: Is this still relevant?

src/AttachmentPreview/AttachmentPreview.less
```diff
@@ -0,0 +1,193 @@
+@import (reference) "../less/utilities";
+
+@HEADER_BAR_HEIGHT: @size_5xl;
+@DESKTOP_BUFFER_SIZE: @size_l;
+@MOBILE_BUFFER_SIZE: @size_xs;
+
+// TODO: can I add this to sizing.less?
```
nbhatia823, line 7: Probably should remove the comment :D

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,113 @@
...
+  useEffect(() => {
+    window.addEventListener("keyup", handleKeyUp);
+    return () => {
+      window.removeEventListener("keyup", handleKeyUp);
+    };
+  }, []);
+
+  return (
+    <div className={classnames(cssClass.CONTAINER, className)}>
+      <div className={cssClass.BACKGROUND} aria-hidden="true" />
```
nbhatia823, line 73: What does aria-hidden do?

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,113 @@
...
+          type="linkPlain"
+          className={cssClass.DOWNLOAD_CONTAINER}
+          onClick={() => onClickDownload(attachmentID)}
+        >
+          <FontAwesome className={cssClass.DOWNLOAD_BUTTON} name="download" />{" "}
+          <span>{downloadButtonTextDesktop}</span>
+        </Button>
+        <Button
+          type="linkPlain"
+          aria-label={closeButtonAriaLabel || "close attachment preview"}
```
nbhatia823, line 89: Maybe you should move the "close attachment preview" to also be the default value for the prop to match the pattern of the others?

src/AttachmentPreview/AttachmentPreview.tsx
```diff
@@ -0,0 +1,113 @@
...
+        </FlexBox>
+      </FlexBox>
+      <FlexBox className={cssClass.FOOTER_BAR}>
+        <Button
+          type="linkPlain"
+          className={cssClass.DOWNLOAD_CONTAINER}
+          onClick={() => onClickDownload(attachmentID)}
+        >
+          <FontAwesome className={cssClass.DOWNLOAD_BUTTON} name="download" />{" "}
+          <span>{downloadButtonTextMobile}</span>
```
nbhatia823, line 108: I imagine the span is unnecessary here, but could be wrong!

src/MessagingAttachment/MessagingAttachment.tsx
nbhatia823: It seems like `attachmentID` could be removed if the parameters `onClickAttachment` and `onRemoveAttachment` didn't took in a function that didnt require the `attachmentID` as a parameter. That seems very possible since the consumer is passing in those functions, so the consumer could just pass the `attachmentID` to those functions so then those functions would each be of type `() => void`

nbhatia823: Also can the comment about the URL potentially being optional be removed?

src/AttachmentPreview/AttachmentPreview.tsx
jonah-clever: makes it so screen readers don't interact with this element

src/MessagingAttachment/MessagingAttachment.tsx
```diff
@@ -33,6 +40,9 @@ export const MessagingAttachment: React.FC<Props> = ({
   isUpload,
   uploadComplete,
 }: Props) => {
+  const [attachmentPreviewIsShowing, setAttachmentPreviewIsShowing] = React.useState(false);
+  const isImageAttachment = fileTypeToIcon(fileType) === "image";
```
nbhatia823, line 44: How come this is using a function called `fileTypeToIcon` but it returns a string instead of an icon?

nbhatia823 (approved):
This looks awesome! Left a few comments, mostly just comments that can be removed but seems good to go from my perspective :excited:

src/AttachmentPreview/AttachmentPreview.tsx
nbhatia823: ooo very nice!

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: yeah that name is misleading. it returns the type of icon to use, the name of the icon if you will. i should change the name of the function to `fileTypeToIconName` or something?

nbhatia823: Yeah maybe `fileTypeToIconType` as it seems `image` would be a `type`? Im not too sure what the other response types are though so whatever you think is best is good and seems like an improvement over just `Icon`

jonah-clever: that's what i ended up doing 👍

src/MessagingAttachment/MessagingAttachment.tsx
jonah-clever: I think I'm gonna leave both for now

nbhatia823: Makes sense, especially since changing the functions of the existing props would require updating the existing use cases of it in fampo/launchpad

PR #686, opened 2021-07-13
M5G: Dynamic sizing of MessagingAttachments (part 1)

jonah-clever:
# Jira: [M5G-546](https://clever.atlassian.net/browse/M5G-546)

# Overview:

Longer MessagingAttachments were looking pretty borked on mobile...

Screen Shot 2021-07-12 at 5 11 02 PM

Width was being set by a desktop selector, so I added width to the mobile selectors. Also made sure attachment-only-messages don't take the margin into account (`calc...`) because they already have space for the timestamp

**NOTE: this does not make the MessagingAttachment grow/shrink with the message text content. I couldn't figure that out within a reasonable amount of time, so because its not a launch blocker, I'm gonna leave it for another time. What this PR does is make MessagingAttachments 1) align correctly at the edges (not get cut off), and render with a width of 230px for attachments with short titles, between 230-360px for medium titles (based on the length of the title), and cutoff at 360px for long titles, with truncated text and ellipsis.**

# Screenshots/GIFs:

Correct max/min widths
Screen Shot 2021-07-12 at 5 03 28 PM
Screen Shot 2021-07-12 at 5 03 38 PM

Also same for teachers
Screen Shot 2021-07-12 at 5 07 49 PM
Screen Shot 2021-07-12 at 5 07 55 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] Edge

# Roll Out:

- Before merging:
  - [ ] Updated docs
- After merging:

### DIFF
--- a/src/MessagingAttachment/MessagingAttachment.less
+++ b/src/MessagingAttachment/MessagingAttachment.less
@@ -215,10 +215,14 @@
 
   .MessagingBubble--Message--Container--Own .MessagingAttachment--Container {
     margin-left: @size_5xl;
+    width: calc(100% - 4rem);
+    overflow: hidden;
   }
 
   .MessagingBubble--Message--Container--Other .MessagingAttachment--Container {
     margin-right: @size_5xl;
+    width: calc(100% - 4rem);
+    overflow: hidden;
   }
 
   // ⬇️ In attachment-only messages, we don't want the margins defined above either ⬇️
@@ -227,6 +231,7 @@
     .MessagingAttachment--Container {
     margin-left: 0;
     margin-right: 0;
+    width: 100%;
   }
 
   .MessagingBubble--Message--Timestamp--Other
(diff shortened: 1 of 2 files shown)


src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -215,10 +215,18 @@
 
   .MessagingBubble--Message--Container--Own .MessagingAttachment--Container {
     margin-left: @size_5xl;
+    width: calc(100% - 4rem);
+    min-width: 13rem;
+    max-width: 21.125rem;
+    overflow: hidden;
```
nbhatia823, line 221: Is it that these four lines only apply to mobile messaging attachments within messaging bubles and that's why these are not in the `MessagingAttachment--Container` class on line 207?

nbhatia823 (approved):
These css classes are getting pretty confusing 😅  i think it all seems good!

src/MessagingAttachment/MessagingAttachment.less
jonah-clever: Ah I think you're right

PR #691, opened 2021-07-22
Restyle MessagingInput in prep for FamPo mobile app

arsalansufi:
***Commit-by-commit reviewing strongly recommended!***

# Jira

[M5G-550](https://clever.atlassian.net/browse/M5G-550)

# Overview

This PR restyles the `MessagingInput` in prep for the FamPo mobile app. The Design team has done some awesome work to ensure that the new designs work not only in the FamPo mobile app but also in all other contexts (student, teacher, guardian, desktop, etc.)! You can see the [Figma mocks](https://www.figma.com/file/WMnR0eBZ0LKoON6GimMkl0/Clever-Messages-v2-Announcements?node-id=3720%3A33032) here.

## Before

before-1
before-2
before-3
before-4

## After

after-1
after-2
after-3
after-4
after-5
after-6
after-8
after-9
after-10
after-22
after-18
after-11
after-12
after-13
after-14
after-15
after-16
after-17
after-19
after-20
after-23

# Testing

See all of the above screenshots! In addition to testing the large matrix of possible states, I've also considered accessibility.

I'm still doing some final stress testing but wanted to get this PR out proactively!

_Browser testing_

- [x] Chrome
- [x] Safari
- [ ] IE11 (In progress)

# Rollout

_Before merging_

- [x] Updated docs

_After merging_


src/MessagingInput/MessagingInput.tsx
```diff
@@ -54,13 +55,14 @@ const MessagingInputRenderFunction: React.ForwardRefRenderFunction<MessagingInpu
...
     disableSendButton,
-    labelText = "Send a message",
+    label,
     newlineOnEnter,
     onBlur,
     onChange,
     onFocus,
     onReplyCancel,
     onSubmit,
+    placeholder = "Send a message",
```
arsalansufi, line 65: A placeholder inside the input is now the default, rather than a label outside.

_Before_

before-1

_After_

after-1

-----

That said, we still need to support labels outside for a few edge cases, e.g. this input in the upcoming welcome flow.

welcome

src/TextArea/TextArea.tsx
```diff
@@ -225,7 +225,7 @@ export class TextArea extends React.Component<Props, State> {
 
     const textAreaProps = {
       ["aria-invalid"]: !valid,
-      ["aria-label"]: this.props.label,
+      ["aria-label"]: this.props.label || this.props.placeholder,
```
arsalansufi, line 228: For accessibility! Since we may no longer pass a `label` through to the underlying `TextArea` component, let's fallback to using the `placeholder` as the `aria-label`, which will now always be specified

docs/components/MessagingInputView.jsx
```diff
@@ -216,6 +219,14 @@ export default class MessagingInputView extends React.PureComponent {
...
             Show checkbox
           </label>
+          <label className={cssClass.CONFIG}>
+            <input
+              type="checkbox"
+              checked={showLabel}
+              onChange={({ target }) => this.setState({ showLabel: target.checked })}
+            />{" "}
+            Show label
+          </label>
```
arsalansufi, line 229: Updated Dewey docs!

docs

src/MessagingInput/MessagingInput.less
```diff
@@ -195,29 +196,6 @@
...
-    .FileInput {
-      position: absolute;
-      right: 2.25rem;
-      bottom: 0;
-
-      .FileInput--Icon {
-        .margin--left--xs();
-        margin-right: 0.3125rem; // 5px
-      }
-    }
```
arsalansufi, line 233: These styles were all previously necessary to account for the space taken up by the in-input send button on mobile. With mobile now more closely mirroring desktop, fewer overrides are necessary for mobile! :tada:

_Before_

before-3

_After_

after-12

src/MessagingInput/MessagingInput.tsx
```diff
@@ -150,8 +157,14 @@ const MessagingInputRenderFunction: React.ForwardRefRenderFunction<MessagingInpu
...
-            onBlur={onBlur}
-            onFocus={onFocus}
+            onBlur={() => {
+              setIsInputActive(false);
+              onBlur();
+            }}
+            onFocus={() => {
+              setIsInputActive(true);
+              onFocus();
+            }}
```
arsalansufi, line 167: We use this state to tweak the styles of the reply area when the input area is focused. There unfortunately wasn't a good way to use CSS alone to target the reply area when the input area is focused. The sibling `+` selector almost does the trick but only works for capturing the second element of two adjacent elements, not the first, which is what we need here (since the reply area comes before the input area in the DOM).

src/MessagingInput/MessagingInput.less
```diff
@@ -204,14 +214,24 @@
...
+  }
+
   .MessagingInput--TextField {
     .margin--right--none();
     min-height: 2rem;
     border-radius: 2rem;
+
+    &.TextArea--inFocus {
+      .border--m(@neutral_silver);
+    }
```
arsalansufi, line 236: I'm having some discussions with Nick and Jemma around accessibility requirements for focus states. So I may end up making some slight tweaks here.

For example, in the mocks, we change the border color on focus on desktop, but not on mobile. We may go back and also change the border color on focus on mobile.

We may also use a darker border color for the focus state, to ensure sufficient contrast.

cofauver (approved):
What awesome documentation of your work in this PR! Thanks for making it easy to review!

My only question is around this screenshot:

image


Rounded edges + scroll is complicated and I'm wondering if there's any less awkward-looking option. No worries if this is where we land!

arsalansufi:
> Rounded edges + scroll is complicated and I'm wondering if there's any less awkward-looking option. No worries if this is where we land!

A great question! I'm running this by Nick and Jemma.

arsalansufi:
I'm going to proceed as is for now but have created a `Backlog - UX` Jira ticket to track this quirk! https://clever.atlassian.net/browse/M5G-611

prime-time:
Wow, sooo many screenshots 📸 🖼️ ✨

arsalansufi:
FYI @cofauver I address the rounded edges + scroll edge case in https://github.com/Clever/components/pull/697!

PR #697, opened 2021-07-26
Reduce mobile input rounding when text spans multiple lines

arsalansufi:
# Jira

[M5G-550](https://clever.atlassian.net/browse/M5G-550) / [M5G-611](https://clever.atlassian.net/browse/M5G-611)

# Overview

We recently restyled the Messaging input (https://github.com/Clever/components/pull/691)! This PR addresses [an edge case surfaced in that original PR](https://github.com/Clever/components/pull/691#pullrequestreview-713934430).

## Before

The new rounded mobile input can look a bit funky when there are multiple lines of text:

before

## After

After discussion with Nick, we've decided to reduce the rounding when there are multiple lines of text, just like we do when there's a reply or attachments:

one-line

multi-line

reduced-rounding

# Testing

_Browser testing_

- [x] Chrome
- [x] Safari
- [ ] IE11

_Other testing_

- [x] Verified that the height detection mechanism works when the screen is zoomed in or out

# Rollout

_Before merging_

- [ ] ~Updated docs~ N/A

_After merging_

- [ ] ~Posted in #eng if I made a breaking change to a beta component~ N/A

### DIFF
--- a/src/MessagingInput/MessagingInput.less
+++ b/src/MessagingInput/MessagingInput.less
@@ -227,11 +227,6 @@
     min-height: 2rem;
     border-radius: 2rem;
 
-    .TextArea--input {
-      // Ensures that multi-line text doesn't spill outside the boundaries of the rounded text input
-      border-radius: 0.8rem;
-    }
-
     &.TextArea--inFocus {
       .border--m(@neutral_silver);
     }
@@ -240,18 +235,14 @@
   .MessagingInput--TextField.MessagingInput--TextField--WithReply {
     .borderRadius--bottom--xl();
     border-top: none;
-
-    .TextArea--input {
-      .borderRadius--0();
-    }
   }
 
   .MessagingInput--TextField.MessagingInput--TextField--WithAttachments {
     .borderRadius--xl();
+  }
 
-    .TextArea--input {
-      .borderRadius--0();
-    }
+  .MessagingInput--TextField.MessagingInput--TextField--MultiLine {
+    .borderRadius--xl();
   }
 
   .MessagingInput--Container {
--- a/src/MessagingInput/MessagingInput.tsx
+++ b/src/MessagingInput/MessagingInput.tsx
@@ -83,6 +83,9 @@ const MessagingInputRenderFunction: React.ForwardRefRenderFunction<MessagingInpu
   const attachmentsArePresent = attachments?.length > 0;
   const [isInputActive, setIsInputActive] = useState(false);
 
+  // Update this if relevant styling changes and this heuristic no longer holds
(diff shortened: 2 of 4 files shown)


src/MessagingInput/MessagingInput.less
```diff
@@ -227,11 +227,6 @@
     min-height: 2rem;
     border-radius: 2rem;
 
-    .TextArea--input {
-      // Ensures that multi-line text doesn't spill outside the boundaries of the rounded text input
-      border-radius: 0.8rem;
```
arsalansufi, line 232: This is no longer necessary since the text was only spilling outside the boundaries of the input when there were multiple lines of text and the input was still fully rounded.

src/MessagingInput/MessagingInput.tsx
```diff
@@ -83,6 +83,9 @@ const MessagingInputRenderFunction: React.ForwardRefRenderFunction<MessagingInpu
   const attachmentsArePresent = attachments?.length > 0;
   const [isInputActive, setIsInputActive] = useState(false);
 
+  // Update this if relevant styling changes and this heuristic no longer holds
+  const hasMultiLineText = textAreaRef.current?.textAreaHeight() > 40;
```
arsalansufi, line 87: I know this heuristic isn't perfect, but it was the best I could come up with for a quick, low-lift fix!

src/TextArea/TextArea.tsx
```diff
@@ -155,6 +155,11 @@ export class TextArea extends React.Component<Props, State> {
     this.textAreaEl.current.focus();
   }
 
+  // Give consumers that set up a React ref access to the <textarea> height
+  textAreaHeight(): number | undefined {
```
jonjonray, line 159: big fan of this pattern!

src/MessagingInput/MessagingInput.tsx
jonjonray: Random thought here: what if you checked if the element `scrollHeight`  is larger than the `clientHeight`? Should hold up even if the input styling changes.

That being said, this looks okay to me as well!

jonjonray (approved):
LGTM! Left an idea on how to compute the multiline boolean but feel free to ignore

src/MessagingInput/MessagingInput.tsx
arsalansufi: Ooh great idea! The one catch is that we also want to decrease the rounding when there are only two lines of text. In this case, there isn't a scrollbar present, so the `scrollHeight > clientHeight` heuristic won't catch that case

PR #705, opened 2021-07-29
[a11y] Improve contrast ratio of success toast text 👍 🟩 🍞

prime-time:
# Jira:
https://clever.atlassian.net/browse/PRTL-735

# Overview:
This change increases the contrast ratio of success toasts to make them more accessible.

# Screenshots/GIFs:
Before
Screen Shot 2021-08-19 at 9 38 28 PM

After
Screen Shot 2021-08-19 at 9 37 55 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.156.3",
+  "version": "2.156.4",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/ToastStack/ToastNotification.less
+++ b/src/ToastStack/ToastNotification.less
@@ -1,5 +1,9 @@
 @import (reference) "../less/utilities";
 
+@success_green: #1d835f;
+@success_button_green: #0a6447;
+@success_button_hover_green: #06563c;
+
 .ToastNotification {
   .borderRadius--m;
   .boxShadow--light;
@@ -26,7 +30,7 @@
 }
 
 .ToastNotification--success {
-  background-color: @alert_green;
+  background-color: @success_green;
 }
 
 .ToastNotification--warning {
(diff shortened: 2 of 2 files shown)


genColors.ts
```diff
@@ -35,6 +35,8 @@ const Colors = {
   // Alert colors:
   Alert: {
     ALERT_GREEN: "#1DA978",
```
adamvictorclever, line 37: Is this a situation where we would eventually want to remove the old color? I guess I'm wondering if we should just replace the value for `ALERT_GREEN` and `ALERT_GREEN_SHADE_1/2`, although admittedly that could be a bigger change.

adamvictorclever (approved):
LGTM!

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.156.3",
+  "version": "2.156.4",
```
prime-time, line 3: Actually, adding a new color variable should make this a minor version I think

genColors.ts
prime-time: Great point. I believe so, but let me look into that some more.

prime-time: I wonder if it would be better to skip introducing new color variables here, and just define them locally in the toast less file. Then we worry about adding new color variables when we have more time to devote to overhauling dewey.

prime-time: I think I'll define them locally for now.

prime-time:
@adamvictorclever I decided to punt on introducing new color variables.

adamvictorclever (approved):
I like the idea of just defining local colors until we think through the larger Dewey change. Butter it!

prime-time:
🧈

PR #718, opened 2021-08-11
added analytics prop onReadReceiptsHover

brianbyrhee:
# Jira: [M5G-690](https://clever.atlassian.net/browse/M5G-690)

# Overview:

With the release of M1, we wanted to implement a Segment analytics counter that keeps track of how often a teacher hovers over the tooltip to check read receipts. More info on analytics for RR can be found [here](https://docs.google.com/document/d/1yM_CsSdI0IfUz78INVpYySItH-xhdzOqw34p67IHCSQ/edit?usp=sharing). To do this, we pass in a new `onReadReceiptsHover` prop to the `AnnouncementBubble` component as a function type in order to pass in analytics function from LP. We do this because we cannot import segment in the components repo, so we pass it in through a prop to AnnouncementBubble, as we can only access the hover state within the component itself. We will implement more analytics with the new features that we add in when rolling out M2 in the future. 

With the opportunity of this PR, we also fix some css styling for RR with one element, as we wanted to make the RR more left-aligned, but the padding was too thin upon rollout. Look below for screenshots of new changes.

# Screenshots/GIFs:

Screen Shot 2021-08-10 at 7 51 03 PM
Screen Shot 2021-08-10 at 7 50 49 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.less
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.less
@@ -163,7 +163,8 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
   .text--semi-bold();
-  padding-left: @size_s;
+  padding-right: @size_xs;
+  padding-left: @size_2xs;
 }
 
 .NormalAnnouncementBubble--readReceipts--icon {
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.less
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.less
@@ -163,7 +163,7 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
   .text--semi-bold();
-  padding-right: @size_xs;
+  padding-right: @size_2xs;
   padding-left: @size_2xs;
 }
 
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
@@ -146,23 +146,23 @@ function formReadReceiptsTooltip(
   const displayRecipientType = recipientType === "guardian" ? "parent" : recipientType;
   const recipientString = readBy.length === 1 ? displayRecipientType : `${displayRecipientType}s`;
   return (
-    <FlexBox className={cssClass("readReceipts--container")} alignItems="center" justify="end">
-      <div onMouseOver={onReadReceiptsHover}>
-        <Tooltip
-          className={cssClass("readReceipts--tooltip")}
-          content={readReceiptString}
-          placement={"top"}
-          textAlign={"left"}
-        >
-          <FlexBox>
+    <FlexBox alignItems="center" justify="end">
+      <Tooltip
+        className={cssClass("readReceipts--tooltip")}
+        content={readReceiptString}
+        placement={"top"}
+        textAlign={"left"}
+      >
+        <div onMouseEnter={onReadReceiptsHover}>
+          <FlexBox className={cssClass("readReceipts--container")}>
             <Checkmark className={cssClass("readReceipts--icon")} />
             <span className={cssClass("readReceipts--text--desktop")}>
               Read by {readReceiptCount} {recipientString}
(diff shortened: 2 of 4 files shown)


src/AnnouncementBubble/NormalAnnouncementBubble.less
```diff
@@ -163,7 +163,8 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
   .text--semi-bold();
-  padding-left: @size_s;
+  padding-right: @size_xs;
+  padding-left: @size_2xs;
```
nbhatia823, line 167: What was the thinking behind having more padding on the right than the left?

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
```diff
@@ -144,20 +147,22 @@ function formReadReceiptsTooltip(
...
-      >
-        <FlexBox>
-          <Checkmark className={cssClass("readReceipts--icon")} />
-          <span className={cssClass("readReceipts--text--desktop")}>
-            Read by {readReceiptCount} {recipientString}
-          </span>
-          <span className={cssClass("readReceipts--text--mobile")}>{readReceiptCount}</span>
-        </FlexBox>
-      </Tooltip>
+      <div onMouseOver={onReadReceiptsHover}>
```
nbhatia823, line 150: I had thought it would make more sense to put the `<div` directly around the the stuff inside the Tooltip instead of around the Tooltip as well. that way the idea if they hover over the trigger (the child of the tooltip) then we would record a log

nbhatia823 (commented):
You may also want to test that the hover prop works as intended by adding a `console.log` as the prop in the AnnouncementBubbleView

src/AnnouncementBubble/NormalAnnouncementBubble.less
brianbyrhee: I think the tooltip looked good for all readBy receipts except for when the array had 1 name. You can currently see how this looks on live, but the name seems too left-aligned so I added some padding on the right as well. There's a bit more padding on the right to make it seem more right-aligned. You can also observe this issue on live LP as well

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
brianbyrhee: Makes sense, I'll change this rn

brianbyrhee: Just wondering, how would this interact if the hover function was wrapped around the tooltip? How would this behave differently relative to the tooltip?

nbhatia823: To be honest im not certain... i would have to test it out to check. I worried that it would double count if someone hovered over the trigger and then again over the tooltip itself

src/AnnouncementBubble/NormalAnnouncementBubble.less
nbhatia823: Okay sounds good, just want to double check that it looks better as is than if there was the same padding on both sides

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
```diff
@@ -146,23 +146,23 @@ function formReadReceiptsTooltip(
...
-          <FlexBox>
+    <FlexBox alignItems="center" justify="end">
+      <Tooltip
+        className={cssClass("readReceipts--tooltip")}
+        content={readReceiptString}
+        placement={"top"}
+        textAlign={"left"}
+      >
+        <div onMouseEnter={onReadReceiptsHover}>
+          <FlexBox className={cssClass("readReceipts--container")}>
```
brianbyrhee, line 157: I moved the container tag to here bc there was a bug where the tooltip would turn blue upon hover across the entire width of the AnnouncementBubble component rather than just the tooltip itself.

nbhatia823 (approved):
Looks good to me!

PR #719, opened 2021-08-12
M5G-616: Family Portal Bubbles refresh

jonah-clever:
# Jira: [M5G-616](https://clever.atlassian.net/browse/M5G-616)

# Overview:

Styling changes for FamPo messaging bubbles.

Changes include:
* Replaced existing `theme` prop (own vs other message) with `bubbleType`. Added new `theme` prop for `launchpad` vs `familyPortal` styling. Currently, launchpad is the default styling, and when the theme is set to `familyPortal`, the new styles are applied
* MessagingBubbles
  * Corner rounding 
  * Colors
  * Font sizes
* AttachmentBubbles
  * Corner rounding
  * Decreased padding
  * drop shadow
  * Spacing between multiple attachments

Also:
* I added real quoted attachments/messages to the examples in Docs, with SegmentedControls for `bubbleType` and `theme`, and updated the prop documentation
* **NOTE: the commit history is a bit of a mess, it's kinda a toss up whether its better to look at the commits or just the files. Up to you, sorry about that.**

# Screenshots/GIFs:

Screen Shot 2021-08-12 at 2 19 37 PM
Screen Shot 2021-08-12 at 2 19 42 PM
Screen Shot 2021-08-12 at 2 19 48 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
     
- After merging:


src/less/colors.less
```diff
@@ -6,59 +6,62 @@
...
-@accent_purple_shade_2: #5201bf;
-@accent_purple_tint_1: #9244fa;
-@accent_purple_tint_2: #a460ff;
+@accent_purple_shade_1: #6609E0;
+@accent_purple_shade_2: #5201BF;
+@accent_purple_tint_1: #9244FA;
+@accent_purple_tint_2: #A460FF;
+
+// FamilyPortalMobile colors:
+@family_portal_mobile_slate: #32315B;
```
jonah-clever: this is the only real addition

src/less/colors.less
```diff
@@ -6,62 +6,63 @@
...
+@accent_purple_shade_1: #6609e0;
+@accent_purple_shade_2: #5201bf;
+@accent_purple_tint_1: #9244fa;
+@accent_purple_tint_2: #a460ff;
 
-// FamilyPortalMobile colors:
-@family_portal_mobile_slate: #32315B;
+// FamilyPortal colors:
+@family_portal_slate: #32315b;
+@family_portal_quoted_announcement: #fef7d2;
```
jonah-clever: again, this is the only addition

src/AnnouncementBubble/DeletedAnnouncementBubble.tsx
```diff
@@ -10,7 +10,8 @@ function cssClass(element: string) {
 }
 
 export interface Props {
-  theme: "deleted";
+  bubbleType: "deleted";
+  theme?: "launchpad" | "familyPortal";
```
arsalansufi, line 14: Quick thought! Rather than calling the theme `"launchpad"` (since that's more of an internal name than a product name), maybe we can just call it `"default"`. (I also considered `"studentTeacherPortal"` for a sec, but there's a world where we differentiate the student and teacher Portals, so that's not the most future proof)

arsalansufi: Also, it might make sense to define this theme in a shared type in [the src/utils/ directory](https://github.com/Clever/components/tree/master/src/utils), e.g.

```ts
type MessagingTheme = "default" | "familyPortal";
```

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -11,6 +11,63 @@
   .text--truncate();
 }
 
+// Family portal specific styling
+.MessagingAttachment--ParentContainer.MessagingAttachment--FamilyPortal {
+  border-radius: @size_m;
+  background-color: @neutral_white;
+  box-shadow: 0px @size_3xs 6px rgba(71, 76, 94, 0.3);
```
arsalansufi, line 18: Just curious! Could we use one of our [box shadow mixines](https://github.com/Clever/components/blob/master/src/less/box-shadow.less) here?

```
.boxShadow--medium();
```

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -11,6 +11,63 @@
...
+    .padding--y--s();
+    .padding--x--xs();
+  }
+
+  .MessagingAttachment--IconContainer {
+    margin-right: 0.5rem;
+  }
+
+  .MessagingAttachment--Title {
+    font-size: 0.875rem;
```
arsalansufi, line 34: There's now a mixin for this font size! `.text--smallMedium()`

arsalansufi:
These bubbles are looking so good!! Just left a light sprinkling of comments since I had some context as we've been discussing this task together

src/AnnouncementBubble/DeletedAnnouncementBubble.tsx
jonah-clever: Would this be in its own file, or in the types.ts file (for example)

arsalansufi: I think it could probably live in its own file! Maybe `messaging.ts`, so that all our team's constants are grouped together for an eventual move to a standalone Messaging components library

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -11,6 +11,63 @@
...
+
+// Attachment-only messages
+.MessagingBubble--HideBubble
+  .MessagingAttachment--ParentContainer.MessagingAttachment--FamilyPortal:first-of-type {
+  margin-top: 0;
+}
+
+// Replying to an announcement that countains an attachment
+.QuotedAnnouncementBubble--container {
+  .MessagingAttachment--ParentContainer.MessagingAttachment--FamilyPortal:first-of-type {
```
arsalansufi, line 47: I see that this first classname in this group includes the theme, but all subsequent styles in this file aren't gated by theme. Is that intentional? (Just sanity checking)

jonah-clever: hmm which theme? `.MessagingAttachment--FamilyPortal`?

arsalansufi: Yup exactly! So the `margin-top: 0` on the next line will only get applied when the theme is `familyPortal` but all styles that come after will get applied everywhere, regardless of theme. Which again may be intentional! Just sanity checking that we're intentionally changing styles outside of FamPo. As we discussed in person, I'm all for applying style changes globally where possible, so this may be okay.

jonah-clever: If so, then yes. The theme class gets applied to the top level container of the component, then subsequent classes are updated based on that

jonah-clever: Unless I'm confused, which I maybe am

arsalansufi: I'm almost wondering the opposite, too. Should the `margin-top: 0` on the next line also get applied regardless of theme (removing the `.MessagingAttachment--FamilyPortal` above)?

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -11,6 +11,64 @@
...
+  .MessagingAttachment--Container {
+    border: unset;
+    border-radius: unset;
+
+    .padding--y--s();
+    .padding--x--xs();
+  }
+
+  .MessagingAttachment--IconContainer {
+    margin-right: 0.5rem;
```
jonjonray, line 31: nit: I believe we have an equivalent mix-in for this

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -11,6 +11,64 @@
...
+
+  .MessagingAttachment--Title {
+    .text--smallMedium();
+  }
+}
+
+// Attachment-only messages
+.MessagingBubble--HideBubble
+  .MessagingAttachment--ParentContainer.MessagingAttachment--FamilyPortal:first-of-type {
+  margin-top: 0;
```
jonjonray, line 42: nit: same as above, should use `.margin--top--none();`

src/MessagingBubble/MessagingBubble.less
```diff
@@ -22,6 +22,42 @@
...
+    .padding--y--xs();
+  }
+
+  .MessagingBubble--Message--Timestamp--Own {
+    margin-right: @size_m;
+  }
+
+  .MessagingBubble--Message--Reply--Own {
+    // TODO: is this the best way to override a background-color?
+    background-color: unset;
```
jonjonray, line 41: this makes sense to me since it's inheriting the background color from `.MessagingBubble--Message--Own`!

src/MessagingBubble/MessagingBubble.less
```diff
@@ -22,6 +22,42 @@
...
+    margin-left: -0.5rem;
+    margin-right: -0.5rem;
+    &.MessagingBubble--Message--Reply--MarginBottom {
+      margin-bottom: @size_xs;
```
jonjonray, line 45: nit: `margin---bottom--xs();`

jonjonray (approved):
LGTM! This is looking great!

Two things:

- It might be worth going through and adding mobile screenshots of the changes for posterity
- If we're planning on merging this tonight or tomorrow, seems like it might be worth letting #frontend know since folks who import the latest components won't be able to build since the name of the announcement bubble theme prop changed  (`theme` -> `bubbleType`)

jonah-clever:
Screenshots from importing into FP:

Screen Shot 2021-08-17 at 4 17 24 PM
Screen Shot 2021-08-17 at 4 17 36 PM

Imported into LP and looking unchanged:

Screen Shot 2021-08-17 at 4 34 23 PM
Screen Shot 2021-08-17 at 4 35 54 PM

jonjonray (approved):

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -162,6 +161,72 @@
...
 
+.MessagingAttachment--ParentContainer.MessagingAttachment--ParentContainer--IsUpload {
+  overflow: visible; // ensures you can see the x icon in top right corner
+}
+
+// Family portal specific styling
+.MessagingAttachment--ParentContainer.MessagingAttachment--FamilyPortal {
+  border-radius: @size_m;
+  background-color: @neutral_white;
+  box-shadow: @size_none @size_3xs 0.375rem rgba(71, 76, 94, 0.3);
```
arsalansufi, line 172: Looks like we can remove this line now since we have the `boxShadow--medium()` on the next line

src/AnnouncementBubble/DeletedAnnouncementBubble.less
```diff
@@ -18,6 +18,25 @@
   .text--medium();
 }
 
+// Family Portal styling
+.MessagingBubble--FamilyPortal {
+  .DeletedAnnouncementBubble--container {
+    background-color: @family_portal_quoted_announcement;
+    color: @family_portal_slate;
+    border: unset;
```
arsalansufi, line 26: Can we get by with a `border: 0` here? Since as we discussed in #frontend, IE11 doesn't support `unset`. No worries if not!

src/MessagingAttachment/MessagingAttachment.less
```diff
@@ -162,6 +161,72 @@
...
+  .boxShadow--medium();
+  margin-top: @size_2xs;
+  color: @neutral_dark_gray;
+
+  .MessagingAttachment--Container {
+    border: unset;
+    border-radius: unset;
```
arsalansufi, line 179: Similar to my earlier comment, can we get by with a `border: 0` and `border-radius: 0` here?

arsalansufi (approved):
These bubbles look great!! Just a handful of quick comments.

As we discussed offline, it'll be nice to come back after this is rolled out to verify which of these styles should be specifically applied to FamPo vs. which we can apply globally.

src/AnnouncementBubble/QuotedAnnouncementBubble.less
```diff
@@ -196,6 +196,33 @@
   .text--underline();
 }
 
+// These styles only apply to Announcement replies sent by the current user, in the Family Portal,
+// i.e. guardians replying to teacher announcements
+.MessagingBubble--Message--Reply--Own
+  .QuotedAnnouncementBubble--container.QuotedAnnouncementBubble--familyPortal {
```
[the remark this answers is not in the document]
jonah-clever: That makes sense! For context, I was following the pattern of this specific file, which doesn't match other files.

src/MessagingAttachment/MessagingAttachment.less
jonah-clever: 🤦

src/AnnouncementBubble/QuotedAnnouncementBubble.less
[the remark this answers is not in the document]
arsalansufi: Got it! Thanks for that context

arsalansufi (approved):
Last set of changes LGTM!

PR #720, opened 2021-08-13
ccccccncrbbvedkkbittirjifkhflehcbfnvjvgertthadded onClick prop for RR

brianbyrhee:
# Jira: [M5G-692](https://clever.atlassian.net/browse/M5G-692)

# Overview:

With the successful rollout of M1 for RR, we are now implementing a sidebar for M2 rollout. The sidebar should render upon an onClick event for every RR tooltip inside each AnnouncementBubble. More info on this sidebar can be found [here](https://www.figma.com/file/qsfYsTriPaJtqvXFBTL3Zv/Clever-Messages-v2-(Read-Receipts-%2B-Reactjis)?node-id=84%3A0). However, because sidebars can only be rendered on LP and not in the Clever components repo and we want to trigger the onClick event within the tooltip itself, we pass the function in as a prop  to AnnouncementBubble. The component still behaves as intended as we just added one additional prop that results in a backwards-compatible change. This would be a version minor update so we would rebase and call `npm version minor`.

# Screenshots/GIFs:

Screen Shot 2021-08-13 at 1 42 36 PM

Screen Shot 2021-08-13 at 1 43 23 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
@@ -166,7 +166,7 @@ function formReadReceiptsTooltip(
         textAlign={"left"}
       >
         <div onMouseEnter={onReadReceiptsHover}>
-          <FlexBox onClick={onReadReceiptsClick}>
+          <FlexBox role="button" tabIndex={0} onClick={onReadReceiptsClick}>
             <Checkmark className={cssClass("readReceipts--icon")} />
             <span className={cssClass("readReceipts--text--desktop")}>
               Read by {readReceiptCount} {recipientString}
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.tsx
@@ -21,6 +21,7 @@ export interface Props {
   inlineErrorMsg?: string;
   numTranslatedLanguages?: number;
   onDelete?: () => void;
+  onReadReceiptsClick?: () => void;
   onReadReceiptsHover?: () => void;
   onReply?: () => void;
   readBy?: string[];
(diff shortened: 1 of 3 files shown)


src/AnnouncementBubble/NormalAnnouncementBubble.tsx
```diff
@@ -161,7 +166,7 @@ function formReadReceiptsTooltip(
         textAlign={"left"}
       >
         <div onMouseEnter={onReadReceiptsHover}>
-          <FlexBox>
+          <FlexBox onClick={onReadReceiptsClick}>
```
arsalansufi, line 169: For accessibility purposes, it's best practice to apply `onClick` handlers to truly clickable elements, like `button`s and `a`s. The Dewey `FlexBox` just renders a `div` under the hood.

I'd recommend changing this Dewey `FlexBox` to [a Dewey `Button`](https://clever.github.io/components/#/components/button) if you want to make it clickable! I think `<Button type="linkPlain">` should give you the desired styling.

arsalansufi (commented):
A quick comment about accessibility!

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
arsalansufi: As another potentially simpler alternative, if using a Dewey button will require too many styling overrides, you can also apply `role="button" tabIndex={0}` to the `FlexBox`, which will tell screen readers to treat it as a button.

brianbyrhee: Good catch! I went with the latter solution instead. The tabIndex is a nifty prop as well, something that I discovered just now!

arsalansufi (approved):
LGTM!

PR #723, opened 2021-08-17
remove hover state from translation bubble msg

jonjonray:
# Overview:
Removing the hover state from the `translated into x languages` tooltip since it's not a clickable element. Double checking with folks to make sure this is the only tweak we want to include here.

# Screenshots/GIFs:
before
Screen Shot 2021-08-17 at 12 05 50 PM

after
Screen Shot 2021-08-17 at 12 05 04 PM

# Testing:

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.153.1",
+  "version": "2.154.1",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


ashleyqw (approved):
Thanks for fixing this so fast! If it's preferable to have the icon as a separate .svg file instead of as <svg> in the component file we can do that now since it doesn't need to change color, but not sure if there's any benefit to that

jonjonray:
> If it's preferable to have the icon as a separate .svg file instead of as in the component file we can do that now since it doesn't need to change color, but not sure if there's any benefit to that

I have a feeling we may need to restyle this at some point, so it can't hurt to leave it as-is!

PR #724, opened 2021-08-18
fixed some string parsing for RR tooltip

brianbyrhee:
# Jira: [M5G-723](https://clever.atlassian.net/browse/M5G-723)

# Overview:

Some quick fix of some string parsing for the Read Receipts tooltip in AnnouncementBubble. According to the Figma [here](https://www.figma.com/file/qsfYsTriPaJtqvXFBTL3Zv/Clever-Messages-v2-(Read-Receipts-%2B-Reactjis)?node-id=84%3A0), the tooltip should should only display the top 3 names in the `readBy` string array with a suffix string of " and n-3 more..." where n is the `readBy` length. I know that we discussed that the tooltip component's props are immutable, but the Figma design does specify that we write one name per line. I wanted to discuss about whether we should abandon this feature or not. Because this is a patch, we should rebase and run `npm version patch`. Other RR functionalities have still retained, making this a backwards compatible change.

# Screenshots/GIFs:

Screen Shot 2021-08-18 at 9 47 51 PM
Screen Shot 2021-08-18 at 9 48 36 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/docs/components/AnnouncementBubbleView.jsx
+++ b/docs/components/AnnouncementBubbleView.jsx
@@ -43,16 +43,16 @@ export default class AnnouncementBubbleView extends React.PureComponent {
     const { colorTheme, theme } = this.state;
 
     const readBy = [
-      "Arsalan",
-      "Nikhil",
-      "Spencer",
-      "Ashley",
-      "Jonathan",
-      "Brian",
-      "Chloe",
-      "Cory",
-      "Jonah",
-      "Nick",
+      "Arsalan Sufi",
+      "Nikhil Bhatia",
+      "Spencer Caton",
+      "Ashley Wang",
+      "Jonathan Ray",
+      "Brian Rhee",
+      "Chloe Caelynn",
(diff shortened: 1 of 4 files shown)


nbhatia823 (commented):
>  I know that we discussed that the tooltip component's props are immutable, but the Figma design does specify that we write one name per line. I wanted to discuss about whether we should abandon this feature or not. 

I dont quite understand this 

Also, since we only want to show the updated string format once the sidebar is out, we likely want to wait to merge this in until we are very close to finishing the LP side of things too

nbhatia823 (approved):
Looks great!!! Could you add updated screenshots to the PR description?

Also before we merge it later today we will need to bump up the package version!

PR #726, opened 2021-08-19
Restyle announcement bubble reply button

arsalansufi:
**Commit-by-commit reviewing recommended**

# Jira

[M5G-635](https://clever.atlassian.net/browse/M5G-635)

# Overview

This PR restyles the announcement bubble reply button! It also fixes a bug with the translated languages list on Safari.

## Before

before

## After

_Base state_

after-1

after-5

_Focus state_

after-2

_Keyboard focus state_

after-3

_Mobile_

after-4

_Disabled_

after-6

_Without a reply button_

Just for completeness!

after-7

after-8

## Unrelated bug fix

_Before and after_

bug-before bug-after

# Testing

- [x] Chrome
- [x] Safari
- [ ] IE11

# Roll out

_Before merging_

- [x] Updated docs (updated the screenshot for the component list page)

_After merging_

- [x] Deployed updated docs (`make deploy-docs`)
- [ ] ~Posted in #eng if I made a breaking change to a beta component~ N/A


Makefile
```diff
@@ -78,7 +78,7 @@ gen-border-radius-styles:
 	@echo "Generating border-radius style definitions..."
 	@node genBorderRadius.js
 
-LINT_MAX_LESS_PROBLEMS := 102
+LINT_MAX_LESS_PROBLEMS := 96
```
nbhatia823, line 81: did this introduce some more lint issues? 😅

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
```diff
@@ -210,7 +206,7 @@ function ReplyButton({
         <img
           alt=""
           className={cssClass("replyButton--icon")}
-          src={require("./messages-black-icon.svg")}
```
nbhatia823, line 213: Looks like this icon isnt being used anywhere after this so we could remove it from the directory

Makefile
arsalansufi: It introduced a false positive actually! And it didn't feel right to appease the linter with a fix for an incorrectly identified issue, so I instead updated stylelint to mitigate the false positive.

Here was the error before I upgraded stylelint:

```
src/AnnouncementBubble/NormalAnnouncementBubble.less
 181:3  ✖  Expected selector ".NormalAnnouncementBubble--translatedLanguages--container .margin--top--m()" to come before selector ".Button.NormalAnnouncementBubble--replyButton .margin--top--m()"   no-descending-specificity
```

It's subtle, but stylelint was misinterpreting our `.margin--top--m` mixin as a CSS selector. So it was stating that `.NormalAnnouncementBubble--translatedLanguages--container .margin--top--m` was of lower specificity that `.Button.NormalAnnouncementBubble--replyButton .margin--top--m` but in truth, the two selectors are unrelated.

`.NormalAnnouncementBubble--translatedLanguages--container ` is independent of `.Button.NormalAnnouncementBubble--replyButton` -- they share no overlap.

src/AnnouncementBubble/NormalAnnouncementBubble.tsx
arsalansufi: Good catch!

Makefile
nbhatia823: Ahh, nice investigation! Good call

nbhatia823 (approved):
It seems everyday the AnnouncementBubble grows up even more 🥲

PR #725, opened 2021-08-18
[M5G-700] Show a11y outlines on tab only, not click

ashleyqw:
# Jira: 
https://clever.atlassian.net/browse/M5G-700
https://clever.atlassian.net/browse/PRTL-730
https://clever.atlassian.net/browse/PRTL-731

# Overview:
This PR hides a11y outlines on click so we only show them on tab. Also changes the outline color to be `@primary_blue_shade_2` for all these outlines. 

source: https://css-tricks.com/the-focus-visible-trick/

# Screenshots/GIFs:

https://user-images.githubusercontent.com/35714960/129939180-08cd5c19-d883-4392-93ba-9c9fd0965eb8.mov

https://user-images.githubusercontent.com/35714960/129939183-81b5923f-9fe3-4a9b-9143-020122266e62.mov

https://user-images.githubusercontent.com/35714960/129939185-9dae7bb7-dcbb-49e5-943a-de0cabd19cc9.mov

# Testing:

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
  - [x] Bumped version in `package.json`
    - Breaking change?
      - If it is a beta component run `npm version minor`
      - If the component is not in beta run `npm version major`
    - New component or backward-compatible component feature change? Run `npm version minor`
    - Only changing documentation? All good. Skip this step.
  - After creating a new component, make sure to add it to the Components List in `ComponentsView.jsx`. To do so:
    - [ ] Add a screenshot of the component in `docs/assets/img` with the format `<COMPONENT URL LINK>.png`
- After merging:
  - [ ] Deployed updated docs (`make deploy-docs`)
  - [ ] Posted in #eng if I made a breaking change to a beta component

PR #727, opened 2021-08-19
Add cursor when hovering read receipts; improve accessibility of read receipts

nbhatia823:
# Overview:

This PR fixes a few read receipt bug bash issues:
1. changes the hover state to the cursor 
2. changes it so when tabbing through, the read receipts is focused once instead of twice
3. makes it so the onClick fires when pressing enter with a keyboard

# Screenshots/GIFs:

https://user-images.githubusercontent.com/26425483/130156396-05a58ee1-da2d-4cce-8c39-875803165b0e.mov

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.less
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.less
@@ -183,10 +183,22 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 }
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
+  cursor: pointer;
   .text--semi-bold();
   padding-right: @size_2xs;
   padding-left: @size_2xs;
   white-space: pre-wrap;
+}
+
+.Button.NormalAnnouncementBubble--readReceipts--button {
+  &:focus {
+    outline: 0.1875rem solid @primary_blue_shade_2;
+    transition: none;
+  }
+
+  &:focus:not(:focus-visible) {
+    outline: 0;
+  }
 }
 
 .NormalAnnouncementBubble--readReceipts--icon {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.156.2",
+  "version": "2.156.3",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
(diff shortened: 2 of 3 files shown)


src/AnnouncementBubble/NormalAnnouncementBubble.less
```diff
@@ -199,12 +199,24 @@ button.NormalAnnouncementBubble--deleteMenuItem {
...
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
+  cursor: pointer;
   .text--semi-bold();
   padding-right: @size_2xs;
   padding-left: @size_2xs;
   white-space: pre-wrap;
 }
 
+Button.NormalAnnouncementBubble--readReceipts--button {
```
arsalansufi, line 209: I think we probably want this to be `.Button` with a dot, though I'm also wondering if Ashley's recent PR has this covered already!

https://github.com/Clever/components/pull/725

https://github.com/Clever/components/blob/3a8143c4c2d592de73e8dc7799479a881736de7a/src/Button/Button.less#L144-L151

nbhatia823: Nice ill drop it!

nbhatia823: Actually it isnt working without this code, and i think its because ashley added that to the link/linkPlain buttons and not the other types and Im using a plain here. Is the intention that the focus behavior should be the same on all button types?

nbhatia823: Also TIL that casing isn't enforced for element selectors. So that's why `Button` worked in this case without the `.`. Even `BUtton` works. But i know we discussed using `.Button` from now on

brianbyrhee (approved):
LGTM to me! Also wanted to say thanks for fixing multiple bugs at once from the Bug Bash such as the blue AnnouncementBubble outline for the selected RR sidebar.

src/AnnouncementBubble/NormalAnnouncementBubble.less
arsalansufi: Ooh great catch -- you're right that Ashley's update was for link/linkPlain buttons and not other types. On that note, applying a one-off fix here seems reasonable!

I do think it's worth checking in with Nick after the fact whether we also want to apply a global fix to the other `Button` types

PR #745, opened 2021-09-15
M5G-774 - Adjust RadioGroup implementation so that consumers can override st

clcaelynn-clever:
# Jira

[M5G-774](https://clever.atlassian.net/browse/M5G-774)

# Overview

Adjust RadioGroup implementation so that consumers can override styling to adjust margins and padding in a way that allows treating the selected-with-additional-text group differently, as currently there is no way to do so.

# Testing

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out

💯 

- Before merging
  - [n/a] Updated docs
    - Breaking change?
- After merging:

# Screenshots/GIFs

## clever-components: unchanged

Kapture 2021-09-15 at 16 33 32

## family-portal, desktop modal: can override styling

Screen Shot 2021-09-15 at 4 25 40 PM
Screen Shot 2021-09-15 at 4 25 47 PM
Screen Shot 2021-09-15 at 4 25 56 PM

## family-portal, mobile fullscreen: can override styling
Screen Shot 2021-09-15 at 4 27 18 PM
Screen Shot 2021-09-15 at 4 27 30 PM
Screen Shot 2021-09-15 at 4 27 52 PM

### DIFF
--- a/src/RadioGroup/RadioGroup.less
+++ b/src/RadioGroup/RadioGroup.less
@@ -5,7 +5,9 @@
     .margin--top--l;
   }
 
-  .RadioGroup--radio {
+  // The latter class (selectedWithAdditionalText) exists purely to be overriden by the consumer.
+  .RadioGroup--radio,
+  .RadioGroup--radio--selectedWithAdditionalText {
     &:not(:first-child) {
       .margin--top--s;
     }
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.164.1",
+  "version": "2.164.2",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/RadioGroup/RadioGroup.tsx
+++ b/src/RadioGroup/RadioGroup.tsx
@@ -33,6 +33,7 @@ const cssClass = {
   CONTAINER: "RadioGroup",
   LABEL: "RadioGroup--label",
   RADIO: "RadioGroup--radio",
+  RADIO_SELECTED_WITH_ADDITIONAL_TEXT: "RadioGroup--radio--selectedWithAdditionalText",
   ADDITIONAL_TEXT: "RadioGroup--additionalText",
 };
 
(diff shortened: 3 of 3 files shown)


src/RadioGroup/RadioGroup.less
```diff
@@ -5,7 +5,9 @@
     .margin--top--l;
   }
 
-  .RadioGroup--radio {
+  // The latter class (selectedWithAdditionalText) exists purely to be overriden by the consumer.
+  .RadioGroup--radio,
+  .RadioGroup--radio--selectedWithAdditionalText {
```
clcaelynn-clever, line 10: Ideally `.RadioGroup--radio--selectedWithAdditionalText` would simply be its own empty class - so that a consumer could override it if desired, but it would otherwise not affect the component's appearance or functionality whatsoever - but we don't generally allow those.

This seems like a reasonable compromise which I wouldn't expect to break styling anywhere? Open to other ideas.

ashleyqw: This seems okay to me! I'm not sure what ends up needing to be overridden once this gets imported into fampo but I think when I tried the first time I had to override the `.margin--top--xs;` in line 24 below across the board in fampo (although I hadn't needed to when I was testing by copying the components dist folder over). But I think I added that line in the first place to make it look right in Dewey. 

Want to check if you're seeing the same thing because it's possible that that top margin isn't actually needed? (if it just gets overridden every time elsewhere)
Basically it felt a little fishy after I actually imported it info fampo so just want to check if you think it's still necessary

ashleyqw (approved):
Thanks Chloe! lgtm
Thanks for the careful checking of that last language in the list as well

src/RadioGroup/RadioGroup.less
clcaelynn-clever: That margin top _does_ actually seem to make a difference in Dewey, so I'll leave it alone! Don't wanna alter the component just to suit our use case in a way that's not part of its inherent design in Dewey after all, even if it's been a bit hard to tell which things are 'issues' and which are just custom needs for our current use case in fampo. 😅

PR #746, opened 2021-09-17
Disc 2758 enhance text input2

antoniodangond:
# Jira: [DISC-2758](https://clever.atlassian.net/browse/DISC-2758)

# Overview:

Modifies `TextInput2` error validation behavior:

* Changes `initialIsInError` prop [(unused)](https://github.com/search?q=org%3AClever+initialIsInError&type=code) to `errorOnEmpty`
* If `errorOnEmpty`, empty error state is displayed while input is out of focus and value is empty, but is removed when input is in focus and value is empty
  * The error removal while value is empty allows normal error validation to take place 
* Only displays errorValidation error on "blur" event
  * Once the event state is set, error is removed either by correcting the error or removing the value (no error state will be set while the value is empty)

# Screenshots/GIFs:
When `errorOnEmpty` is false:
* error is not shown while value is empty
* after entering invalid value, error is not shown until focus is removed from the input
* error is removed by removing the value
* error is removed by correcting the error (i.e. entering a valid value)
error

When `errorOnEmpty` is true:
* error is displayed while input is out of focus and value is empty
* empty error is removed on focus
* entering a value allows normal error validation behavior to occur (as described above)
errorOnEmpty

# Testing:

**Note:** Added tests use Enzyme's `mount` function instead of `shallow`, because `useEffect` calls [aren't made during shallow renders](https://github.com/enzymejs/enzyme/issues/2086).

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/TextInput2/TextInput2.tsx
+++ b/src/TextInput2/TextInput2.tsx
@@ -20,7 +20,7 @@ export interface Props {
   icon?: React.ReactNode;
   requirement?: Values<typeof FormElementRequirement>;
   obscurable?: boolean;
-  initialIsInError?: boolean;
+  errorOnEmpty?: boolean;
   // returns an error message, null for no error
   errorValidation?: (value: string) => string | null;
   value: string;
@@ -61,7 +61,7 @@ const TextInput2: React.FC<Props> = ({
   helpText,
   icon,
   requirement,
-  initialIsInError,
+  errorOnEmpty,
   errorValidation,
   obscurable,
   value,
@@ -72,38 +72,29 @@ const TextInput2: React.FC<Props> = ({
 }) => {
   const id = name;
   const [isFocused, setIsFocused] = useState(false);
-
   const [isObscured, setIsObscured] = useState(true);
   const inputType = obscurable && isObscured ? "password" : "text";
-
-  // empty string is an error state with no message (e.g. required)
-  const [errorMessage, setErrorMessage] = useState(initialIsInError ? "" : null);
-
-  useEffect(() => {
-    if (requirement === FormElementRequirement.REQUIRED && value === "") {
-      setErrorMessage(initialIsInError ? "" : null);
-    }
-  }, [initialIsInError]);
+  // NOTE: setting errorMessage to an empty string displays an error state with no message (e.g. required)
+  const [errorMessage, setErrorMessage] = useState(null);
 
   useEffect(() => {
-    // don't show error if nothing has happened yet
-    if (!isFocused && errorMessage === null) {
-      return;
-    }
-
-    if (requirement === FormElementRequirement.REQUIRED && value === "") {
-      setErrorMessage("");
+    // if value is empty, either:
+    // 1) display empty error if errorOnEmpty is true and input isn't in focus
+    // 2) remove error
+    if (!value) {
+      setErrorMessage(errorOnEmpty && !isFocused ? "" : null);
       return;
     }
 
     const newErrorMessage = errorValidation(value);
-    if (newErrorMessage) {
+    // only show error when input isn't in focus
+    if (newErrorMessage && !isFocused) {
       setErrorMessage(newErrorMessage);
       return;
     }
-
-    setErrorMessage(null);
-  }, [value, isFocused]);
+    // remove error if it has been corrected
+    if (!newErrorMessage) setErrorMessage(null);
+  }, [value, isFocused, errorOnEmpty]);
 
   return (
     <div className={classnames(cssClass.CONTAINER, formElementSizeClassName(size), className)}>
@@ -170,7 +161,7 @@ const TextInput2: React.FC<Props> = ({
 };
 
 TextInput2.defaultProps = {
-  initialIsInError: false,
+  errorOnEmpty: false,
   errorValidation: () => null,
   size: FormElementSize.FULL_WIDTH,
 };
--- a/docs/components/TextInput2View.jsx
+++ b/docs/components/TextInput2View.jsx
@@ -32,7 +32,7 @@ export default class TextInput2View extends React.PureComponent {
     showIcon: false,
     requirement: FormElementRequirement.REQUIRED,
     obscurable: false,
-    initialIsInError: false,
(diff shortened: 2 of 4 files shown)


src/TextInput2/TextInput2.tsx
```diff
@@ -83,7 +89,16 @@ const TextInput2: React.FC<Props> = ({
 
   useEffect(() => {
     // don't show error if nothing has happened yet
-    if (!isFocused && errorMessage === null) {
+    if (!showErrorOnBlurOnly && !isFocused && errorMessage === null) {
```
antoniodangond, line 92: This change should be backwards compatible, since existing implementations of `TextInput2` won't provide the `showErrorOnBlurOnly` prop- hence, `! showErrorOnBlurOnly` will evaluate to `true` for those component implementations.

src/TextInput2/TextInput2.tsx
```diff
@@ -92,7 +107,6 @@ const TextInput2: React.FC<Props> = ({
       return;
     }
 
-    const newErrorMessage = errorValidation(value);
```
antoniodangond, line 95: This variable definition is just moved to the new line 96, so that it can be used earlier in case `showErrorOnBlurOnly` is `true`.

src/TextInput2/TextInput2.tsx
```diff
@@ -134,9 +148,16 @@ const TextInput2: React.FC<Props> = ({
           value={value}
           placeholder={placeholder}
           disabled={requirement === FormElementRequirement.DISABLED}
-          onChange={onChange}
+          onChange={(e) => {
+            // set isInputDirty to true after value is changed for the first time
+            if (!isInputDirty) setIsInputDirty(true);
+            onChange(e);
```
antoniodangond, line 154: Maintains backwards compatibility.

src/TextInput2/TextInput2.tsx
```diff
@@ -94,7 +100,11 @@ const TextInput2: React.FC<Props> = ({
 
     const newErrorMessage = errorValidation(value);
     if (newErrorMessage) {
-      setErrorMessage(newErrorMessage);
+      // if showErrorOnBlurOnly, only display error after user has
+      // modified initial input value and removed focus from input
+      if (!showErrorOnBlurOnly || (!isFocused && isInputDirty)) {
```
antoniodangond, line 105: Same backwards compatibility comment as above^

src/TextInput2/TextInput2.tsx
```diff
@@ -83,7 +91,7 @@ const TextInput2: React.FC<Props> = ({
 
   useEffect(() => {
```
johnhuangclever, line 92: After reading about some best practices for error messaging, we can probably rewrite this entire thing. Something like this maybe:
```
useEffect(() => {
  // if input is empty and requirement error is active, then do nothing and return
  // if input is empty, then clear error and return


  // check error validation function
  // if no error validation, clear error and return
  // if errormessage exists and is not focused, then set error and return
});
```

johnhuangclever (commented):
Nice PR Antonio! Looks like you put alot of thought into the problems with this component and your changes to solve them. I wouldn't worry about backwards compatibility too much here, as I can't imagine anyone would rely on the live error message functionality for anything. Let me know if you see it differently.

I didn't put too much thought into the initial implementation of the error messaging, and this PR has made me do some reading. I found some nice articles and looked at some component libraries:
https://baymard.com/blog/inline-form-validation
https://cxl.com/blog/form-validation/
https://stripe.dev/elements-examples/

What I understand here is that we probably want some of these principles:
* Errors should show only on blur, they clear instantly if they are no longer in error
* If the input has no text entered, it should never show an error
* The `REQUIRED` error should probably come via a prop and not be internal to the component
  * We can use `initialIsInError` to handle REQUIRED errors that come from an event. e.g. clicking submit triggering an error state. [It doesn't look like this is being used anywhere right now](https://github.com/search?q=org%3AClever+initialIsInError&type=code), so potential to rename to `requirementError` or something 
* `onBlur` and `onFocus` prop callbacks are good to have in general. However, if we wanted to force the user into input formatting, then rather than using onBlur we could probably add two function props (naming is hard)
  * `formatValue`: before onChange, call this function then pass the resulting string into onChange. e.g. no trailing/leading spaces, only numbers, auto-case, limit number of characters, etc.
  * `formatDisplay`: if we wanted other text in the input, we can change how the text is formatted. e.g. phone numbers (XXX) XXX-XXXX, MM/YYYY dates, etc.


Again, I don't think preserving backwards compatibility is too critical. Let me know your thoughts. Thanks!

johnhuangclever (commented):
Nice tests! I think if we add the `errorOnEmpty` prop to the `useEffect` dependency list, we can simplify a little bit

src/TextInput2/TextInput2.tsx
```diff
@@ -72,37 +72,35 @@ const TextInput2: React.FC<Props> = ({
...
-    const newErrorMessage = errorValidation(value);
-    if (newErrorMessage) {
+    // only show new error message if there's a value and input isn't in focus
+    if (!isFocused && newErrorMessage) {
       setErrorMessage(newErrorMessage);
-      return;
     }
-
-    setErrorMessage(null);
   }, [value, isFocused]);
```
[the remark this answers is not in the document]
antoniodangond: Ah good call!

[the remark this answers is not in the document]
antoniodangond: Good call! https://github.com/Clever/components/pull/746/commits/2733131a6ee5bd928ce5dd008e2bb4a0675ecaa9

johnhuangclever (approved):
🙌  Let me know how the implementation goes! Thanks for all your help here and humoring my comments

antoniodangond:
The implementation seems to be working how we want it! I updated the GIFs in the PR description to show the new implementation. Thanks again for the thorough reviews.

PR #755, opened 2021-09-29
Don't read additional floating buttons from screenreader unless the button group is active

sandyasankarram:
# Jira: [PRTL-1072](https://clever.atlassian.net/browse/PRTL-1072)
# Jira: [PRTL-1073](https://clever.atlassian.net/browse/PRTL-1073)

# Overview:
Previously, if a user navigated to a floating button but did not click to open it, it would still read out all its additional buttons on the screen reader. With this fix, the additional button options are only read when the floating button is active.

# Screenshots/GIFs:

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/FloatingButton/FloatingButton.tsx
+++ b/src/FloatingButton/FloatingButton.tsx
@@ -1,16 +1,16 @@
-import * as _ from "lodash";
-import * as classnames from "classnames";
+import "./FloatingButton.less";
+
 import * as PropTypes from "prop-types";
 import * as React from "react";
 import * as RootCloseWrapper from "react-overlays/lib/RootCloseWrapper";
+import * as _ from "lodash";
+import * as classnames from "classnames";
 
 import { Button } from "../Button/Button";
 import FlexBox from "../flex/FlexBox";
 import { Icon } from "../Icon/Icon";
-import { breakpointS } from "../utils/Constants";
 import { Values } from "../utils/types";
-
-import "./FloatingButton.less";
+import { breakpointS } from "../utils/Constants";
 
 export interface Props {
   additionalButtons?: any[];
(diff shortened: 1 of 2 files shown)


src/FloatingButton/FloatingButton.tsx
```diff
@@ -1,16 +1,16 @@
...
+import * as classnames from "classnames";
 
 import { Button } from "../Button/Button";
 import FlexBox from "../flex/FlexBox";
 import { Icon } from "../Icon/Icon";
-import { breakpointS } from "../utils/Constants";
 import { Values } from "../utils/types";
-
-import "./FloatingButton.less";
+import { breakpointS } from "../utils/Constants";
```
prime-time, line 13: Not a fan of how these were ordered, eh 😛

prime-time (approved):
Nice 🎈🔘 🔇

prime-time:
Don't forget to version this. A patch would make sense in this case.

src/FloatingButton/FloatingButton.tsx
sandyasankarram: My formatter did this! Hmm is this not all right?

prime-time: Ah I figured it was a formatter thing. Which formatter do you use? It's not a big deal, but I was confused for a second trying to see which dependency was added or removed.

sandyasankarram: I don't have anything special set up, i just have set eslint to format the doc when I save it. I think it would follow the rules set in the formatting files in the repo?

PR #764, opened 2021-10-15
[TextInput2] support autofocus

prime-time:
# Jira:

# Overview:
I want to swap a `TextInput`s with `TextInput2`, however, the `TextInpu` is using the `autoFocus` prop which the latter does not support. If you look at the source code for `TextInput`, you'll notice that it doesn't explicitly list `autoFocus` as a prop. It does however allow clients to pass additional props through to the underlying `input` element https://github.com/Clever/components/blob/a6415b0083a2f1c7f393ebff51148013da724354/src/TextInput/TextInput.tsx#L244. It seems like we're moving away from that pattern, and prefer to list a component's props explicitly which is what I've done in this pull request. 

https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus

# Screenshots/GIFs:

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:


prime-time:
👁️ 🐽 👁️

PR #771, opened 2021-11-17
[M5G-851] Hovercrafts can now land on tooltips

ashleyqw:
# Jira: 
https://clever.atlassian.net/browse/M5G-851

# Overview:
In order to be accessible, tooltips should be hoverable and escape-able. Tooltips were previously not hoverable since the tooltip would disappear before the mouse can be moved to it. This PR fixes that which required some changes internally (mainly, we can't use OverlayTrigger anymore since we need to more manually control `show` of the tooltip).

# Screenshots/GIFs:

Before: 

https://user-images.githubusercontent.com/35714960/142266346-6e7166cf-98e3-4a09-9bec-cda3a1f38f60.mov

After:

https://user-images.githubusercontent.com/35714960/142266390-375b390e-6638-43f7-9c97-2ea2d45fa46f.mov

# Testing:

- Checked other places that Tooltip is used in components
- Checked new Tooltip in launchpad use case on actual a11y ticket
- Checked Tooltip in a few other use cases in launchpad
- Tested on 3 browsers
- Tested keyboard a11y
- Tested disabled state

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [x] Firefox
  - [ ] IE11

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AnnouncementBubble/NormalAnnouncementBubble.less
+++ b/src/AnnouncementBubble/NormalAnnouncementBubble.less
@@ -209,7 +209,6 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 }
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
-  cursor: pointer;
   .text--semi-bold();
   padding-right: @size_2xs;
   padding-left: @size_2xs;
--- a/src/Tooltip/Tooltip.tsx
+++ b/src/Tooltip/Tooltip.tsx
@@ -1,9 +1,10 @@
 import * as _ from "lodash";
 
-import * as classnames from "classnames";
+import * as React from "react";
 import * as BootstrapTooltip from "react-bootstrap/lib/Tooltip";
 import * as Overlay from "react-bootstrap/lib/Overlay";
-import * as React from "react";
+import * as RootCloseWrapper from "react-overlays/lib/RootCloseWrapper";
+import * as classnames from "classnames";
 import * as PropTypes from "prop-types";
 
 import { Values } from "../utils/types";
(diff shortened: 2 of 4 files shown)


src/AnnouncementBubble/NormalAnnouncementBubble.less
```diff
@@ -209,7 +209,6 @@ button.NormalAnnouncementBubble--deleteMenuItem {
 }
 
 .NormalAnnouncementBubble--readReceipts--tooltip {
-  cursor: pointer;
```
ashleyqw, line 212: Don't want the list of names in read receipt tooltip to look clicky since it's not

src/Tooltip/Tooltip.tsx
```diff
@@ -107,26 +133,28 @@ export default class Tooltip extends React.Component<Props> {
...
+        <>
+          {React.cloneElement(child, {
+            ref: this.tooltipTarget,
+            tabIndex: 0,
+            "aria-describedby": this.id,
+            onFocus: handleShowTooltip,
+            onBlur: handleHideTooltip,
+            onMouseEnter: handleMouseEnter,
+            onMouseLeave: handleMouseLeave,
+            onMouseDown: !clickTrigger || hide ? undefined : handleOnClick,
```
jonjonray, line 146: nit:  Can we move the ternary operation on the right (`hide ? undefined : handleOnClick`) to a variable or handle this logic from within `handleOnClick`?

jonjonray: Also might be worth including a comment explaining this functionality

src/Tooltip/Tooltip.tsx
```diff
@@ -96,8 +105,25 @@ export default class Tooltip extends React.Component<Props> {
...
+      if (hide) return;
+      setTimeout(() => this.setState({ showTooltip: true }), delayMs);
+    };
+    const handleHideTooltip = () => {
+      if (hide) return;
+      setTimeout(() => this.setState({ showTooltip: false }), delayHideMs);
+    };
+    const handleMouseEnter = clickTrigger ? undefined : handleShowTooltip;
+    const handleMouseLeave = clickTrigger ? undefined : handleHideTooltip;
+    const handleOnClick = this.state.showTooltip ? handleHideTooltip : handleShowTooltip;
```
jonjonray, line 118: nit: can these callbacks check for `clickTrigger` in the function and no-op if that's the case? Something like 

```
    const handleShowTooltip = () => {
      if (hide || clickTrigger) return;
      setTimeout(() => this.setState({ showTooltip: true }), delayMs);
    };
```

It'll save you the extra assignments and you can directly reference these in the component

jonjonray (approved):
LGTM!

Given that we're changing the root element for this component, we should be careful about the rollout for this change since we use this component in a lot of places in-product and we might have some application-specific styling overrides that may change as a result

src/Tooltip/Tooltip.tsx
ashleyqw: Don't think we can since onFocus and onBlur use the handleShow/HideTooltip functions and the onMouseEnter/Leave use the handleMouseEnter/Leave functions, since we want onFocus and onBlur to work regardless of whether it's a clickTrigger tooltip or not for keyboard a11y

PR #778, opened 2022-05-02
Pre-select rows in Table2 and announcement bubble fix

jonjonray:
# Overview:

This PR is changing two things:
- Adding a new prop that allows rendering the selectable table2 component with a pre-populated selection. Intentionally kept this only for non-lazy loading since the state management for selections would make it difficult for persisting pre-selections vs actual selections.
- Fixing an issue in `UnavailableAnnouncementBubble` that was causing it to break when importing into SD2 due to a require statement that had an inline dynamic value. Moved the image imports to the top of the file instead

# Screenshots/GIFs:

# Testing:

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/AnnouncementBubble/UnavailableAnnouncementBubble.tsx
+++ b/src/AnnouncementBubble/UnavailableAnnouncementBubble.tsx
@@ -1,7 +1,8 @@
 import * as React from "react";
 import * as cx from "classnames";
 import { FlexBox } from "../";
-
+const darkColorThemeIcon = require("./megaphone-slash-darkColorTheme.svg");
+const lightColorThemeIcon = require("./megaphone-slash-lightColorTheme.svg");
 import "./UnavailableAnnouncementBubble.less";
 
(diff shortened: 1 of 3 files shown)


src/AnnouncementBubble/UnavailableAnnouncementBubble.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from "react";
 import * as cx from "classnames";
 import { FlexBox } from "../";
-
+const darkColorThemeIcon = require("./megaphone-slash-darkColorTheme.svg");
```
adamvictorclever, line 4: This changes LGTM

jkwik (approved):
Lgtm!

clcaelynn-clever:
Thanks so much for fixing this, @jonjonray!

PR #780, opened 2022-05-05
[DID-605] Label: update colors

ashleyqw:
# Jira:
https://clever.atlassian.net/browse/DID-605

# Overview:
For the nav redesign, we need to make Label component colors not transparent (since the background of the nav bar is dark, the colors look whack). Also adds new color mixins for these new hex values

# Screenshots/GIFs:

Before in sd2

Screen Shot 2022-05-02 at 10 43 48 AM

After in sd2

Screen Shot 2022-05-05 at 11 52 44 AM

now in dewey (mov upload is failing in github but i'll do anything not to rerecord it lol #lazy):
https://drive.google.com/file/d/1aNJIsQNOK45vrGE3nnoCDq2RvFsv-IRg/view?usp=sharing

# Testing:

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/Label/Label.less
+++ b/src/Label/Label.less
@@ -26,6 +26,11 @@
   cursor: help;
   // Relative positioning needed to provide tooltip anchor.
   position: relative;
+
+  &:focus, 
+  &:active {
+    .border--s(@neutral_dark_gray);
+  }
 }
 
 .Label--s {
--- a/src/Label/Label.less
+++ b/src/Label/Label.less
@@ -27,7 +27,7 @@
   // Relative positioning needed to provide tooltip anchor.
   position: relative;
 
-  &:focus, 
+  &:focus,
   &:active {
     .border--s(@neutral_dark_gray);
   }
@@ -48,23 +48,23 @@
 }
 
 .Label--new-feature {
-  background-color: #dbc2fe;
+  background-color: @label_purple;
 }
 
 .Label--success {
-  background-color: #b9ecda;
+  background-color: @label_green;
 }
 
 .Label--warning {
-  background-color: #fbe7bc;
+  background-color: @label_yellow;
 }
 
 .Label--error {
-  background-color: #febdc4;
+  background-color: @label_red;
 }
 
 .Label--category-2 {
-  background-color: #ffcbe6;
+  background-color: @label_pink;
 }
 
 .Label--tag {
--- a/src/Label/Label.less
+++ b/src/Label/Label.less
@@ -39,92 +39,30 @@
 }
 
 .Label--default {
-  .colorize(@neutral_silver);
+  background-color: @neutral_silver;
 }
 
 .Label--new-feature {
-  .colorize(@accent_teal_tint_2);
+  background-color: #dbc2fe;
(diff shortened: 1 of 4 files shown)


src/Label/Label.less
```diff
@@ -26,6 +26,11 @@
   cursor: help;
   // Relative positioning needed to provide tooltip anchor.
   position: relative;
+
+  &:focus,
+  &:active {
+    .border--s(@neutral_dark_gray);
+  }
```
ashleyqw, line 33: Jemma approved this

src/Label/Label.less
```diff
@@ -39,92 +44,30 @@
...
 .Label--error {
-  .colorize(@alert_red_tint_2);
-}
-
-.Label--category-1 {
-  .colorize(@accent_purple_tint_2);
+  background-color: @label_red;
 }
 
 .Label--category-2 {
```
ashleyqw, line 66: unfortunate naming but this is still used in prod (I think just in apps dash) so I can't get rid of it, but design also didn't provide this as an option technically in the official dewey figma I think https://www.figma.com/file/IBVNvo2dI221k4Dye4Hg0S/Dewey-2.0-WIP?node-id=21222%3A943

src/Label/Label.tsx
```diff
@@ -29,14 +29,14 @@ const Color = {
   ERROR: "error",
 
   // Accents:
-  CATEGORY_1: "category-1",
```
ashleyqw, line 32: this was purple and now new-feature is purple

src/less/colors.less
```diff
@@ -58,6 +58,13 @@
 @accent_purple_tint_1: #9244fa;
 @accent_purple_tint_2: #a460ff;
 
+// Label colors:
+@label_green: #b9ecda;
+@label_pink: #ffcbe6;
+@label_purple: #dbc2fe;
+@label_red: #febdc4;
+@label_yellow: #fbe7bc;
```
ashleyqw, line 66: not sure if there's a better naming scheme here (does it make more sense to add it as a tint or something instead?), Jemma seems okay with it though

SpiffyUrsa: If these are referring to just purely the label colors then I think this naming scheme is fine. If we get different shades of each color then we can consider adding those as tints.

src/Label/Label.less
SpiffyUrsa: Good to know! We aren't using `label--category-1` anywhere though. Is that correct?

SpiffyUrsa (approved):
LGTM once you finish testing!

src/Label/Label.less
ashleyqw: We are using the color option CATEGORY_1 but not the class. In Label.tsx we map colors to classes, so CATEGORY_1 now maps to new-feature which is the purple one, so it'll stay purple (see the Label.tsx file)

PR #799, opened 2022-08-16
ID 2132

samuelwshen:
[ticket](https://clever.atlassian.net/browse/ID-2132)

# Overview:
I think we're just missing a role here. Looking at [this article](https://www.deque.com/blog/creating-accessible-svgs/) at pattern 9, we're just missing `role` and the results table at the bottom seems to be reasonable for what we want (reads title, "image")

Unfortunately since the issue seems to show up only on iOS it's pretty hard to test this locally, so I want to 🚢 this, update the dev version of the consuming repo and test a dev link that way.


ashleyqw (approved):

zohnnyclever (approved):
LGTM! just post the results of your test on production to this PR and ping me once you've tested it.

PR #800, opened 2022-08-17
[PRTL-2035] removes `aria-selected` attribute from `MessagingThreadListItem`

cofauver:
# Jira ticket: [PRTL-2035](https://clever.atlassian.net/browse/PRTL-2035)

# Overview:

This PR removes the `aria-selected` attribute from the `MessagingThreadListItem` because it was not an allowable aria attribute for this type of element. In order to have an `aria-selected` attribute, the element must have a `role` of `gridcell`, `option`, `row`, or `tab` ([documentation here](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-selected)), but each of these require a parent element to have a specific containing `role` as well. Because parent elements are declared in the repos that import this `components` repo, we decided that the `role` addition required too much knowledge to live across the boundaries of several repositories and that the benefit was not worth the challenge.

# Screenshots/GIFs:

image

# Testing:

To QA, pull down this branch and:

1. On the branch, run `make build` in components
2. Take the dist folder and drop it into your local fampo's node_modules/clever-components directory to replace the dist that's there
3. Run axe audit to see if [the issue](https://clever.atlassian.net/browse/PRTL-2035) has gone away ("Elements must only use allowed ARIA attributes")
4. Open both the updated local version and the current clever-dev to run a comparison test with VoiceOver to ensure that the functionality hasn't gotten worse

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.195.0",
+  "version": "2.196.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/MessagingThreadListItem/MessagingThreadListItem.tsx
+++ b/src/MessagingThreadListItem/MessagingThreadListItem.tsx
@@ -148,7 +148,6 @@ export const MessagingThreadListItem: React.FC<
         )}
         onClick={onClick}
         alignItems={ItemAlign.CENTER}
-        aria-selected={selected}
       >
         <div className={cssClasses.ICON}>{icon}</div>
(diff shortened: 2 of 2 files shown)


Maddiebusta (approved):
Makes sense to just do this. looks good!

zohnnyclever:
@cofauver do you plan on updating the version of `clever-components` in the family-portal `package.json` or waiting?

PR #804, opened 2022-09-06
[PRTL-2505] add aria-label to multi-select toggle carat button

cofauver:
# Jira ticket: [PRTL-2505](https://clever.atlassian.net/browse/PRTL-2505)

# Overview:
This PR adds an `aria-label` to the carat button ( 🔼 / 🔽 ) on the multiselect component (it toggles if the dropdown options are showing on the multiselect). This allows assistive technologies like screen readers to read the button and indicate what it's for.

# Screenshots/GIFs:

## Before

https://user-images.githubusercontent.com/6520345/188757547-7453afaf-ac52-492d-80ed-0a82dfcfd4f9.mov

## After

https://user-images.githubusercontent.com/6520345/188757558-aff2be73-f84b-43da-9a71-4d7d4cc3ef58.mov

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/MultiSelect/MultiSelect.tsx
+++ b/src/MultiSelect/MultiSelect.tsx
@@ -87,6 +87,12 @@ export function getSelectableOptions(
   return [...creatableOption, ...selectableOptions];
 }
 
+// pulling out English string literals to eventually be i18n friendly
+// when i18n is prioritized for these components
+const i18nStrings = {
+  TOGGLE_BUTTON_LABEL: "toggle dropdown options",
+};
+
 /*
   Multi-item select drop down component
 */
@@ -261,7 +267,7 @@ const MultiSelect: React.FC<Props> = ({
           />
         </div>
         <button
-          aria-label="toggle dropdown options"
+          aria-label={i18nStrings.TOGGLE_BUTTON_LABEL}
           className={cssClass.CARET_BUTTON}
           {...getToggleButtonProps()}
         >
--- a/src/MultiSelect/MultiSelect.tsx
+++ b/src/MultiSelect/MultiSelect.tsx
@@ -260,7 +260,11 @@ const MultiSelect: React.FC<Props> = ({
             )}
           />
         </div>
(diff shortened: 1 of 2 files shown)


src/MultiSelect/MultiSelect.tsx
```diff
@@ -260,7 +260,11 @@ const MultiSelect: React.FC<Props> = ({
             )}
           />
         </div>
-        <button className={cssClass.CARET_BUTTON} {...getToggleButtonProps()}>
+        <button
+          aria-label="toggle dropdown options"
```
zohnnyclever, line 264: My only issue with this is that it's in English.

if you can confirm that nothing else in `clever-components` is internationalized, then this change is fine and we deal w/ i18n with `dew3y`.  

The alternative approach is that the user of this component passes in the aria-label value, which is awkward b/c it means the client has too much knowledge about the implementation of this component.

The right way is to actually support i18n with the component itself.

so long winded way of saying this is not an issue, if indeed nothing is i18n'ized in `clever-component`

zohnnyclever (approved):
LGTM! (just see the i18n comment)

src/MultiSelect/MultiSelect.tsx
cofauver: @zohnnyclever [here is](https://github.com/Clever/components/blob/master/src/MessagingBubble/NormalMessagingBubble.tsx#L42-L50) the only mention of `i18n` in the components repo. I can take this approach if it makes the most sense, but no other component seems to have taken these steps.

zohnnyclever: oh, i remember doing that now. 

Yeah, i think you can do the same thing by adding a prominent comment about it. it's unfortunate b/c i can't think of a solution to this except for having `dew3y` support i18n and us migrating this component over to `dew3y`.

PR #808, opened 2022-09-15
PRTL-2601: add alt text to table2beta actions and Checkboxes

jonah-clever:
# Jira: [PRTL-2601](https://clever.atlassian.net/browse/PRTL-2601)

# Overview:

A11y project, need to add alt text to action icons on Student Analytics page. This is handled in the Table2Beta component, by adding alt text to icon images using the title of the action

Checkboxes used in T2B were also missing alt-text, except those which had a label. For the checkboxes without a label, I added a generic aria-label "select/unselect" as this component could be used elsewhere. 

# Screenshots/GIFs:

# Testing:

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
    - **New component or backward-compatible component feature change? Run `npm version minor`**
- After merging:

### DIFF
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -86,7 +86,7 @@ export default class Checkbox extends React.PureComponent<Props> {
     return (
       <button
         aria-checked={ariaCheckedState}
-        aria-labelledby={hasLabel && this._labelID}
+        aria-labelledby={hasLabel ? this._labelID : undefined}
         // TODO: aria-label needs to be translated
         aria-label={!hasLabel ? "select/unselect" : undefined}
         className={classnames(
--- a/src/Checkbox/Checkbox.tsx
+++ b/src/Checkbox/Checkbox.tsx
@@ -81,11 +81,14 @@ export default class Checkbox extends React.PureComponent<Props> {
     if (ariaCheckedState && partial) {
       ariaCheckedState = "mixed";
     }
+    const hasLabel = !!children;
 
     return (
       <button
         aria-checked={ariaCheckedState}
-        aria-labelledby={this._labelID}
+        aria-labelledby={hasLabel && this._labelID}
+        // TODO: aria-label needs to be translated
+        aria-label={!hasLabel ? "select/unselect" : undefined}
         className={classnames(
           cssClass.CONTAINER,
           checked && cssClass.CHECKED,
--- a/src/Table2Beta/Table.tsx
+++ b/src/Table2Beta/Table.tsx
@@ -528,7 +528,7 @@ export class Table2Beta extends React.Component<Props, State> {
                 <img
                   className={cssClass.SINGLE_ACTION_TRIGGER}
                   src={require("./ellipsis.svg")}
-                  // TODO: is this a good alt text for the ellipsis icon?
+                  // TODO: alt-texts need to be translated
                   alt="more actions"
                 />
               </Button>
--- a/src/Table2Beta/Table.tsx
+++ b/src/Table2Beta/Table.tsx
@@ -462,6 +462,27 @@ export class Table2Beta extends React.Component<Props, State> {
     return this._getLazyData();
   }
 
+  actionIconAndHoverIcon = (action: ActionInput) => {
+    return (
+      <>
+        {action.icon && (
+          <img
+            className={cssClass.ACTION_ICON}
+            src={action.icon}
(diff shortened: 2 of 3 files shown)


src/Checkbox/Checkbox.tsx
```diff
@@ -81,11 +81,14 @@ export default class Checkbox extends React.PureComponent<Props> {
     if (ariaCheckedState && partial) {
       ariaCheckedState = "mixed";
     }
+    const hasLabel = !!children;
```
jonah-clever, line 84: Checkboxes with children use the child as the aria-labelled by. In cases where the checkbox has no child, there is nothing to act as an aria-label

src/Checkbox/Checkbox.tsx
```diff
@@ -81,11 +81,14 @@ export default class Checkbox extends React.PureComponent<Props> {
...
 
     return (
       <button
         aria-checked={ariaCheckedState}
-        aria-labelledby={this._labelID}
+        aria-labelledby={hasLabel && this._labelID}
```
jonah-clever, line 89: this could be a ternary, like the aria-label below. not sure if its necessary though

src/Table2Beta/Table.tsx
```diff
@@ -509,7 +525,12 @@ export class Table2Beta extends React.Component<Props, State> {
           <Menu
             trigger={
               <Button size="small" className={cssClass.ACTION_MENU} type="link">
-                <img className={cssClass.SINGLE_ACTION_TRIGGER} src={require("./ellipsis.svg")} />
+                <img
+                  className={cssClass.SINGLE_ACTION_TRIGGER}
+                  src={require("./ellipsis.svg")}
+                  // TODO: alt-texts need to be translated
+                  alt="more actions"
```
jonah-clever, line 532: is this text good?

cofauver:
@jonah-clever reviewing this now! Can you please add a screenshot to the PR description just to give an idea of the appearance of the component in context?

src/Table2Beta/Table.tsx
```diff
@@ -472,10 +472,18 @@ export class Table2Beta extends React.Component<Props, State> {
...
               {singleActions[0].hoverIcon && (
-                <img className={cssClass.ACTION_HOVER_ICON} src={singleActions[0].hoverIcon} />
+                <img
+                  className={cssClass.ACTION_HOVER_ICON}
+                  src={singleActions[0].hoverIcon}
+                  alt={singleActions[0].title.singular.toString()}
+                />
               )}
               <div className={cssClass.ACTION_TITLE}>{singleActions[0].title.singular}</div>
             </>
```
cofauver, line 489: With this exact code block (I think) repeated below, can you capture the JSX in a variable and use that variable instead of duplicating? Especially with this markup appearing within a prop, `value`, it feels like it could simplify reading the `Button` component.

src/Table2Beta/Table.tsx
cofauver: I'm not exactly sure the context, but maybe "More options"? I don't know the range of "actions" being used here, but the word "actions" feels like it's tied to the implementation and maybe not as clear to the user (who doesn't know that the button is an `ACTION_MENU`).

cofauver (approved):
A few nit-picky things here and there that you can decide how to respond to. None should block this from merging though!

jonah-clever:
here's the screenshot

Screen Shot 2022-09-15 at 4 23 30 PM

src/Checkbox/Checkbox.tsx
```diff
@@ -81,11 +81,14 @@ export default class Checkbox extends React.PureComponent<Props> {
...
+        // TODO: aria-label needs to be translated
+        aria-label={!hasLabel ? "select/unselect" : undefined}
```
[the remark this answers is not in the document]
jonah-clever: Sounds good to me!

src/Table2Beta/Table.tsx
jonah-clever: Love it!

jonah-clever:
https://github.com/Clever/components/pull/808#discussion_r972491176

I think this makes sense given that the column is by default called "Actions" (which you wouldn't've known without the screenshot)

cofauver (approved):

PR #810, opened 2022-09-19
adding aria-label prop

Maddiebusta:
# Jira: [PRTL-2597](https://clever.atlassian.net/browse/PRTL-2597)

# Overview:
adding aria-label prop to select component for accessibility

# Screenshots/GIFs:
Screen Shot 2022-09-19 at 5 38 50 PM

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/Select/Select.tsx
+++ b/src/Select/Select.tsx
@@ -15,7 +15,6 @@ export interface SelectValueType {
 }
 
 export interface Props {
-  labelText?: string;
   id: string;
   name: string;
   clearable?: boolean;
@@ -42,6 +41,7 @@ export interface Props {
   error?: string;
   size?: Size;
   closeMenuOnSelect?: boolean;
+  ariaLabel?: string;
 }
 
 interface State {
@@ -54,7 +54,6 @@ const selectValuePropType = PropTypes.shape({
 });
 
 const propTypes = {
-  labelText: PropTypes.string,
   id: PropTypes.string.isRequired,
   name: PropTypes.string.isRequired,
   clearable: PropTypes.bool,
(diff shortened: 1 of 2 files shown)


src/Select/Select.tsx
```diff
@@ -15,6 +15,7 @@ export interface SelectValueType {
 }
 
 export interface Props {
+  labelText?: string;
```
cofauver, line 18: Curious about the name `labelText` vs. `ariaLabel`. If it's called `ariaLabel` it feels very clear from the outside that the value passed along will eventually be used for a11y purposes (it'll eventually be used as an `aria-label` in the DOM). `labelText` on the other hand, is difficult to distinguish from `label` as far as its eventual purpose. Somebody may need to dive into the source code to figure out what it's used for and how the two are different.

This piece matters less, but I'm curious what you think of listing this new prop last on these lists rather than first. The new prop is optional and currently is not widely used, so maybe that first position in the list could be saved for the two required props (`id` and `name`)?

cofauver (approved):
I'd suggest shifting the prop name back to `ariaLabel` and I have one more comment outside of that, but if you take those suggestions, I give it a 👍 ! If you go another direction from those suggestions, just tag me in a comment and we can discuss!

src/Select/Select.tsx
Maddiebusta: good point! I'll make those edits

PR #816, opened 2022-10-19
add optional placeholder prop to Select2

incakoala:
# Overview:
Adding an optional `placeholder` prop to `Select2` component. Motivated by a design change in [PRTL-3047](https://clever.atlassian.net/browse/PRTL-3047) that requires placeholders in the teacher resource dropdowns

# Screenshots/GIFs:
Oct-19-2022 13-01-23

**Tested locally in launchpad:**

Oct-19-2022 13-09-10

# Testing:

- [ ] Unit tests
- Manual tests:
  - [ ] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.207.0",
+  "version": "2.208.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/Select2/Select2.tsx
+++ b/src/Select2/Select2.tsx
@@ -29,6 +29,7 @@ export interface Props {
   value: string;
   onChange: (value: string) => void;
   size?: Values<typeof FormElementSize>;
+  placeholder?: string;
 }
 
 export const cssClass = {
@@ -71,6 +72,7 @@ const Select2: React.FC<Props> = ({
   value,
   onChange,
   size,
+  placeholder,
 }) => {
   const [createdOption, setCreatedOption] = useState<Option | null>(null);
   const options = createdOption ? [...initialOptions, createdOption] : initialOptions;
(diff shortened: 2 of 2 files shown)


cofauver:
Because `Select` has been ported over to dew3y, it might make sense to also update the component there so that they continue to support the same feature-set. [This would be the file to update](https://github.com/Clever/clever-ui/blob/master/packages/ui-select/src/lib/Select.tsx) in that effort. 

Probably worth checking: @zohnnyclever , is the `Select` in dew3y meant to support the same features as this `Select2` component?

cofauver (approved):
This seems like a good idea! We might also need to add support for a `placeholder` in dew3y because we have a `Select` component ported over there too.

zohnnyclever:
> This seems like a good idea! We might also need to add support for a `placeholder` in dew3y because we have a `Select` component ported over there too.

That is an excellent point.. thank you for bringing it up!

@incakoala , the current `dew3y` Select doesn't seem to support the `placeholder` attribute as far as I can tell

https://master--62fbbb0daff0aa52aea739ac.chromatic.com/?path=/docs/clever-ui-packages-clever-ui-select-select--primary

Please at the least, reach out to @ashleyqw , who authored the `dew3y Select` https://github.com/Clever/clever-ui/tree/master/packages/ui-select and see if she thinks we should add the attribute.

If so, I'm happy to be your guide to the new world of `dew3y`. it's a worthwhile skill to learn since are soon moving to that in Q1 and Q2 next year.

incakoala:
Thanks both! I created a backlog task for myself to port this over to dew3y: https://clever.atlassian.net/browse/FEE-215

zohnnyclever:
Please put that tix in the dew3y epic, PRTL-2919

PR #833, opened 2023-07-10
change WizardLayout subtitle prop type from String to ReactNode

incakoala:
# Overview:
As part of https://clever.atlassian.net/browse/DID-1489, we want to pass in `ReactNode` as prop to `subtitle`. By passing in `ReactNode` we can support more styling in the subtitle, and this also won't break any existing instances that pass in `String`

# Testing:
- [x] tested locally:
Screenshot 2023-07-10 at 2 31 26 PM
<br></br>

- [x] copied `dist` directory to test with local sd2 and made sure nothing breaks and that the change doesn't cause backward issues:
Screenshot 2023-07-10 at 3 04 01 PM

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

### DIFF
--- a/src/WizardLayout/WizardLayout.tsx
+++ b/src/WizardLayout/WizardLayout.tsx
@@ -22,7 +22,7 @@ export interface Props {
   prevStepButtonDisabled?: boolean;
   prevStepButtonText?: React.ReactNode;
   stepper: React.ReactNode;
-  subtitle?: string;
+  subtitle?: React.ReactNode;
   title: string;
 }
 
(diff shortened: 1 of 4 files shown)


src/WizardLayout/WizardLayout.tsx
```diff
@@ -112,7 +112,7 @@ export default class WizardLayout extends React.PureComponent<Props> {
           {headerImg && <div className={cssClass.HEADER_IMG}>{headerImg}</div>}
           <div>
             <p className={cssClass.HEADER_TITLE}>{title}</p>
-            {subtitle && <p className={cssClass.HEADER_SUBTITLE}>{subtitle}</p>}
+            {subtitle && <div className={cssClass.HEADER_SUBTITLE}>{subtitle}</div>}
```
dengjulie, line 115: > major web browsers will render a <p> tag with margin above and below the paragraph. A <div> tag will be rendered without any margin at all

Can we conditionally render p and div tags depending on if the subtitle is a string so that there's no UI regression?

dengjulie (approved):
LGTM

PR #835, opened 2023-08-02
Changing WizardLayout title prop type from String to ReactNode

acestti:
# Jira: [UM-3030](https://clever.atlassian.net/browse/UM-3030)
Making step titles type be react nodes instead of strings so that GCRi wizard can have an "Optional" pill in the name. This meant changing the tag in which it was displayed in from `<p>` to `<div>`.

Screenshot 2023-08-03 at 1 56 44 PM

# testing:
manual testing:
Making sure styling is still correct for steps with subtitles and ones without. 

Step with subtitle (before):
Screenshot 2023-08-03 at 2 01 58 PM

step with subtitle (after)
Screenshot 2023-08-03 at 2 01 50 PM

Step without subtitle (before):
Screenshot 2023-08-03 at 2 02 42 PM

step without subtitle (after):
Screenshot 2023-08-03 at 2 02 38 PM

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:

[UM-3030]: https://clever.atlassian.net/browse/UM-3030?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.218.0",
+  "version": "2.219.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.219.0",
+  "version": "2.217.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.217.0",
+  "version": "2.218.0",
   "description": "A library of helpful React components and less styles",
   "repository": {
     "type": "git",
--- a/src/WizardLayout/WizardLayout.less
+++ b/src/WizardLayout/WizardLayout.less
@@ -89,6 +89,7 @@
   .text--large();
   .text--semi-bold();
   .margin--bottom--none();
+  .padding--y--xs();
   width: 100%;
   color: @neutral_dark_gray;
 }
--- a/src/WizardLayout/WizardLayout.less
+++ b/src/WizardLayout/WizardLayout.less
@@ -89,7 +89,7 @@
   .text--large();
   .text--semi-bold();
   .margin--bottom--none();
-  .padding--y--xs();
+  line-height: 1.5;
   width: 100%;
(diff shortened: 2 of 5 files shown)


src/WizardLayout/WizardLayout.less
```diff
@@ -89,6 +89,7 @@
   .text--large();
   .text--semi-bold();
   .margin--bottom--none();
+  .padding--y--xs();
```
brollin, line 92: Does padding match the default padding of the <p> tag out of curiosity?

shannenlm: I think we should follow what [a recent PR](https://github.com/Clever/components/pull/833) did that's almost identical to the changes here: They conditionally render a `<p>` or `<div>` tag depending on if the prop is a string or a React node

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "clever-components",
-  "version": "2.217.0",
+  "version": "2.218.0",
```
shannenlm, line 3: you might have to bump the version in package-lock.json too?

shannenlm: Running `npm version minor` from [the PR template](https://github.com/Clever/components/blob/master/.github/pull_request_template.md) should handle this ?

src/WizardLayout/WizardLayout.less
acestti: This pr is a little different because it handles the subtitle of the entire wizard which makes changes between paddings inconspicuous. In mine, it is a step in a wizard and when stepping through small differences can be noticed fairly well. This is why I wanted to make them all have the same padding but it is a little different than the default. I can try further to make them match so we can do the render conditionally without drawing attention when going to the next step.

shannenlm: ah, okay, makes sense. I defer to Ben's question then!

acestti: @brollin now it is the same. it was due to the line-height property of the `<p>` tag.

brollin: Seems reasonable to me! 👍

brollin (approved):

shannenlm (approved):

shannenlm:
oh can you make sure to update the docs too? like the rollout plan of the PR says?

PR #836, opened 2023-08-14
UM-3054: Adding tooltip for next button in WizardLayout

acestti:
# Jira: [UM-3054](https://clever.atlassian.net/browse/UM-3054)

# Overview:
We want to be able to show a tooltip for the next button (not just the next button text).

# Screenshots/GIFs:
Screenshot 2023-08-14 at 1 26 01 PM

# Testing:
ensured other wizards still worked. These are optional props so it shouldn't break anything. tooltip is hidden on default. 

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

[UM-3054]: https://clever.atlassian.net/browse/UM-3054?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/components/WizardLayoutView.jsx
+++ b/docs/components/WizardLayoutView.jsx
@@ -257,6 +257,20 @@ export default class WizardLayoutView extends React.PureComponent {
             optional: true,
           },
           {
+            name: "nextStepButtonTooltipEnabled",
+            type: "Boolean",
+            description: "Optional tooltip for the next step button",
+            defaultValue: false,
+            optional: true,
+          },
+          {
+            name: "nextStepButtonTooltipText",
+            type: "String",
+            description: "Optional next button tooltip text",
+            defaultValue: "default text",
+            optional: true,
+          },
+          {
             name: "onNextStep",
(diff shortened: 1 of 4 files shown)


brollin:
@acestti just wondering if this change is design approved? Its a little bit of an unfamiliar pattern is all.

I also think it is maybe fine, but wrapping the button in a tooltip and a span in every case going forward could cause unexpected downstream effects.

I could imagine use doing something like:
- only adding one additional prop, the tooltip text
- if the tooltip text is missing or an empty string, we don't render the enclosing tooltip or span

This way, in the default case we don't change the wait that things are rendering today. Also, with a single prop we can avoid an invalid situation: the tooltip is enabled, but no tooltip text is supplied. Designing props such that only valid states are possible is a good idea where we can do so. Happy to pair on this if any of that is unclear!

brollin (approved):
Nice!

PR #842, opened 2023-10-19
Prtl 4849 allow custom ref for menu trigger element

chloehunter:
# Jira: [PRTL-4849](https://clever.atlassian.net/browse/PRTL-4849)

# Overview:

Currently, passing a `ref` to the Menu puts it on the Menu element rather than the trigger, and passing a `ref` to the trigger element leads to it being ignored in favor of the in-component use of the trigger ref.

For the use case of needing to return focus for accessibility reasons (e.g. when a modal is closed, you must return focus to the element which opened it), we need the option to override the default ref behavior. As the menu is closed when the modal opens, we need to return focus to the trigger element of the menu.

# Screenshots/GIFs:

# Testing:

- [ ] Unit tests
- Manual tests:
  - [x] Chrome
  - [x] Firefox

# Roll Out:

- Before merging:
  - [x] Updated docs
    - Breaking change?
- After merging:

[PRTL-4849]: https://clever.atlassian.net/browse/PRTL-4849?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/docs/components/MenuView.jsx
+++ b/docs/components/MenuView.jsx
@@ -343,6 +343,13 @@ export default class MenuView extends React.PureComponent {
             ),
             default: false,
           },
+          {
+            name: "triggerRefOverride",
+            type: "React.RefObject<HTMLElement>",
+            description:
+              "Optional ref override for the triggering element, for accessibility purposes (e.g. returning focus after modal close).",
+            optional: true,
+          },
         ]}
         className={cssClass.PROPS}
       />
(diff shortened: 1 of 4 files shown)


Maddiebusta (approved):
yeehaw

PR #866, opened 2025-05-19
Dd 6430 rootclosewrapper

dxu429:
# Jira: 
https://clever.atlassian.net/browse/DD-6430

# Overview:
More pre-work to allow sd2 to start using react 17. This PR primarily updates a really old dependency `react-overlays` where we use the `RootCloseWrapper` + several other changes to clean up this package
1. Move dependencies that should be in dev deps to `devDependencies` 
2. remove redundant `excludes` in tsconfig, it should represent a filter on top of the `includes` files/globs
3. Update the `react-overlays` dependency to its latest `@restart/ui`, the major change here is that RootCloseWrapper has been removed since the underlying API uses `findDOMNode` which is not recommended for use by React. The library provides a hook `useRootClose` which takes a ref for the container. Since its a hook and all uses of the old component are in class components, I've shimmed the functionality using a functional component, rather than re-writing all the class components as functional components. 

# Screenshots/GIFs:
https://github.com/user-attachments/assets/58549e3f-7b2c-4bab-ad3f-c41fd54c1cb5

https://github.com/user-attachments/assets/a42fc809-3e84-4a26-9680-532dff6f45c5

https://github.com/user-attachments/assets/939780c3-5ba0-4600-9974-c1bde53560b9

# Testing:

- [x] Unit tests
- Manual tests:
  - [x] Chrome
  - [ ] Safari
  - [ ] IE11

# Roll Out:

- Before merging:
  - [ ] Updated docs
    - Breaking change?
- After merging:


src/LeftNav/NavLink.tsx
```diff
@@ -102,8 +104,12 @@ export class NavLink extends React.PureComponent<Props> {
               })}
             </div>
           )}
-          <div className={cssClass.LABEL_CONTAINER}>
-            <div className={cssClass.LABEL} title={(title || label) as any}>
+          <div className={cssClass.LABEL_CONTAINER} aria-hidden={_collapsed}>
+            <div
+              className={cssClass.LABEL}
+              title={(title || label) as any}
```
dominicgar, line 110: question: should we add stronger typing here?

src/LeftNav/LeftNav.tsx
```diff
@@ -139,43 +139,50 @@ export class LeftNav extends React.PureComponent<Props, State> {
...
+            <TransitionGroup
+              className={classnames(cssClass.SUBNAV, openChild && cssClass.SUBNAV_OPEN)}
+              component="div"
+              transitionName={cssClass.SUBNAV_CONTENT_ANIM}
+              aria-hidden={!openNavGroup}
+            >
+              {openChild && (
+                <div className={cssClass.SUBNAV_CONTENT} role="menu">
+                  {React.Children.map(openChild.props.children, (child) => {
+                    // ^ cannot use normal array map, because open.props.children may be an array or a single react element
```
dominicgar, line 168: praise: Thanks for adding a comment to explain what's going on here 🎉

src/LeftNav/LeftNav.tsx
```diff
@@ -139,43 +139,50 @@ export class LeftNav extends React.PureComponent<Props, State> {
...
+              narrow && cssClass.NARROW,
+              openChild && cssClass.WITH_SUBNAV_OPEN,
+              className,
             )}
-          </TransitionGroup>
-        </nav>
+          >
+            <div
+              className={classnames(cssClass.TOPNAV, _collapsed && cssClass.TOPNAV_COLLAPSED)}
+              role="menu"
```
dominicgar, line 155: nit: I wonder if maintainability would improve if we convert this and the other role names into constants? For example, this can be renamed as MENU_ROLE. I don't particularly feel strong about making this change though!

dominicgar (approved):
LGTM. Just had a few questions/nits. Also great job on just shimming the functionality! 🎉

src/LeftNav/NavLink.tsx
dxu429: Probably but label is a ReactNode and div.title should be a string... I actually didn't touch this code the changes here are whitespace cahnges & also this component isn't used in our codebase. I'm just updating it to fix the RootCloseWrapper & tests

